PR #26, opened 2019-07-03
[cudamapper] Generating absolute paths for cudamapper tests

mimaric:
Until now the tests used relative paths which made the CI sensitive to the execution location. Now it's done like in cudapoa


cudamapper/tests/Test_CudamapperMatcher.cpp
```diff
@@ -10,16 +10,17 @@
...
 #include "../src/matcher.hpp"
 
 namespace claragenomics {
 
     TEST(TestCudamapperMatcher, SampleTest) {
-        IndexGeneratorCPU index_generator("data/one_read_one_minimizer.fasta", 4, 1);
-        IndexCPU index(index_generator);
-        Matcher matcher(index);
+        IndexGeneratorCPU index_generator(std::string(CUDAMAPPER_BENCHMARK_DATA_DIR) + "/one_read_one_minimizer.fasta", 4, 1);
+//        IndexCPU index(index_generator);
```
tijyojwad: let's remove these commented lines if they're not necessary

mimaric: I uncommented those lines as that's dummy data we're going to do the tests on once the corresponding code is ready

PR #28, opened 2019-07-08
[cga] adding error checks around all kernels calls

tijyojwad:


tijyojwad:
@ahehn-nv wanted to close the loop on this one. Looks like that CUDA 10.0 genfe error shows up when using c++14 standard (which was being required for constexpr usage in the kernels). In this PR I've changed it from `constexpr` to `const`. Let me know if this is okay, or if it'll cause perf issues down the line (or if this change will limit future work).

Once we standardize to CUDA 10.1 we will be okay, but for now CI is limited to CUDA 10.0

ahehn-nv:
It could have performance implications, depending on how smart the compiler is. I haven't checked... 
In any case, I would not just replace them by `const`, because we will forget about it and never change it back that way. I would use a macro `CGA_CONSTEXPR` or something, which you define depending on the NVCC version. How does that sound to you?

ahehn-nv:
If this can wait one day, I'd like to give it another thought.

tijyojwad:
@ahehn-nv yeah can definitely wait. i like the nvcc version based `ifdef`, but wasn't sure if that could be queried.

ahehn-nv:
Yes, I guess the best solution is a macro definition depending on the compiler version. You can check the version via `__CUDACC_VER_MAJOR__` + `__CUDACC_VER_MINOR__` or `__CUDACC_VER_BUILD__`.

tijyojwad:
@ahehn-nv since the WAR for CUDA 10.0 was checked in, this works now

tijyojwad:
@ahehn-nv if all looks good, can you approve?

PR #32, opened 2019-07-11
[cudamapper] Working Matcher implementation and tests

mimaric:
Still not taking sketch element direction into account, i.e. if one sketch element has a certain representation in forward and the other one in reverse complement they are going to be matched

### DIFF
--- a/cudamapper/src/matcher.cu
+++ b/cudamapper/src/matcher.cu
@@ -40,7 +40,7 @@ namespace claragenomics {
     /// Assume q0p4t2p8 means anchor of read id 0 at position 4 and read id 2 at position 8.
     /// Assume read 0 has 30 sketch elements with certain representation, read 1 40 and read 2 50.
     /// Anchors for read 0 as query and that represtnation looks like this:
-    /// q0p0t1p0, q0p0t1p1 .. q0p0t1p39, q0p0t2p0, q0p0t2p1 ... q0p0t2p49, q0p1t1p0, q0p1t1p1 ... q0p1t1p39, q0p1t2p0 .. q0p1t2p49, q0p2p1p0 ...
+    /// q0p0t1p0, q0p0t1p1 .. q0p0t1p39, q0p0t2p0, q0p0t2p1 ... q0p0t2p49, q0p1t1p0, q0p1t1p1 ... q0p1t1p39, q0p1t2p0 .. q0p1t2p49, q0p2t1p0 ... q0p2t1p39, q0p2t2p0 ... q0p2t2p49, q0p3t1p0 ... q0p29t2p49
     ///
     /// \param positions_in_reads_d positions of sketch elements in their reads first sorted by representation and then by read id
     /// \param read_ids_d read ids of reads sketch elements belong to first sorted by representation and then by read id (elements with the same index from positions_in_reads_d belong to the same sketch element)
--- a/cudamapper/src/matcher.cu
+++ b/cudamapper/src/matcher.cu
@@ -48,7 +48,7 @@ namespace claragenomics {
     /// \param read_id_to_sketch_elements_to_check_d every element points to a section of positions_in_reads_d and read_ids_d that belong to sketch elements with the same representation and read_id larger than some value
     /// \param read_id_to_pointer_arrays_section_d every element belongs to one read_id and points to its sections of read_id_to_sketch_elements_d and read_id_to_sketch_elements_to_check_d
     /// \param anchors_d pairs of sketch elements with the same representation that belong to different reads
-    /// \param read_id_to_anchors_section_d points to parts of andchors_d in which all anchors have the same read_id
+    /// \param read_id_to_anchors_section_d points to parts of anchors_d in which all anchors have the same read_id
     __global__ void generate_anchors(const position_in_read_t* const positions_in_reads_d,
                                      const read_id_t* const read_ids_d,
                                      // const SketchElement::DirectionOfRepresentation* const directions_of_reads_d, // currently we don't use direction
--- a/cudamapper/src/matcher.cu
+++ b/cudamapper/src/matcher.cu
@@ -8,107 +8,236 @@
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
 
+#include <algorithm>
+#include <memory>
 #include "matcher.hpp"
+#include <cudautils/cudautils.hpp>
 
 namespace claragenomics {
 
-/*    typedef struct Anchor{
-        read_id_t query_read_id_;
-        read_id_t target_read_id_;
-        position_in_read_t query_position_in_read_;
-        position_in_read_t target_position_in_read_;
-    } Anchor;*/
+    // TODO: move to cudautils?
+    /// \brief creates a unique pointer to device memory which gets deallocated automatically
+    ///
+    /// \param num_of_elems number of elements to allocate
+    ///
+    /// \return unique pointer to allocated memory
+    template<typename T>
+    std::unique_ptr<T, void(*)(T*)> make_unique_cuda_malloc(std::size_t num_of_elems) {
+        T* tmp_ptr_d = nullptr;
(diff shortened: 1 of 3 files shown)


cudamapper/src/matcher.hpp
```diff
@@ -10,19 +10,49 @@
...
+    ///
+    /// Anchor is a pair of two sketch elements with the same sketch element representation from different reads.
+    /// Anchors are symmetrical, so only one anchor is generated for each pair (if sketch element from read 5 overlaps a sketch element from read 8
+    /// then the same sketch element from read 8 overlaps the sketch element from read 5).
+    /// 
+    /// Anchors are grouped by query read id and within that by representation (both in increasing order).
+    /// Assume q0p4t2p8 means anchor of read id 0 at position 4 and read id 2 at position 8.
+    /// Assume read 0 has 30 sketch elements with certain representation, read 1 40 and read 2 50.
+    /// Anchors for read 0 as query and that represtnation looks like this:
+    /// q0p0t1p0, q0p0t1p1 .. q0p0t1p39, q0p0t2p0, q0p0t2p1 ... q0p0t2p49, q0p1t1p0, q0p1t1p1 ... q0p1t1p39, q0p1t2p0 .. q0p1t2p49, q0p2p1p0 ...
```
vellamike: I don't understand the last entry `q0p2p1p0`, why the 3 ps?

cudamapper/tests/Test_CudamapperMatcher.cpp
```diff
@@ -17,11 +17,379 @@
...
+        test_index.representation_to_all_its_sketch_elements(representation_to_all_its_sketch_elements);
+
+        Matcher matcher(test_index);
+
+        const std::vector<Matcher::Anchor>& anchors = matcher.anchors();
+        ASSERT_EQ(anchors.size(), 2500);
+
+        for (std::size_t read_0_sketch_element = 0; read_0_sketch_element < 50; ++read_0_sketch_element) {
+            for (std::size_t read_1_sketch_element = 0; read_1_sketch_element < 50; ++read_1_sketch_element) {
+                ASSERT_EQ(anchors[read_0_sketch_element*50 + read_1_sketch_element].query_read_id_, 0) << read_0_sketch_element << " " <<read_1_sketch_element;
```
vellamike: Should we be using `EXPECT_EQ` in these situations so that if there is a failure all information is printed (subsequent checks in this function) ?

cudamapper/tests/Test_CudamapperMatcher.cpp
```diff
@@ -17,11 +17,379 @@
...
+            }
+        }
+
+        // read 0 - rep 3: 130
+        // read 1 - rep 3: 70
+        // read 2 - rep 3: 100
+        // read 3 - rep 3: 80
+        for (std::size_t query = 0; query < 130; ++query) { // block starts from 20*30 = 600
+            for (std::size_t target = 0; target < 70; ++target) { // read 1 - no shift
+                ASSERT_EQ(anchors[600 + query*250 + target].query_read_id_, 0)  << "query: " << query << ", target: " << target;
```
vellamike: As above re `EXPECT` vs `ASSERT`

cudamapper/src/matcher.cu
```diff
@@ -8,107 +8,236 @@
...
-
-/*    __global__ void matcher(const position_in_read_t* const positions_in_read_d,
-                            const read_id_t* const read_ids_d,
-                            const ArrayBlock* const blocks_belonging_to_reads_d,
-                            const ArrayBlock* const sketch_elements_with_same_representation_to_check_d,
-                            const ArrayBlock* const sketch_elements_with_same_representation_and_read_d,
-                            Anchor** const anchors_for_read_d,
-                            std:uint32_t* const total_anchors_for_read_d) {
-        extern __shared__ position_in_read_t query_positions_in_read[]; // number of elements == blockDim.x
+    // TODO: move to cudautils?
```
vellamike: Agree with this TODO. Should we move it in this PR?

cudamapper/src/matcher.cu
```diff
@@ -8,107 +8,236 @@
...
+    ///
+    /// \param positions_in_reads_d positions of sketch elements in their reads first sorted by representation and then by read id
+    /// \param read_ids_d read ids of reads sketch elements belong to first sorted by representation and then by read id (elements with the same index from positions_in_reads_d belong to the same sketch element)
+    /// \param read_id_to_sketch_elements_d every element points to a section of positions_in_reads_d and read_ids_d that belong to sketch elements with the same read_id and representation
+    /// \param read_id_to_sketch_elements_to_check_d every element points to a section of positions_in_reads_d and read_ids_d that belong to sketch elements with the same representation and read_id larger than some value
+    /// \param read_id_to_pointer_arrays_section_d every element belongs to one read_id and points to its sections of read_id_to_sketch_elements_d and read_id_to_sketch_elements_to_check_d
+    /// \param anchors_d pairs of sketch elements with the same representation that belong to different reads
+    /// \param read_id_to_anchors_section_d points to parts of andchors_d in which all anchors have the same read_id
```
vellamike: typo. anchords

cudamapper/src/matcher.cu
```diff
@@ -8,107 +8,236 @@
...
+    /// Each thread block works on one query read. read_id_to_pointer_arrays_section_d points sections of read_id_to_sketch_elements_d and read_id_to_sketch_elements_to_check_d.
+    /// Kernel matches sketch elements that have the same representation. These sketch elements are pointed to by read_id_to_sketch_elements_d and read_id_to_sketch_elements_to_check_d.
+    ///
+    /// Kernel has an extern shared memory array which should be at least as large as the largest number of sketch elements with the same representation in any block
+    ///
+    /// Anchors are grouped by query read id and within that by representation (both in increasing order).
+    /// Assume q0p4t2p8 means anchor of read id 0 at position 4 and read id 2 at position 8.
+    /// Assume read 0 has 30 sketch elements with certain representation, read 1 40 and read 2 50.
+    /// Anchors for read 0 as query and that represtnation looks like this:
+    /// q0p0t1p0, q0p0t1p1 .. q0p0t1p39, q0p0t2p0, q0p0t2p1 ... q0p0t2p49, q0p1t1p0, q0p1t1p1 ... q0p1t1p39, q0p1t2p0 .. q0p1t2p49, q0p2p1p0 ...
```
vellamike: don't understand the 3 ts in the last element `q0p2p1p0`

cudamapper/tests/Test_CudamapperMatcher.cpp
mimaric: That was my initial implementation, but it's highly likely that if one test fails all tests will fail and we'd end up with 50x50x4 error messages (or even more in the lower test).
If all checks in both big tests would fail that would cause ~400000 error messages.
An alternative would be to have first three checks do `EXPECT` and the last one `ASSERT`, assuming all four anchors would be wrong at the same time (which does not have to be the case). What would you suggest?

cudamapper/src/matcher.cu
mimaric: fixed

cudamapper/tests/Test_CudamapperMatcher.cpp
vellamike: I see your point. regarding 10 thousand error messages.. OK Let's leave it as is.

cudamapper/src/matcher.hpp
mimaric: Middle `p` should've been `t`. Fixed and extended the comment to make it more clear

cudamapper/src/matcher.cu
mimaric: fixed, see above

cudamapper/src/matcher.cu
mimaric: Moved to `cudautils`. Please review new commit and resolve if ok.

mimaric: We kept `unique_ptr_cuda_malloc` in `matcher.cu`. Moving it to `cudautils` will likely happen on the main branch

PR #35, opened 2019-07-15
[cga] Smart device pointers

mimaric:
Added helper function which allocate device memory and pack it into a unique/smart pointer

### DIFF
--- a/cmake/Tests.cmake
+++ b/cmake/Tests.cmake
@@ -15,7 +15,7 @@ get_property(enable_tests GLOBAL PROPERTY enable_tests)
 function(cga_add_tests NAME SOURCES LIBS)
     # Add test executable
     if (enable_tests)
-        add_executable(${NAME} ${SOURCES})
+        cuda_add_executable(${NAME} ${SOURCES})
 
         # Link gtest to tests binary
         target_link_libraries(${NAME}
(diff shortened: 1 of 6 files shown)


ahehn-nv:
I think we should discuss on our next meeting whether we want to manage memory using smart ptrs or have custom container-like objects (e.g. promote https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/master/cudaaligner/src/device_storage.cuh).

tijyojwad:
@ahehn-nv @mimaric I think these utilities are generally useful to have. Custom container objects can be specific to modules that want to implement them for their use cases, so we can support both. What do you think?

and I think we can also move device_storage under `cudautils.hpp`, which could use `make_unique_cuda_malloc` in its implementation

mimaric:
Ok, let's discuss whether we want this in our sync call. Merging it is not urgent.

tijyojwad:
@mimaric can this be closed?

mimaric:
Yes it can. Did the device container work get merged/moved to utils?

ahehn-nv:
Not yet, thanks for the reminder!

mimaric:
Closing this bug. Similar behavior will be added in PR 63

PR #36, opened 2019-07-15
[cga] generate shared libs

tijyojwad:
added cmake optoin to generate shared libs for
cga modules

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,6 +4,7 @@ project(ClaraGenomicsAnalysisCPP)
 # Process options.
 option(cga_enable_tests "Build ClaraGenomicsAnalysis unit tests" OFF)
 option(cga_enable_benchmarks "Build ClaraGenomicsAnalysis benchmarks" OFF)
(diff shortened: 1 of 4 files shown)


tijyojwad:
@vellamike this is needed for the python bindings, so checking it in first

tijyojwad:
@vellamike considered a few alternatives, but this seemed the least intrusive out of - 

1. wrap the add_library/cuda_add_library function and handle the logic within it, but I don't want to do that unless absolutely necessary. I am personally not a fan of wrapping important API calls.
2. expose STATIC/SHARED as global define (e.g. -Dcga_build_type in the cmake command line). that means using the `${cga_build_type}` directly in the add_library function, which is shorter true but I think ties us to a very strict API.
3. expose an option that enabled shared libs. then each module can choose what to do with that option. the global property is the canonical way to save global settings in cmake. if in the future modules want to do something more when `cga_build_shared` is set, they can.
4. use the `CMAKE_BUILD_SHARED` flag, but that builds ALL libraries as shared (including gtest, benchmarks, spoa, etc). but we don't want those to be installed as part of CGA.

vellamike:
OK I agree that the solution you've gone for is the best one. Merged

PR #41, opened 2019-07-18
[ci] add check for copyright header

tijyojwad:
Add script to check for copyright header
Fail CI if copyright headers are missing

### DIFF
--- a/ci/checks/check_copyright.py
+++ b/ci/checks/check_copyright.py
@@ -1,0 +1,164 @@
+#!/usr/bin/env python
+
+#
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+"""
+Script to check for NVIDIA Copyright headers in source files and
+add optionally copyright header to them.
+"""
+
+import argparse
+import io
+import os
+import re
+import stat
+from subprocess import check_output
+import sys
+
+cpp_exts = {".hpp", ".cpp", ".cu", ".cuh", ".cc", ".h", ".c"}
+
+cpp_copyright = r"""/*
+* Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+*
+* NVIDIA CORPORATION and its licensors retain all intellectual property
+* and proprietary rights in and to this software, related documentation
+* and any modifications thereto.  Any use, reproduction, disclosure or
+* distribution of this software and related documentation without an express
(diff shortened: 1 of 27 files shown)


ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+
+    # Git list of files missing headers.
+    missing_header_files = list()
+    for file in files:
+        if(not copyright_present(file)):
+            missing_header_files.append(file)
+
+    # Check/Add headers if need be.
+    if (args.add_header):
+        map(add_copyright, missing_header_files)
```
vellamike: a matter of style, but list comprehensions are [preferred to map](https://www.artima.com/weblogs/viewpost.jsp?thread=98196)

ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+#
+
+"""
+Script to check for NVIDIA Copyright headers in source files and
+add optionally copyright header to them.
+"""
+
+import sys
+import os
+import mmap
```
vellamike: unused import

ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+import argparse
+import re
+from subprocess import check_output
+from stat import *
```
vellamike: best to never use import *

ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+"""
+Script to check for NVIDIA Copyright headers in source files and
+add optionally copyright header to them.
+"""
+
+import sys
```
vellamike: could these imports be sorted alphabetically

ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+
+
+if __name__ == "__main__":
+    args = parse_args()
+
+    # Get list of files to check.
+    script_dir = os.path.dirname(os.path.realpath(__file__))
+    root_dir = os.path.abspath(os.path.join(script_dir, "../../"))
+    files = get_tracked_files(root_dir)
+    files = filter_files(files, "LICENSE|README|data\/|docs\/")
```
vellamike: this string should be a raw string r"LICENSE|README|data\/|docs\/". Otherwise the `\/` character combination can be interpreted as an escape sequence (probably an invalid one) by different tools, editors etc

ci/checks/check_copyright.py
```diff
@@ -0,0 +1,169 @@
...
+            missing_header_files.append(file)
+
+    # Check/Add headers if need be.
+    if (args.add_header):
+        map(add_copyright, missing_header_files)
+    else:
+        if (len(missing_header_files) > 0):
+            print("List of files missing copyright headers - ")
+            print("\n".join(missing_header_files))
+            sys.exit(1)
```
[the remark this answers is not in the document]
tijyojwad: to communicate that there are files without the copyright header so other tests can use that

PR #42, opened 2019-07-19
[common] Added macros for CUDA version specific code

ahehn-nv:
Macros to guard CUDA version specific code which are easy to grep for.

### DIFF
--- a/common/cudautils/include/cudautils/cudautils.hpp
+++ b/common/cudautils/include/cudautils/cudautils.hpp
@@ -26,9 +26,19 @@
         claragenomics::cudautils::gpuAssert((ans), __FILE__, __LINE__); \
     }
 
+/// \def CGA_CUDA_BEFORE_XX_X
+/// \brief Macros to enable/disable CUDA version specific code
+#define CGA_CUDA_BEFORE_XX_X 1
+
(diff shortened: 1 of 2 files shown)


common/cudautils/include/cudautils/cudautils.hpp
```diff
@@ -26,9 +26,19 @@
...
+#define CGA_CUDA_BEFORE_10_1 0
+#endif
+
 /// \def CGA_CONSTEXPR
 /// \brief C++ constexpr for device code - falls back to const for CUDA 10.0 and earlier
-#if __CUDACC_VER_MAJOR__ < 10 || (__CUDACC_VER_MAJOR__ == 10 && __CUDACC_VER_MINOR__ < 1)
+#if CGA_CUDA_BEFORE_10_1
 #define CGA_CONSTEXPR const
 #else
 #define CGA_CONSTEXPR constexpr
```
vellamike: won't this `#define CGA_CONSTEXPR constexpr` for CUDA=10.1, which causes the nvcc error?

ahehn-nv: Yes, but `constexpr` should not be a problem in 10.1.

PR #49, opened 2019-07-22
[cudapoa] optimizations to reduce memory footprint and improve resource utilization

akkamesh:
This PR introduces the following optimizations to cudapoa -
- Reduces memory footprint required for POA calculation
- Add support for processing a variable number of windows per cuda-batch instead of a statically sized number of windows per batch. This improves resource utilization on the device.

Overall it improves the performance of the CUDAPolisher by 15-20%.

Note - This PR requires [racon-gpu PR #8](https://github.com/clara-genomics/racon-gpu/pull/8).


cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -68,6 +68,25 @@ namespace claragenomics
...
+};
+enum
+{
+    MASK = ALIGN - 1
+};
+
+template <typename IntType>
+__host__ __device__ __forceinline__
+    IntType
+    align(const IntType& size)
```
tijyojwad: we should rename this to `alignTo4` since this is aligning to a boundary of 4

PR #50, opened 2019-07-25
[common] Use gtest *_EQ on std::strings

ahehn-nv:
Replaced the ASSERT_STREQ/EXPECT_STREQ with ASSERT_EQ/EXPECT_EQ for std::string.

*_STREQ is only required for c-strings, where you pass a pointer but want to compare content rather than the addresses.


tijyojwad:
rerun tests

ahehn-nv:
rerun tests

tijyojwad:
rerun tests

PR #52, opened 2019-07-26
[cgautils] update include paths in CGA

tijyojwad:
move cudautils to cgautils folder


GPUtester:

dillon-cullinan:
add to whitelist

dillon-cullinan:
rerun tests

tijyojwad:
In terms of the repo directory structure it feels odd to me to have a top level `claragenomocs` folder when it's already under the ClaraGenomicsAnalysis repo folder. For installation we can always change the folder name and install into a dir called `claragenomics` and the structure would look like
```
claragenomics/include/cudapoa
claragenomics/include/cudaaligner
```
Are you concerned that we have two `cgautila` folders?

ahehn-nv:
No, I'm not concerned about two `cgautils` folders, I just prefer file/path names which make it easy to guess where a file belongs to. If we install the includes to `include/claragenomics/utils/` then there is no need to call it `cgautils`, right?

Same goes for `#includes`:
I don't care too much about the actual repo directory structure, but I would like to include all ClaraGenomics related files via `#include <claragenomics/subproject/file.hpp>` ideally. Then by just looking at it, I know immediately that this file is owned by the project and not by some dependency or the system libraries.

tijyojwad:
> If we install the includes to include/claragenomics/utils/ then there is no need to call it cgautils
Yes , that's correct

>  I would like to include all ClaraGenomics related files via #include <claragenomics/subproject/file.hpp> ideally
This is something I sort of remember discussing in the early days of setting up the structure. I think the resolution was that prepending `claragenomics` (then `genomeworks`) to every include was making it too long, and things like `cudapoa` and `cudaaligner` were unique enough for the public API (e.g. `cuda` does it by just prepending `cuda_` or `cu` to the header names. No folders whatsoever). But I'm open to changing it if most people agree its better. We can bring it up in our weekly sync.

The initial intention was for `cudautils` to be the only public utils folder, and for the rest to be private. But even then `cudautils` is not a common enough name. That's what motivated me to rename it to `cgautils` since our other utils may be of help too. I'm actually thinking that it may be a good idea to make the `utils` folder completely private to CGA. Only thing mostly used by other applications is the `CGA_CU_CHECK_ERR` but apps should just implement their own cuda error handling.

cudaaligner/benchmarks/singlealignment/main.cpp
```diff
@@ -10,9 +10,9 @@
 
 #include <benchmark/benchmark.h>
 
-#include "utils/genomeutils.hpp"
-#include "cudautils/cudautils.hpp"
-#include "cudaaligner/aligner.hpp"
+#include "claragenomics/utils/genomeutils.hpp"
```
ahehn-nv: The new includes look much nicer in my opinion.

But nit-picky me has another request: Could you use `#include <>` for all public headers, and `#include ""` only for internal headers? :)

cudaaligner/benchmarks/singlebatchalignment/main.cpp
```diff
@@ -12,9 +12,10 @@
 #include <cuda_runtime_api.h>
 #include <random>
 
-#include "utils/genomeutils.hpp"
-#include "cudautils/cudautils.hpp"
-#include "cudaaligner/aligner.hpp"
+#include "claragenomics/utils/genomeutils.hpp"
+#include "claragenomics/utils/cudautils.hpp"
```
ahehn-nv: Same here, prefer `<>` to distinguish public from private headers.

cudaaligner/include/claragenomics/cudaaligner/aligner.hpp
```diff
@@ -12,10 +12,9 @@
 
 #include <memory>
 #include <vector>
-
 #include <cuda_runtime_api.h>
 
-#include "cudaaligner/cudaaligner.hpp"
+#include "claragenomics/cudaaligner/cudaaligner.hpp"
```
ahehn-nv: `<>`

cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
```diff
@@ -14,7 +14,7 @@
 #include <vector>
 #include <string>
 
-#include "cudaaligner/cudaaligner.hpp"
+#include "claragenomics/cudaaligner/cudaaligner.hpp"
```
ahehn-nv: `<>`

cudaaligner/src/aligner.cpp
```diff
@@ -8,7 +8,8 @@
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
 
-#include "cudaaligner/aligner.hpp"
+#include "claragenomics/cudaaligner/aligner.hpp"
```
ahehn-nv: `<>`

cudaaligner/src/aligner_global.hpp
```diff
@@ -10,11 +10,13 @@
 
 #pragma once
 
-#include "cudaaligner/aligner.hpp"
+#include <thrust/system/cuda/experimental/pinned_allocator.h>
+
+#include "claragenomics/cudaaligner/aligner.hpp"
```
ahehn-nv: `<>`

cudaaligner/benchmarks/singlealignment/main.cpp
ahehn-nv: Actually, I just noticed, that we decided to make these files private. So you only need to change the file below.
Somehow having them in `""` feels funny. For most projects I worked on all files which required the compiler to get a `-I<directory>` were included via `<>` and those were usually the public headers. This is the first time that these two criteria do not coincide, because the project is somewhat divided into subprojects.

ahehn-nv:
why do we need `common/utils/include/utils` instead of just `common/include/utils`?
To separate the logging project? Can we either make this part of utils or make it a full project like cudapoa and cudaaligner?

cudaaligner/benchmarks/singlealignment/main.cpp
tijyojwad: Indeed as I was making these changes I was bothered by the mixture of `""` and `<>`, so I started making the changes locally and then got distracted by other things XD.
So far I haven't pushed the changes to make the `utils` folders private. So I'm making all of them use `<>` for now. This led me to think that even the private headers will probably get added by the `-I` directive in the compiler based on how `CMake` sets them up. I was thinking of reserving the `""` only for the headers included with relative paths (like local files). What do you think of this?

tijyojwad: Another point that struck me here is that otherwise whenever we make one module public to private, or vice versa, we'll have to go in and change `""` or `<>` in its include files...

ahehn-nv: Good point, although that will prevent us from making too many irrational decisions. ;) Anyhow, yes, I like the idea of local/relative path vs. -I files. Makes sense.

tijyojwad: Done!

tijyojwad:
> why do we need `common/utils/include/utils` instead of just `common/include/utils`?
> To separate the logging project? Can we either make this part of utils or make it a full project like cudapoa and cudaaligner?

Yes, to separate it from `utils`. to me semantically `logging` is different from small `utils` (even though one can argue logging is a utility). It was also that logging right now uses `spdlog`, but in the future we may have to write our own logger. so keeping it separated felt more natural.

I personally like to have `common` folder to draw attention to the fact that library wide shared code should exist in a central location. keeps things more organized in my mind.

cudaaligner/src/myers_gpu.cu
```diff
@@ -8,16 +8,18 @@
...
 #include <vector>
 #include <numeric>
-#include <utils/signed_integer_utils.hpp>
-#include <utils/mathutils.hpp>
+
+#include <claragenomics/utils/signed_integer_utils.hpp>
+#include <claragenomics/utils/mathutils.hpp>
+#include <claragenomics/utils/cudautils.hpp>
+
+#include "myers_gpu.cuh"
```
ahehn-nv: The header file (.h) belonging to the source file (.cpp/.cu) should be the first include, such that the compile fails if the header file is missing an include to be self-contained (e.g. you use vector<T> in your interface, but you didn't include <vector> in your header.)

cudaaligner/src/needleman_wunsch_cpu.cpp
```diff
@@ -8,12 +8,14 @@
...
 
-#include "needleman_wunsch_cpu.hpp"
-#include "utils/mathutils.hpp"
 #include <tuple>
 #include <cassert>
 #include <algorithm>
 
+#include <claragenomics/utils/mathutils.hpp>
+
+#include "needleman_wunsch_cpu.hpp"
```
ahehn-nv: see above

cudaaligner/src/ukkonen_gpu.cu
```diff
@@ -8,13 +8,14 @@
...
-#include "ukkonen_gpu.cuh"
-#include "batched_device_matrices.cuh"
-#include <thrust/tuple.h>
 #include <limits>
 #include <cstdint>
 #include <algorithm>
 #include <cassert>
+#include <thrust/tuple.h>
+
+#include "ukkonen_gpu.cuh"
```
ahehn-nv: see above

cudapoa/src/allocate_block.cpp
```diff
@@ -10,11 +10,11 @@
...
 
-#include "allocate_block.hpp"
+#include <claragenomics/utils/cudautils.hpp>
+#include <claragenomics/logging/logging.hpp>
+#include <claragenomics/utils/signed_integer_utils.hpp>
 
-#include <cudautils/cudautils.hpp>
-#include <logging/logging.hpp>
-#include <utils/signed_integer_utils.hpp>
+#include "allocate_block.hpp"
```
ahehn-nv: see above

cudapoa/src/cudapoa_kernels.cu
```diff
@@ -10,14 +10,15 @@
 
 // Implementation file for CUDA POA kernels.
 
+#include <claragenomics/utils/cudautils.hpp>
+
```
ahehn-nv: see above

cudaaligner/src/aligner_global.cpp
```diff
@@ -10,15 +10,14 @@
...
 #include <algorithm>
-
 #include <cuda_runtime_api.h>
 
-#include <utils/signed_integer_utils.hpp>
+#include <claragenomics/utils/signed_integer_utils.hpp>
+#include <claragenomics/utils/cudautils.hpp>
+#include <claragenomics/logging/logging.hpp>
 
 #include "aligner_global.hpp"
```
ahehn-nv: could you move `aligner_global.hpp` to the top here as well?

cudaaligner/src/aligner_global_myers.cpp
```diff
@@ -8,9 +8,10 @@
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
 
+#include <claragenomics/utils/mathutils.hpp>
+
```
ahehn-nv: see above

cudaaligner/src/alignment_impl.cpp
```diff
@@ -8,8 +8,9 @@
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
 
+#include <claragenomics/utils/signed_integer_utils.hpp>
+
```
ahehn-nv: see above

PR #59, opened 2019-08-01
[cudapoa] fix memory alignment issues

tijyojwad:


cudapoa/src/cudapoa_batch.cpp
```diff
@@ -294,7 +294,7 @@ bool CudapoaBatch::reserve_buf(uint32_t max_seq_length)
 {
```
akkamesh: It seems I missed this in the previous PR, change the signature of `reserve_buf(...)` to 
```
bool CudapoaBatch::reserve_buf(int32_t max_seq_length)
```
To keep it consistent with the input type and to avoid any potential overflow issues from automatic conversion between `int32_t` and `uint32_t`.

tijyojwad: Good catch, I'll do that

tijyojwad:
@akkamesh yeah I was debating that myself. The actual size of the buffers is still the same, but because of the alignment the footprint will be a bit larger than the actual size. Since the max POA calculations is still based on a heuristic, I left it like this. After this release I'll make the change to fully dynamically fill up the buffer based on each incoming POA, and in that case we won't need the calculate size function anymore and we'll do everything based off the remaining buffer size

cudapoa/src/cudapoa_batch.cpp
tijyojwad: Changed it to `int32_t`

PR #63, opened 2019-08-06
[common] move device_buffer class to utils

ahehn-nv:


ahehn-nv:
rerun tests

common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -48,7 +48,7 @@ public:
 
     ~device_buffer()
     {
-        CGA_CU_CHECK_ERR(cudaFree(data_));
```
mimaric: Why is this check not necessary?

ahehn-nv: This check would throw an exception in the destructor. Removing this check should just postpone the error detection to the next API call.

ahehn-nv: Does this answer your question?

mimaric: Yes, that makes sense. It will make it slightly harder to debug the code, but it's better than throwing in the destructor.

common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -32,26 +31,24 @@ public:
...
     using value_type = T;
-    device_storage() = delete;
-    device_storage(size_t n_elements, int32_t device_id)
+    device_buffer()  = delete;
+    device_buffer(size_t n_elements, int32_t device_id)
         : size_(n_elements)
-        , device_id_(device_id)
     {
-        CGA_CU_CHECK_ERR(cudaSetDevice(device_id_));
+        CGA_CU_CHECK_ERR(cudaSetDevice(device_id));
```
mimaric: Regarding ease of debugging... Something I experienced in my implementation and that will happen here as well is that only this line is reported, but not the line that actually wanted to make the allocation, meaning one has to set a breakpoint and check the stack trace in order to know where does the call actually comes from. I don't see an easy way to work around this this though...

common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -32,26 +31,24 @@ public:
...
 template <typename T>
-class device_storage
+class device_buffer
 {
 public:
     using value_type = T;
-    device_storage() = delete;
-    device_storage(size_t n_elements, int32_t device_id)
+    device_buffer()  = delete;
+    device_buffer(size_t n_elements, int32_t device_id)
```
akkamesh: Can we drop `device_id` argument from the constructor? Most use cases have consecutive/multiple calls to `device_buffer` and calling `cudaSetDevice` for each one of them seems unnecessary. 
We should let the caller set the context before calling these methods.

common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -32,26 +31,24 @@ public:
 };
 
 template <typename T>
-class device_storage
+class device_buffer
 {
 public:
     using value_type = T;
```
akkamesh: `value_type` is unused, can be removed.

common/utils/include/claragenomics/utils/device_buffer.cuh
ahehn-nv: Sure, makes sense. Originally, I added it (and kept it as a data member) because I thought you have to ensure you are on the right device before calling cudaFree.
So, you are right, since that's gone there is no added value in having that. Will remove it.

common/utils/include/claragenomics/utils/device_buffer.cuh
ahehn-nv: This something that concerns the implementation of CGA_CU_CHECK_ERR in general, not this class, right?

common/utils/include/claragenomics/utils/device_buffer.cuh
ahehn-nv: I added the `value_type` to ease writing generic functions. However, yes, the interface of the class is cleaner if I use `value_type` in the return types instead of `T`.

common/utils/include/claragenomics/utils/device_buffer.cuh
mimaric: Sure, but when `CGA_CU_CHECK_ERR` is used in the usual way (e.g. wrapping a `cudaMemcpy`) this is not an issue. Once a level of indirection is added (like in this case) this becomes an issue.
As said, I just wanted to point it out, I have no good idea how to fix it.
@akkamesh @tijyojwad

ahehn-nv:
I disentangled the changes in cudaaligner from the `device_buffer` class and incorporated the requested changes in a new PR (#64). Closing this PR.

common/utils/include/claragenomics/utils/device_buffer.cuh
ahehn-nv: I would use boost::stacktrace for this, but boost would be a pretty heavy dependency.

PR #69, opened 2019-08-08
[cudaaligner] Use device_buffer instead of device_storage

ahehn-nv:
* Use common `device_buffer<T>` in cudaaligner and remove `device_storage<T>` class.
* Use `scoped_device_switch` in cudaaligner.
* Remove `device_id` from batched_device_matrices


tijyojwad:
rerun tests

PR #74, opened 2019-08-09
[cudaaligner] added missing virtual and noexcept to destructors

ahehn-nv:


tijyojwad:
@ahehn-nv looks good to me. Can you explain the value of `noexcept` in this change? If it is to declare explicitly that the destructors don't throw any exception, where can that information be used?

ahehn-nv:
This change was mostly about clear semantics and a bit about optimization. However, in the meantime, I learned that destructors are (in almost all cases) `noexcept` by default since C++11. So there is no need to be more verbose about it. Closing.

PR #75, opened 2019-08-13
[common] add a device_memset_async for device_buffer

ahehn-nv:
* [common] device_buffer added initialize() method
* [cudaaligner] AlignerGlobal: initialize device buffers

### DIFF
--- a/common/utils/include/claragenomics/utils/device_buffer.cuh
+++ b/common/utils/include/claragenomics/utils/device_buffer.cuh
@@ -13,6 +13,10 @@
 #include <claragenomics/utils/cudautils.hpp>
 #include <exception>
 #include <utility>
+#include <cassert>
+#ifndef NDEBUG
+#include <limits>
+#endif
 
 namespace claragenomics
 {
@@ -76,5 +80,12 @@ private:
     value_type* data_ = nullptr;
     size_t size_      = 0;
 };
+
+template <typename T>
+void device_memset_async(device_buffer<T>& buffer, int value, cudaStream_t stream)
+{
+    assert(value <= std::numeric_limits<unsigned char>::max());
+    CGA_CU_CHECK_ERR(cudaMemsetAsync(buffer.data(), value, sizeof(typename device_buffer<T>::value_type) * buffer.size(), stream));
+}
 
 } // end namespace claragenomics
--- a/cudaaligner/src/aligner_global.cpp
+++ b/cudaaligner/src/aligner_global.cpp
@@ -44,15 +44,19 @@ AlignerGlobal::AlignerGlobal(int32_t max_query_length, int32_t max_target_length
     , stream_(stream)
     , device_id_(device_id)
 {
-    scoped_device_switch dev(device_id);
-    sequences_d_        = device_buffer<char>(2 * std::max(max_query_length, max_target_length) * max_alignments);
-    sequence_lengths_d_ = device_buffer<int32_t>(2 * max_alignments);
-    results_d_          = device_buffer<int8_t>(calc_max_result_length(max_query_length, max_target_length) * max_alignments);
-    result_lengths_d_   = device_buffer<int32_t>(max_alignments);
     if (max_alignments < 1)
     {
         throw std::runtime_error("Max alignments must be at least 1.");
     }
+    scoped_device_switch dev(device_id);
+    sequences_d_        = device_buffer<char>(sequences_h_.size());
+    sequence_lengths_d_ = device_buffer<int32_t>(sequence_lengths_h_.size());
+    results_d_          = device_buffer<int8_t>(results_h_.size());
+    result_lengths_d_   = device_buffer<int32_t>(result_lengths_h_.size());
+    device_memset_async(sequences_d_, 0, stream);
+    device_memset_async(sequence_lengths_d_, 0, stream);
+    device_memset_async(results_d_, 0, stream);
+    device_memset_async(result_lengths_d_, 0, stream);
 }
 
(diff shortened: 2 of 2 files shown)


common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -68,6 +68,11 @@ public:
         cudaFree(data_);
     }
 
+    void initialize(cudaStream_t stream)
```
mimaric: Passing a stream should be enough to indicate that this is an async operation, but it feels like `initialize_async` would be it even clearer.

ahehn-nv: yes, makes sense. I changed it from `initialize(stream)` to `initialize_async(stream)`. Thanks!

mimaric:
Now that I thought a bit more about it... How about adding a `bool initialize = false` (or an internal enum class) argument to the constructor? Initialization is tightly related to construction and if someone wants to set the array to zero after already doing something with the array they should do it on their own, no need for the class to provide a method for it

ahehn-nv:
I thought about that, but decided against it, because it would make the semantics less clear in my opinion. Right now, the `device_buffer` just provides you bare device memory - period. You have to take care of initializing it (either manually or by calling `initialize_async()`).
If you add an `initialize` argument to the constructor it will be less clear if you can expect the buffer to be initialized or not (at least if you provide a default value), and the meaning of passing the stream to the constructor is not obvious.
I considered `device_buffer buf(size, async_initializer(stream))` for a moment, but `buf.initialize_async(stream)` seemed the easier solution for what I needed.

Maybe the name `initialize` is not ideal? `memset_async(value,stream)` would be more general and maybe a bit clearer...

mimaric:
Yes, I also thought about the ugliness of such constructor...
The more I think about it the more I wonder if we need something like this in `device_buffer` at all. I think we should keep it all simple as possible. If we would add something like this why not add `copy_data_to` and `copy_data_from` as well?
@tijyojwad @akkamesh @vellamike What do you think?

ahehn-nv:
I just wanted to spare myself the lengthy ugly
```
cudaMemset(sequences_d_.data(), 0, sequences_d_.size()*sizeof(char), stream);
cudaMemset(sequence_lengths_d_.data(), 0, sequence_lengths_d_.size()*sizeof(int32_t), stream);
cudaMemset(results_d_.data(), 0, results_d_.size()*sizeof(int8_t), stream);
cudaMemset(result_lengths_d_.data(), 0, result_lengths_d_.size()*sizeof(int32_t), stream);
```
But good point with `copy_data_to/from`...

mimaric:
I'm rather undecided on this one... Let's hear some more opinions...

ahehn-nv:
Yes, me too. 
Yet another possibilty: I could also just add a
```
template <typename T>
device_memset_async(device_buffer<T>& buf, int32_t byte, cudaStream_t stream)
{
    cudaMemsetAsync(buf.data(), byte, sizeof(T)*buf.size(), stream);
}
```
either publicly to the `device_buffer.cuh` to decouple it from the class a bit, or just as utility to the `aligner_global.cpp`. But let's wait for more opinions...

tijyojwad:
@ahehn-nv @mimaric I think it should be part of the `device_buffer` class itself, with an API like `memset_async(byte, stream)`. As for the other functions, it indeed would be nice to have the `copy_to` and `copy_from` APIs too, but I want to be wary of effectively re-writing the thrust `device_vector` APIs which we want to move to soon :)

mimaric:
I would say there are three options:
1) Do nothing
   + keeps `device_buffer` lean
   + `CGA_CU_CHECK_ERR` always reports the line on which memset was done
   - Much typing, hard to read, easy to make mistake
2) Standalone function
   + keeps `device_buffer` lean
   + easy to use
   - `CGA_CU_CHECK_ERR` reports the like withing that function, not where memset was actually done
3) Add method to `device_buffer`
   + easy to use
   +- `device_buffer` interface grows
   - `CGA_CU_CHECK_ERR` reports the like withing that method, not where memset was actually done

I do agree with @tijyojwad that we might be reimplementing `thrust::device_vector`

ahehn-nv:
I slept a night on it and I disagree with @tijyojwad .
I think the `device_buffer` should remain the way it is - minimal - just providing device memory. memset and copies have nothing to do with this functionality. These operations do not manipulate the internal state of the object (the pointer and size), so there is no need to make them methods.

However, I would like to add the previously mentioned convenience function to `device_buffer.cuh`, which uses the existing interface of `device_buffer`:
```
template <typename T>
device_memset_async(device_buffer<T>& buf, int32_t byte, cudaStream_t stream)
{
    cudaMemsetAsync(buf.data(), byte, sizeof(T)*buf.size(), stream);
}
```
It mimics the underlying CUDA call and if we move away from `device_buffer`, this function can also easily be adapted/overloaded.
Any thoughts?

ahehn-nv:
@mimaric and my comments crossed. So, in my comment I propose to go with @mimaric's option 2.
The `CGA_CU_CHECK_ERR` argument is not very helpful in my opinion. If you follow this through it means you should write spaghetti code, doesn't it? I would not expect `CGA_CU_CHEC_ERR` to give you the full error information. If something goes sideways, you should take a debugger to have a closer look and obtain a stacktrace there (with asynchronous execution it might not that helpful anyways).

mimaric:
I agree. Let's go for a standalone function!

ahehn-nv:
I pushed my new proposal. Please review.

ahehn-nv:
@tijyojwad do you have further thoughts on this?

tijyojwad:
@ahehn-nv I'm on board with keeping the class minimal too. Go ahead and merge

PR #78, opened 2019-08-14
[cudaaligner] Hirschberg + Myers algorithm

ahehn-nv:


cudaaligner/src/batched_device_matrices.cuh
```diff
@@ -23,9 +23,9 @@ namespace claragenomics
 namespace cudaaligner
 {
 
-__device__ inline bool error(int t)
+__device__ inline bool error(int x, int y)
```
tijyojwad: to be consistent, let's change the `int` to `int32_t`?

tijyojwad:
@ahehn-nv can you also change the class instantiated in`aligner.cpp` to be HirschbergMyers?

cudaaligner/src/batched_device_matrices.cuh
ahehn-nv: good point. fixed.

ahehn-nv:
Yes, changed the default aligner for global alignment to HirschbergMyers.

PR #80, opened 2019-08-14
[README] add min requirements

tijyojwad:


tijyojwad:
README changes don't trigger tests, so merging directly.

PR #83, opened 2019-08-16
[cudamapper] New chaining algorithm

vellamike:
* `OverlapperTriggered` Overlapper provides more sophisticated chaining than previous naive approach
* Fixed several issues with PAF output correctness
* Fixed bug in PAF evaluation
* Temporarily changed default index generation back to CPU as I was encountering a compilation and need assistance from @mimaric to resolve it.


cudamapper/src/overlapper.cpp
```diff
@@ -12,17 +12,31 @@
...
+        }
+        return filtered_overlaps;
+    }
+
     void Overlapper::print_paf(const std::vector<Overlap> &overlaps){
+        std::vector<Overlap> filtered_overlaps = filter_overlaps(overlaps);
+
         std::string relative_strand = "+";
-        for(const auto& overlap: overlaps){
+        for(const auto& overlap: filtered_overlaps){
```
mimaric: This is a performance vs clean code discussion, but how about moving the functionality of `filter_overlaps` here and avoiding building a completely new vector?

cudamapper/src/overlapper_triggered.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+#include <vector>
+
+#include "cudamapper/types.hpp"
+#include "cudamapper/overlapper.hpp"
+#include "matcher.hpp"
+
+namespace claragenomics {
+
+    /// OverlapperNaive - generates overlaps and displays them on screen. This overlapper uses a greedy approach where
```
mimaric: Copy pasted comment from `OverlapperNaive`?

cudamapper/src/overlapper_triggered.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include "overlapper_triggered.hpp"
+#include "cudamapper/overlapper.hpp"
+#include "matcher.hpp"
+
+namespace claragenomics {
+    std::vector<Overlap> OverlapperTriggered::fuse_overlaps(std::vector<Overlap> unfused_overlaps){
+        std::vector<Overlap> fused_overlaps;
+        Overlap fused_overlap = unfused_overlaps[0];
+        for (size_t i=0; i< unfused_overlaps.size() - 1; i++){
+            Overlap next_overlap = unfused_overlaps[i+1];
```
mimaric: `const Overlap&`

cudamapper/src/overlapper_triggered.cpp
```diff
@@ -0,0 +1,128 @@
...
+#include <string>
+#include <vector>
+#include <unordered_map>
+
+#include "overlapper_triggered.hpp"
+#include "cudamapper/overlapper.hpp"
+#include "matcher.hpp"
+
+namespace claragenomics {
+    std::vector<Overlap> OverlapperTriggered::fuse_overlaps(std::vector<Overlap> unfused_overlaps){
```
mimaric: please add a short comment explaining what the algorithm is doing

cudamapper/src/overlapper_triggered.hpp
```diff
@@ -0,0 +1,40 @@
...
+        /// \param anchors vector of anchors
+        /// \param index Index
+        /// \return vector of Overlap objects
+        const std::vector<Overlap> get_overlaps(const std::vector<Anchor> &anchors, const Index &index) override;
+
+        /// \brief given a vector of overlaps, combines all overlaps from the same read pair
+        ///
+        /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
+        /// \return vector of overlaps
+        std::vector<Overlap> fuse_overlaps(std::vector<Overlap> unfused_overlaps);
```
mimaric: Does this method have to be public?

cudamapper/src/overlapper_triggered.cpp
```diff
@@ -0,0 +1,128 @@
...
+                fused_overlap.target_end_position_in_read_ = next_overlap.target_end_position_in_read_;
+            } else {
+                fused_overlaps.push_back(fused_overlap);
+                fused_overlap = unfused_overlaps[i+1];
+            }
+        }
+        return fused_overlaps;
+    }
+
+    std::vector<Overlap> const OverlapperTriggered::get_overlaps(const std::vector<Anchor> &anchors, const Index &index) {
```
mimaric: please add some more clarification here as well

cudamapper/src/index_generator.cpp
```diff
@@ -10,9 +10,11 @@
 
 #include "bioparser/bioparser.hpp"
 #include "index_generator_gpu.hpp"
+#include "index_generator_cpu.hpp"
+
 
 namespace claragenomics {
     std::unique_ptr<IndexGenerator> IndexGenerator::create_index_generator(const std::string &query_filename, std::uint64_t kmer_length, std::uint64_t window_size) {
-        return std::make_unique<IndexGeneratorGPU>(query_filename, kmer_length, window_size);
+        return std::make_unique<IndexGeneratorCPU>(query_filename, kmer_length, window_size);
```
mimaric: Let's discuss the issues you have before switching back to `IndexGeneratorCPU`

cudamapper/src/overlapper_naive.cpp
```diff
@@ -36,12 +37,14 @@ namespace claragenomics {
...
                 new_overlap.query_read_name_ = read_names[anchor.query_read_id_];
+                new_overlap.query_length_ = read_lengths[anchor.query_read_id_];
 
                 new_overlap.target_read_id_ = anchor.target_read_id_;
                 new_overlap.target_read_name_ = read_names[anchor.target_read_id_];
+                new_overlap.target_length_ = read_lengths[anchor.target_read_id_];
 
                 new_overlap.query_start_position_in_read_ = anchor.query_position_in_read_;
-                new_overlap.target_start_position_in_read_ = anchor.query_position_in_read_;
+                new_overlap.target_start_position_in_read_ = anchor.target_position_in_read_; // TODO seems like an error?
```
mimaric: Is this TODO still valid?

cudamapper/include/cudamapper/index.hpp
```diff
@@ -33,6 +33,10 @@ namespace claragenomics {
         /// \return an array of directions in which sketch elements were read
         virtual const std::vector<SketchElement::DirectionOfRepresentation>& directions_of_reads() const = 0;
 
+        /// \brief returns an array of directions in which sketch elements were read
```
mimaric: Fix the comment

cudamapper/src/index_generator_cpu.hpp
```diff
@@ -43,6 +43,8 @@ namespace claragenomics {
         /// \return mapping of internal read id that goes from 0 to number_of_reads-1 to actual read name from the input
         const std::vector<std::string>& read_id_to_read_name() const override;
 
+        const std::vector<uint32_t >& read_id_to_read_length() const override;
```
mimaric: Add tests for this method

cudamapper/src/overlapper_triggered.hpp
```diff
@@ -0,0 +1,40 @@
...
+#include "cudamapper/types.hpp"
+#include "cudamapper/overlapper.hpp"
+#include "matcher.hpp"
+
+namespace claragenomics {
+
+    /// OverlapperNaive - generates overlaps and displays them on screen. This overlapper uses a greedy approach where
+    /// very simple chaining is used - if two strings have more than one anchor this is considered to be an overlap.
+    /// The extend of the overlap is the two anchors which are furthest apart.
+    class OverlapperTriggered: public Overlapper {
```
mimaric: Add unittests for this class

cudamapper/src/main.cpp
```diff
@@ -27,27 +28,27 @@ int main(int argc, char *argv[])
     // TODO: pass kmer and window size as parameters
     std::unique_ptr<claragenomics::IndexGenerator> index_generator = claragenomics::IndexGenerator::create_index_generator(std::string(argv[1]), 15, 15);
     CGA_LOG_INFO("Created index generator");
-    std::cout << "Index generator execution time: " << std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time).count() << "ms" << std::endl;
+    std::cerr << "Index generator execution time: " << std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time).count() << "ms" << std::endl;
```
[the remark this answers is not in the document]
vellamike: I was thinking about this, but we don't have a verbose flag yet and if I use `CGA_LOG_INFO` it will only work for Release builds, which is not what we want for this instrumentation code. An alternative is to use `CGA_LOG_ERROR` but this seems wrong since I'm not reporting an error.

cudamapper/src/overlapper_triggered.hpp
vellamike: yes! fixed.

cudamapper/src/overlapper_triggered.cpp
vellamike: This is documented in `overlapper_triggered.hpp`. I now added more documentation and an example to the declaration.

cudamapper/src/overlapper_triggered.hpp
vellamike: No, good point. fixed

cudamapper/src/overlapper_naive.cpp
vellamike: No. fixed

cudamapper/src/overlapper_triggered.cpp
vellamike: Fixed

cudamapper/src/overlapper.cpp
vellamike: another option is to remove the constnesss of `overlaps` and have `filter_overlaps` do an in-place filtering? IMO the performance impact of this is likely so negligible that we should leave it as is for now.

cudamapper/src/index_generator_cpu.hpp
vellamike: done

cudamapper/include/cudamapper/index.hpp
vellamike: done

cudamapper/src/overlapper.cpp
```diff
@@ -12,17 +12,31 @@
 #include "index_cpu.hpp"
 
 namespace claragenomics {
+
+    std::vector<Overlap> Overlapper::filter_overlaps(const std::vector<Overlap> &overlaps,size_t  min_residues, size_t  min_overlap_len) {
+        std::vector<Overlap> filtered_overlaps;
+        for(auto overlap: overlaps){
```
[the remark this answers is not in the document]
vellamike: That's nice. I'm not convinced that this is cleaner code but I guess it's a matter of taste - changed it regardless.

cudamapper/src/overlapper_triggered.cpp
vellamike: Added docs to `overlapper_triggered.hpp`

cudamapper/src/main.cpp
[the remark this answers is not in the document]
mimaric: I agree. Let's keep it this was for now.

cudamapper/src/overlapper.cpp
mimaric: Yes, doing it in-place would be an option. Have a look at https://en.wikipedia.org/wiki/Erase%E2%80%93remove_idiom

mimaric: Or as you suggest in another discussion you can make another function/class/lambda which returns next valid element (i.e. the one which would not be filtered out) on each call. This would actually probably have the best performance and lowest memory consumption

vellamike: @mimaric I suggest a separate PR for this, what do you think?

mimaric: Yes, let's keep it like it is for now

cudamapper/src/overlapper_triggered.hpp
vellamike: done

cudamapper/src/index_cpu.hpp
```diff
@@ -60,6 +60,8 @@ namespace claragenomics {
         /// \return mapping of internal read id that goes from 0 to number_of_reads-1 to actual read name from the input
         const std::vector<std::string>& read_id_to_read_name() const override;
 
+        const std::vector<std::uint32_t>& read_id_to_read_length() const override;
```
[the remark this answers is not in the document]
vellamike: done

PR #85, opened 2019-08-21
[cga] Added free() method to device_buffer

mimaric:


common/utils/include/claragenomics/utils/device_buffer.cuh
```diff
@@ -67,6 +81,21 @@ public:
         return *this;
     }
 
+    void free()
+    {
+        cudaError_t err = cudaFree(data_);
```
ahehn-nv: Shouldn't the regular `CGA_CU_CHECK_ERR()` suffice here? As far as I understand (not 100% sure) `cudaFree()` itself will only fail with `cudaErrorInvalidValue` if you pass it an address which was not allocated with `cudaMalloc` (or 0), which is guaranteed not to happen by the class, or a serious memory/driver/hardware corruption from which you will not be able to recover.

Actually, I have no idea how I would recover from a failed deallocation either... I could ignore it, but is there anything else I could do?

mimaric: Makes sense. Let me update that

PR #90, opened 2019-08-23
[cudamapper] Vector instead of map in Index<->Matcher interface

mimaric:
Using `std::vector` instead of `std::map` in Index<->Matcher interface.
This change has a huge performance benefit.
@akkamesh FYI


vellamike:
Could you have a look at the CI failure? It looks like a linting issue which you can regnerate with `make doc`

PR #92, opened 2019-08-28
[packaging] add basic deb packaging

tijyojwad:


cmake/Packaging.cmake
```diff
@@ -0,0 +1,15 @@
...
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+SET(CPACK_GENERATOR "DEB")
+SET(CPACK_DEBIAN_PACKAGE_MAINTAINER "NVIDIA Corporation")
+set(CPACK_PACKAGE_VERSION "${CGA_VERSION}")
```
vellamike: uppercase

README.md
```diff
@@ -40,6 +42,20 @@ cmake .. -DCMAKE_BUILD_TYPE=Release -DCMAKE_INSTALL_PREFIX=install
...
+
+```bash
+make install
+```
+
+### Debian Package
+To generate a Debian package for the SDK -
+
+```bash
+make package
```
vellamike: It's not clear from this how to do a build for a different CUDA toolkit version?

tijyojwad:
@vellamike the error above is coming from the `make install` command. looks like `docs` wasn't built, so install is failing since it's trying to copy an `html` folder. Let me fix that separately (doc generation shouldn't be required for installation)

README.md
tijyojwad: The packaging is independent of the toolkit version. the packaging just puts the build libraries/headers into a deb package.
To build for a different cuda version cmake would need to be run with -DCUDA_TOOLKIT_DIR (i.e. the cmake command for setting a custom toolkit folder)

vellamike: Could we clarify this in the Readme?

vellamike:
> @vellamike the error above is coming from the `make install` command. looks like `docs` wasn't built, so install is failing since it's trying to copy an `html` folder. Let me fix that separately (doc generation shouldn't be required for installation)

OK - let me know when it's fixed, I'd like to build the debs and test locally before merging since this is an important PR.

tijyojwad:
@vellamike Doc generation fixed in #125

README.md
tijyojwad: Yep, clarified that it only packages what is built by the `make` command and doesn't automatically support cross platform/CUDA support

PR #99, opened 2019-08-30
[cudamapper] Removed IndexGeneratorCPU

mimaric:
`IndexGeneratorCPU` has a significantly worse performance than `IndexGeneratorGPU`. As having a GPU is a requirement for running cudamapper (for example `Matcher` has only GPU implementation) we decided to remove this component in order to keep the project simpler

In addition, added virtual destructors to public interfaces


mimaric:
rerun tests

vellamike:
rerun tests

PR #105, opened 2019-09-03
[cudamapper] Minimizer-specific parts of index generation moved to Minimizer

mimaric:
Large parts of index generation are minimizer-specific and would not hold for other SketchElement implementation. This code was moved to `Minimizer::generate_sketch_elements()`.

`Minimizer::generate_sketch_elements()` is referenced directly from `IndexGeneratorGPU`, which is not a good design (`IndexGeneratorGPU` should be able to accept any implementation of `SketchElement`). This will be fixed in the next PR which will make `IndexGenerator` a template class which accepts `SketchElement` implementation as template parameter.

In addition `Minimizer::kmer_to_representation` has been removed as it was only needed for `IndexGeneratorCPU`


mimaric:
rerun tests

PR #116, opened 2019-09-03
[cudapoa] make api functions device neutral

### LINKED ISSUE #113: Make SDK functions "current device"-neutral
If we assume our users use CUDA also outside of our library, we should also ensure that our methods are "current device"-neutral, i.e. that we reset the device (cudaSetDevice) at the end of each method to the value it had when it entered the method.

tijyojwad:
Making `cudapoa` API device neutral

Closes #113

### DIFF
--- a/cudapoa/src/cudapoa_batch.cpp
+++ b/cudapoa/src/cudapoa_batch.cpp
@@ -91,10 +91,11 @@ CudapoaBatch::CudapoaBatch(int32_t max_sequences_per_poa,
                                   cuda_banded_alignment))
     , max_poas_(batch_block_->get_max_poas())
 {
+    scoped_device_switch dev(device_id_);
+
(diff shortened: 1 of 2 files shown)


cudapoa/src/cudapoa_batch.cpp
```diff
@@ -91,10 +91,11 @@ CudapoaBatch::CudapoaBatch(int32_t max_sequences_per_poa,
                                   cuda_banded_alignment))
     , max_poas_(batch_block_->get_max_poas())
 {
+    scoped_device_switch dev(device_id_);
+
     bid_ = CudapoaBatch::batches++;
 
     // Set CUDA device
```
ahehn-nv: this comment should be removed, too...

tijyojwad: Moved the comment to the scoped device switch call

PR #117, opened 2019-09-03
[utils] updating CGA_CU_CHECK_ERR to abort

### LINKED ISSUE #96: revert CGA_CU_CHECK_ERR to abort on error
Revert the CGA_CU_CHECK_ERR functionality to abort on error for Release builds, and `assert(false)` and then abort for Debug builds.

Potentially add cudaDeviceSynchronize in debug builds to catch errors when they occur

tijyojwad:
Updates CGA_CU_CHECK_ERR function to abort on error
Adds optional cudaDeviceSynchronize to underlying checking function

Closes #96

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -15,6 +15,7 @@ project(ClaraGenomicsAnalysisCPP)
 option(cga_enable_tests "Build ClaraGenomicsAnalysis unit tests" OFF)
 option(cga_enable_benchmarks "Build ClaraGenomicsAnalysis benchmarks" OFF)
 option(cga_build_shared "Build ClaraGenomicsAnalysis libraries as shared objects" OFF)
+option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHECK_ERR calls" OFF)
 # The spoa_accurate option runs a different (and slower) version of
 # the topological sort in cudapoa which exactly matches the output
 # of the topological sort implementation in the SPOA library. This is
@@ -43,14 +44,13 @@ else()
     set_property(GLOBAL PROPERTY cga_library_type STATIC)
 endif()
 
+include(cmake/CXX.cmake)
 include(cmake/CUDA.cmake)
 include(cmake/Doxygen.cmake)
 include(cmake/3rdparty.cmake)
 include(cmake/Tests.cmake)
 include(cmake/Benchmarks.cmake)
 include(cmake/Format.cmake)
-
-set(CMAKE_CXX_STANDARD 17)
 
 # Add ClaraGenomicsAnalysis projects.
 add_subdirectory(common/logging)
--- a/common/utils/include/claragenomics/utils/cudautils.hpp
+++ b/common/utils/include/claragenomics/utils/cudautils.hpp
@@ -16,7 +16,7 @@
 #include <claragenomics/logging/logging.hpp>
 
 #include <cuda_runtime_api.h>
-#include <stdexcept>
+#include <cassert>
 
 /// \ingroup cudautils
 /// \{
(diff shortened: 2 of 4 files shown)


common/utils/include/claragenomics/utils/cudautils.hpp
```diff
@@ -43,24 +43,28 @@ namespace cudautils
...
 /// \param line File line number of the calling function
-/// \param abort If true, hard-exit on CUDA error
-inline void gpu_assert(cudaError_t code, const char* file, int line, bool abort = false)
+/// \param cuda_sync Synchronize CUDA device before checking error
+inline void gpu_assert(cudaError_t code, const char* file, int line, bool cuda_sync = false)
 {
+    // This device synchronize forces the most recent CUDA call to fully
+    // complete, increasing the chance of catching the CUDA error near the
+    // offending function.
+    if (cuda_sync)
```
ahehn-nv: `cudaDeviceSynchronize()` could potentially overwrite a previous `code` value. I think changing this if to `cuda_sync && code == cudaSuccess)` should fix the problem.

common/utils/include/claragenomics/utils/cudautils.hpp
```diff
@@ -43,24 +43,28 @@ namespace cudautils
 /// \param code The CUDA status code of the function being asserted
 /// \param file Filename of the calling function
 /// \param line File line number of the calling function
-/// \param abort If true, hard-exit on CUDA error
-inline void gpu_assert(cudaError_t code, const char* file, int line, bool abort = false)
+/// \param cuda_sync Synchronize CUDA device before checking error
+inline void gpu_assert(cudaError_t code, const char* file, int line, bool cuda_sync = false)
```
ahehn-nv: Do we need/want `cuda_sync` as runtime parameter? Is there a likely use case, where we will call `gpu_assert` with `cuda_sync` set in some places, but not in others?

I guess, I would introduce macro `CGA_DEVICE_SYNCHRONIZE` to switch on the device sync. Similar to `NDEBUG`, such that there is an easy way to switch it on when building the project.

ahehn-nv:
Furthermore, `#include <stdexcept>` is no longer needed, but now we need `#include <cassert>`.

tijyojwad:
> Furthermore, `#include <stdexcept>` is no longer needed, but now we need `#include <cassert>`.

Yep, CI caught this too :).

common/utils/include/claragenomics/utils/cudautils.hpp
tijyojwad: Wouldn't the last bad error code stick for `cudaDeviceSynchronize()`? I always assumed that was the behavior of that call, and that error states in the driver couldn't be overwritten. But for safety, let me change it.

common/utils/include/claragenomics/utils/cudautils.hpp
tijyojwad: > Is there a likely use case, where we will call gpu_assert with cuda_sync set in some places, but not in others?

The use case I had in mind for this was enabling it say for `cudapoa` but not `cudaaligner` when running racon to get to the poa parts faster, since the device synchronize slows down execution.

I can go the macro route as well. That would not require any code changes to enable either, so more user friendly too

common/utils/include/claragenomics/utils/cudautils.hpp
ahehn-nv: The more serious errors will, but if you get an out-of-memory from a cudaMalloc for example, it will not reappear when you call `cudaDeviceSynchronize()`.

common/utils/include/claragenomics/utils/cudautils.hpp
ahehn-nv: Ok, but then we would need to generalize the `CGA_CU_CHECK_ERROR` macro, right?

common/utils/include/claragenomics/utils/cudautils.hpp
tijyojwad: Hmm good point, that would indeed be nasty if an our of memory error got undetected. Pushed the changed

common/utils/include/claragenomics/utils/cudautils.hpp
tijyojwad: Yes it would, and I missed that :). Changed it to macro, if it proves bothersome will change back to runtime parameter later

CMakeLists.txt
```diff
@@ -15,6 +15,7 @@ project(ClaraGenomicsAnalysisCPP)
 option(cga_enable_tests "Build ClaraGenomicsAnalysis unit tests" OFF)
 option(cga_enable_benchmarks "Build ClaraGenomicsAnalysis benchmarks" OFF)
 option(cga_build_shared "Build ClaraGenomicsAnalysis libraries as shared objects" OFF)
+option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHECK_ERR calls" OFF)
```
ahehn-nv: Ideally I wouldn't make this a separate option, but just add it the default Debug mode flags. However, I don't know how to do that (it seems cumbersome), so I guess this is the best solution.

PR #121, opened 2019-09-05
[cudamapper] Merged IndexGenerator and Index into Index

mimaric:


cudamapper/include/cudamapper/index.hpp
```diff
@@ -65,7 +68,7 @@ namespace claragenomics {
 
         /// \brief generates a mapping of (k,w)-kmer-representation to all of its occurrences for one or more sequences
         /// \return instance of Index
-        static std::unique_ptr<Index> create_index(const IndexGenerator& index_generator);
+        static std::unique_ptr<Index> create_index(const std::string& query_filename, const std::uint64_t minimizer_size, const std::uint64_t window_size);
```
vellamike: should we stop using the "minimzier" word here?

mimaric: Yes. I'm chaining that in a PR I'm working on right now

PR #126, opened 2019-09-06
[pyclaragenomics] Fixed readme

mimaric:


pyclaragenomics/README.md
```diff
@@ -4,6 +4,8 @@ Python libraries and utilities for manipulating genomics data
 
 ## Installation
 
+Make sure `ClaraGenomicsAnalysis/pyclaragenomics/cga_build/install/lib/` is added to `LD_LIBRARY_PATH`
```
vellamike: I don't think this comment should be in here, we need to figure out what was causing this issue and solve it.

PR #130, opened 2019-09-11
[common] clear CUDA runtime error in device_buffer before exception is thrown

ahehn-nv:
Clear the CUDA runtime error in device_buffer before an out-of-memory exception is thrown.

### DIFF
--- a/common/utils/include/claragenomics/utils/device_buffer.cuh
+++ b/common/utils/include/claragenomics/utils/device_buffer.cuh
@@ -48,7 +48,16 @@ public:
     {
         cudaError_t err = cudaMalloc(reinterpret_cast<void**>(&data_), size_ * sizeof(T));
         if (err == cudaErrorMemoryAllocation)
+        {
+            // Clear the error from the runtime...
+            err = cudaGetLastError();
+            // Did a different async error happen in the meantime?
+            if (err != cudaErrorMemoryAllocation)
+            {
+                CGA_CU_CHECK_ERR(err);
+            }
             throw device_memory_allocation_exception();
+        }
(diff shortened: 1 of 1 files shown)


mimaric:
Can you provide more background on this? Why is this check not needed anymore?

ahehn-nv:
I'm not removing any checks, just clearing the error state of the CUDA runtime after the error is registered and then handled by throwing an exception afterwards.
I was hoping the comments would explain what I intend to do...
Anyhow, the background is the following:
If `cudaMalloc()` fails, it will return `cudaErrorMemoryAllocation` and the device_buffer will throw an exception. However, the CUDA runtime will still keep the error state `cudaErrorMemoryAllocation` and a subsequent call to `cudaPeekAtLastError()` will still return this error. So, if we manage to handle the out-of-memory exception somehow and continue running, the program will abort after the next kernel launch, since the `CGA_CU_CHECK_ERR(cudaPeekAtLastError());` right after the kernel launch will still see the `cudaErrorMemoryAllocation`.

This fix calls `cudaGetLastError();` after a failed `cudaMalloc()`, but before throwing the exception, such that the error is cleared from the CUDA runtime.
There's a small twist to it: since an asynchronous CUDA operation could fail between the `cudaMalloc()` and the `cudaGetLastError()`, I double check if `cudaGetLastError()` still reports `cudaErrorMemoryAllocation`. If it does, I throw the exception, if it doesn't, because some asynchronous error happened, I pass the error to the regular `CGA_CU_CHECK_ERR`.

mimaric:
Oh, sorry.... I read "clar" and totally ignored the fact that the background behind the changes was green, not red :(
Makes sense!

Issue #106, opened 2019-09-03
Sample app for cuda aligner


tijyojwad:
Write a sample application for cuda aligner

PR #132, opened 2019-09-11
[pycga] adding cudaaligner python API

tijyojwad:
Add python API for cudaaligner module

### DIFF
--- a/cudaaligner/benchmarks/singlealignment/main.cpp
+++ b/cudaaligner/benchmarks/singlealignment/main.cpp
@@ -33,7 +33,7 @@ static void BM_SingleAlignment(benchmark::State& state)
     std::unique_ptr<Aligner> aligner = create_aligner(genome_size,
                                                       genome_size,
                                                       1,
-                                                      AlignmentType::global,
+                                                      AlignmentType::global_alignment,
                                                       0,
                                                       0);
     aligner->add_alignment(genome_1.c_str(), genome_1.length(),
--- a/cudaaligner/include/claragenomics/cudaaligner/aligner.hpp
+++ b/cudaaligner/include/claragenomics/cudaaligner/aligner.hpp
@@ -51,7 +51,8 @@ public:
     /// on the GPU to come to completion.
     virtual StatusType sync_alignments() = 0;
 
-    /// \brief Add new alignment object
+    /// \brief Add new alignment object. Only strings with characters
+    ///        from the alphabet [ACGT] are guaranteed to provide correct results.
     ///
     /// \param query Query string
     /// \param query_length  Query string length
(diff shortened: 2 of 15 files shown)


tijyojwad:
Closes #132

vellamike:
A few general comments:

1. Could we add some samples to the sample directory?
2.  This code works:
```
device = cuda.cuda_get_device()
stream = cuda.CudaStream()
query = "CHCHCHCHACH"
target = "CHCHCHCHACH"
batch = CudaAlignerBatch(len(query), len(target), 1, alignment_type="global", stream=stream, device_id=device)
batch.add_alignment(query, target)
batch.align_all()
alignments = batch.get_alignments()
for a in alignments:
    print(a)
    print(a.cigar)
```

printing out:

```
11M2I
```

but it seems like this ought to be undefined behaviour - could we either prevent this or document this behaviour (apologies if its documented, I missed it)?


3. `print(a)` just prints `<claragenomics.bindings.cudaaligner.CudaAlignment object at 0x7fe2bde926d8>`. Could you implement the [`__repr__` and `__str__`](https://stackoverflow.com/questions/1436703/difference-between-str-and-repr) methods?

4. From [this test](https://github.com/clara-genomics/ClaraGenomicsAnalysis/pull/132/files#diff-8130e1603c3dd434666ca00b26d70243R32) it seems that all alignments need to be the same length - is this correct?

tijyojwad:
> Could we add some samples to the sample directory?

Yes, will create another PR for samples of cudaaligner (both C++ and python) - https://github.com/clara-genomics/ClaraGenomicsAnalysis/issues/106

> This code works:
device = cuda.cuda_get_device()
stream = cuda.CudaStream()
query = "CHCHCHCHACH"
target = "CHCHCHCHACH"
batch = CudaAlignerBatch(len(query), len(target), 1, alignment_type="global", stream=stream, device_id=device)
batch.add_alignment(query, target)
batch.align_all()
alignments = batch.get_alignments()
for a in alignments:
    print(a)
    print(a.cigar)
printing out:
11M2I
but it seems like this ought to be undefined behaviour - could we either prevent this or document this behaviour (apologies if its documented, I missed it)?

Undefined in terms of the alphabet supported? We don't really have any limitation on alphabet family size. I can add that as a note in our documentation

> print(a) just prints <claragenomics.bindings.cudaaligner.CudaAlignment object at 0x7fe2bde926d8>. Could you implement the __repr__ and __str__ methods?

I was initially thinking about this, but wasn't sure what to print for `__str__` method of alignment. THe CIGAR string? The formatted aligmnet? The actual alignment?

> From this test it seems that all alignments need to be the same length - is this correct?

All alignments don't have to be the same length. What can I change in the test to make that clearer?

Issue #133, opened 2019-09-13
After PR #131 results of cudamapper are different


mimaric:
For a 10 megabases 30x coverage input cudamapper returns different results than before the merge of #131 
Please investigate and unittest to cover that case.

PR #137, opened 2019-09-13
[cudamapper] Fixes issue #133

vellamike:
- When adding anchors to anchors vector iterator was not being updated
and old anchors were being overwritten. This commit fixes that issue.

- Changes max buffer size definition from GB to MiB, for better
granularity.


vellamike:
> Can you add a utittest which covers this bug? (looks like something which could be covered with a unittest with a few anchors)
> Does this also affect `overlapper_naive`?

Will do. It does not affect `overlapper_naive` because `overlapper_naive` does not use the order of anchors.

PR #149, opened 2019-09-19
[cudamapper] Added setting of relative strand

vellamike:
* Added support for setting the realtive strand in an Overlap and
printing it to PAF
* Fixed a bug where query_length was incorrectly being replaced with
target length in Overlap.
* Added test for relative strand.
* Deleted naive overlapper source files.


cudamapper/include/cudamapper/types.hpp
```diff
@@ -68,6 +68,8 @@ typedef struct Overlap {
     std::string query_read_name_;
     /// target read name (e.g from FASTA)
     std::string target_read_name_;
+    /// Relative strand: "+" or "-"
+    std::string relative_strand;
```
mimaric: Can this be a `char` or even better an `enum class : unsigned char`?

PR #151, opened 2019-09-19
[cudamapper] Added command line options

vellamike:
* Added option to set kmer and window size via the command line
* Added help option

### DIFF
--- a/cudamapper/src/main.cpp
+++ b/cudamapper/src/main.cpp
@@ -60,6 +60,12 @@ int main(int argc, char *argv[])
         exit(1);
     }
 
+    if (k > claragenomics::Index::maximum_kmer_size()){
+        std::cerr << "kmer of size " << k << " is not allowed, maximum k = " <<
+            claragenomics::Index::maximum_kmer_size() << std::endl;
+        exit(1);
+    }
+
     std::string input_filepath = std::string(argv[optind]);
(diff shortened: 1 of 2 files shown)


vellamike:
rerun tests

cudamapper/src/main.cpp
```diff
@@ -46,3 +82,18 @@ int main(int argc, char *argv[])
     overlapper.print_paf(overlaps);
     return  0;
 }
+
+void help() {
+    std::cout<<
```
mimaric: Have a look at https://www.youtube.com/watch?v=DiZ-az_nJMM, but this is also fine (and probably even easier to understand).

cudamapper/src/main.cpp
```diff
@@ -8,24 +8,60 @@
...
-    // TODO: pass kmer and window size as parameters
-    std::unique_ptr<claragenomics::Index> index = claragenomics::Index::create_index(std::string(argv[1]), 15, 15);
+
+    if (optind >= argc){
+        help();
+        exit(1);
+    }
+
+    std::string input_filepath = std::string(argv[optind]);
+    std::unique_ptr<claragenomics::Index> index = claragenomics::Index::create_index(input_filepath, k, w);
```
mimaric: Make sure `k` is small enough to fit `representation_t`

PR #158, opened 2019-09-26
[cudaaligner] Improvements for Hirschberg+Myers

ahehn-nv:


cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -505,80 +483,159 @@ __device__ void hirschberg_myers_single_char_warp(int8_t*& path, int32_t* path_l
...
+            else if (data_->active_warps_ <= 0)
+            {
+                assert(data_->buffer_begin_ == data_->cur_end_);
+                result = false;
+                break;
+            }
+            __syncwarp();
+            release_mutex();
+#if __CUDA_ARCH__ >= 700
+            asm("nanosleep.u32 100000;");
```
tijyojwad: why do we need a sleep command here?

cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -505,80 +483,159 @@ __device__ void hirschberg_myers_single_char_warp(int8_t*& path, int32_t* path_l
...
+    __device__ inline bool pop(T& element)
     {
-        assert(cur_end_ > buffer_begin_);
-        if (cur_end_ - sizeof(T) >= buffer_begin_)
-            cur_end_ -= sizeof(T);
+        lock_mutex();
+        if (threadIdx.x % warp_size == 0)
+        {
+            assert(data_->active_warps_ > 0);
+            --(data_->active_warps_);
```
tijyojwad: I'm not sure I follow this - could you explain why the active warp count is decremented at the beginning of the pop command, and then incremented later in the condition `if (data_->buffer_begin_ < data_->cur_end_)`?

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: i think i get it. is it to ensure the while loop doesn't just take over an SM while spinning?

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: hmm maybe I understand now. so the `pop` command spins till it gets more data. when it gets data, it knows there will be an active warp to process it. and if new data never comes along, then it ends. did I sort of get it?

cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -505,80 +483,159 @@ __device__ void hirschberg_myers_single_char_warp(int8_t*& path, int32_t* path_l
...
+    __device__ inline void lock_mutex_high_priority() const
     {
-        return buffer_begin_ == cur_end_;
+        if (threadIdx.x % warp_size == 0)
+        {
+            atomicOr(&(data_->mutex_), 0x0000'0002u); // reserve mutex
+            while (2 != atomicCAS(&(data_->mutex_), 2, 1))
+            {
+#if __CUDA_ARCH__ >= 700
+                asm("nanosleep.u32 1000;");
```
tijyojwad: how do you decide the sleep time for these while loop? the ones above have much longer sleeps

cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -505,80 +483,159 @@ __device__ void hirschberg_myers_single_char_warp(int8_t*& path, int32_t* path_l
...
-            cur_end_ += sizeof(T);
-            return true;
+            success = true;
         }
         else
         {
-            if (threadIdx.x == 0)
+            if (threadIdx.x % warp_size == 0)
             {
                 printf("ERROR: stack full!");
```
tijyojwad: This is from a previous PR, but should this be an assert?

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: Okay, I think I see now. Since this is the high priority version, you want the loop to check more often that the normal lock, hence the sleep time is smaller.

cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -505,80 +483,159 @@ __device__ void hirschberg_myers_single_char_warp(int8_t*& path, int32_t* path_l
...
+        __threadfence_block();
     }
 
-    __device__ inline bool empty() const
+    __device__ inline void lock_mutex_high_priority() const
     {
-        return buffer_begin_ == cur_end_;
+        if (threadIdx.x % warp_size == 0)
+        {
+            atomicOr(&(data_->mutex_), 0x0000'0002u); // reserve mutex
```
tijyojwad: this reserving mechanism is really neat btw :)

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: This is from a previous PR, but should this be an assert?

ahehn-nv: No, this should not be checked with an assert. It is something that might actually happen for valid input (if it is too large) and we need to report an error.

ahehn-nv: The error mechanism could need some improvement though...

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: :) glad you like it.

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: Essentially yes, that's what this should do. It can't really take over an SM, but it wastes resources locking and unlocking the mutex all the time. In general, if there was no work just now, the probability that there's new work now is rather low... So, give it a tiny bit of time. :)
Also, while the warp sleeps, it is one warp less to fight for the atomic operations on the mutex, which will make it easier for the `high_priority_lock()` to get the lock, which in return will make it easier for a warp trying to add new work to the queue.
In general the sleep times were just a bit of trial-and-benchmark. I haven't really optimized these values (yet).

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: Yes, I think you got it. Here's my explanation just to make sure:
The `active_warps_` variable is used to determine when all work is done and we can exit. Simply checking for an empty queue would be wrong, because a warp working on a task could fill the queue with new tasks again. Therefore, to check for completion, we need to check that the queue is empty and that there are no warps working on any task anymore. `active_warps_` counts the number of warps that are currently working on a task. If a warp calls `pop()` the warp has obviously no task to work on anymore, so it is not active anymore. When it then gets an task item from the queue (potentially after a bit of spinning in the `while(1)` loop), it will be active again.

This line https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/86a22a94ac5eb46740674a2aab668b4d53e4a25f/cudaaligner/src/hirschberg_myers_gpu.cu#L576 is the check which let's `pop()` return false (meaning "everything done!") and will exit the work loop in the caller.

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: Yes, exactly, I want to be a bit more greedy to get the lock.

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: Let's file an issue to improve the error mechanism later? So the failed alignments can be handled on the CPU potentially

tijyojwad:
rerun tests

Issue #162, opened 2019-10-01
python build fails when Doxygen not present


cjw85:
When setup.py first runs `cmake` it is noted that:

```-- Could NOT find Doxygen (missing: DOXYGEN_EXECUTABLE)
-- Doxygen not found. Doc gen disabled.
```

but the subsequent `cmake --build ... docs install` call fails.

tijyojwad:
Hi @cjw85 - fixed the issue with https://github.com/clara-genomics/ClaraGenomicsAnalysis/pull/163 but merged into our current release development branch (`dev-v0.3.0`). Could you try that branch out and confirm if the changes work for you now?

tijyojwad:
Closing (fix verified here - https://github.com/clara-genomics/ClaraGenomicsAnalysis/pull/163#issuecomment-537049630 )

PR #163, opened 2019-10-01
[pycga] remove docs installation in pycga

tijyojwad:
docs installation is now part of the `all` target. so if Doxygen
is available, docs will automatically installed. And not otherwise.
Explicitly calling docs target causes an error when doxygen is not installed.

Addresses #162


cjw85:
Yes, that works (the change it what I already had done).

PR #165, opened 2019-10-02
[io] add hts lib based fasta parser

tijyojwad:
1. add fasta parser
2. update cudamapper to use fasta faster


vellamike:
rerun tests

tijyojwad:
rerun tests

tijyojwad:
rerun tests

PR #166, opened 2019-10-02
[cudamapper] overlaps to paf file in async thread

tijyojwad:
write overlaps to paf in an async thread so it does't
block the rest of the processing

### DIFF
--- a/cudamapper/src/main.cpp
+++ b/cudamapper/src/main.cpp
@@ -12,10 +12,16 @@
 #include <getopt.h>
 #include <iostream>
 #include <string>
+#include <deque>
+#include <mutex>
+#include <future>
+#include <thread>
+#include <atomic>
 
 #include <claragenomics/logging/logging.hpp>
 
 #include "cudamapper/index.hpp"
+#include "cudamapper/overlapper.hpp"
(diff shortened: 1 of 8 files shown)


cudamapper/src/main.cpp
```diff
@@ -180,6 +232,15 @@ int main(int argc, char *argv[])
         query_start = query_end + 1;
         query_end = query_start + index_size;
     }
+
+    // Insert empty overlap vector to denote end of processing.
```
vellamike: What happens if two indices just have no overlaps? This seems like a bad poison pill.

cudamapper/src/overlapper_triggered.cu
```diff
@@ -166,6 +166,6 @@ namespace claragenomics {
         }
 
         //Return fused overlaps
-        return fuse_overlaps(overlaps);
+        fuse_overlaps(fused_overlaps, overlaps);
```
vellamike: If this function is now `void` then this comment is a tiny bit confusing, it is writing to `&fused_overlaps` really?

tijyojwad:
rerun tests

PR #174, opened 2019-10-04
[CGA] Listed additional packages which should be installed on fresh U…

mimaric:
…buntu installation in order to support htslib

`htslib` installation is really helpful with this, but it's better to list it here explicitly


tijyojwad:
@mimaric this is pretty useful. I'm cha going the target branch to dev 0.3 and we'll merge that branch into ashg-2019 again today

mimaric:
rerun tests

PR #177, opened 2019-10-06
[cudamapper] adding 1 fasta to another fasta mapping

tijyojwad:
Adding fasta to fasta mapping, to check perf of reference guided assembly

### DIFF
--- a/cudamapper/src/main.cpp
+++ b/cudamapper/src/main.cpp
@@ -85,12 +85,19 @@ int main(int argc, char *argv[])
     std::string query_filepath = std::string(argv[optind++]);
     std::string target_filepath = std::string(argv[optind++]);
 
-    bool denovo = false;
+    if (query_filepath == "" || target_filepath == "")
+    {
+        std::cerr << "Invalid inputs. Please refer to the help function." << std::endl;
+        help();
+        exit(1);
+    }
+
+    bool all_to_all = false;
     if (query_filepath == target_filepath)
     {
-        denovo = true;
+        all_to_all = true;
         target_index_size = index_size;
-        std::cerr << "NOTE - Since query and target files are same, activating denovo mode. Target index size used for both files." << std::endl;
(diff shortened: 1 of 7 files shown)


cudamapper/src/main.cpp
```diff
@@ -76,11 +82,25 @@ int main(int argc, char *argv[])
...
-    //Now carry out all the looped polling
-    size_t query_start = 0;
-    size_t query_end = query_start + index_size - 1;
+    std::string query_filepath = std::string(argv[optind++]);
+    std::string target_filepath = std::string(argv[optind++]);
+
+    bool denovo = false;
+    if (query_filepath == target_filepath)
+    {
+        denovo = true;
```
vellamike: Change this to all_vs_all

cudamapper/src/main.cpp
```diff
@@ -76,11 +82,25 @@ int main(int argc, char *argv[])
...
+        target_index_size = index_size;
+        std::cerr << "NOTE - Since query and target files are same, activating denovo mode. Target index size used for both files." << std::endl;
```
vellamike: should change "denovo" to "all vs all" everywhere

cudamapper/src/main.cpp
```diff
@@ -151,63 +176,46 @@ int main(int argc, char *argv[])
...
-        ranges.push_back(query_range);
-
-        std::unique_ptr<claragenomics::Index> index = claragenomics::Index::create_index(input_filepath, k, w, ranges);
+        auto overlapper = claragenomics::OverlapperTriggered();
 
-        CGA_LOG_INFO("Created index");
-        std::cerr << "Index execution time: " << std::chrono::duration_cast<std::chrono::milliseconds>(
-                std::chrono::high_resolution_clock::now() - start_time).count() << "ms" << std::endl;
+        size_t target_start = 0;
+        // If denovo mode, then we can optimzie by starting the target sequences from the same index as
```
vellamike: as above

PR #190, opened 2019-10-12
[cudamapper] Added CIGAR string generation using cudaaligner

ahehn-nv:


ahehn-nv:
will overhaul the whole thing and submit a new PR soon.

PR #192, opened 2019-10-21
Merge ashg hackathon changes into dev branch

tijyojwad:
Only new change is adding cigar string format to PAF output


tijyojwad:
Got it approved earlier by @ahehn-nv . Merging into dev branch now since @vellamike and @mimaric are out.

PR #203, opened 2019-11-04
[cudaaligner] Combine benchmarks into a single binary

### LINKED ISSUE #148: [cudaaligner] combine benchmarks into single application
Use google bench filters instead to choose which set of benchmarks to run

ohadmo:
Fixes #148


cudaaligner/benchmarks/CMakeLists.txt
```diff
@@ -8,9 +8,19 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
-# Add benchmarks
-add_subdirectory(singlealignment)
-add_subdirectory(singlebatchalignment)
+
+project(benchmark_cudaaligner_singlebatch)
```
tijyojwad: Let's call this `benchmark_cudaaligner` like we did for `cudapoa`. READMEs would need to be updated as well.

ahehn-nv:
really glad to see this change! :)

PR #207, opened 2019-11-07
Used pip instead of setuptools directly.

ohadmo:
- Created an installation shell script for pyclaragenomics.
- Separated Clara Genomics SDK installation from setup.py script.
- Updated readme and CI test.


pyclaragenomics/build_clara_genomics_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+
+
+if __name__ == "__main__":
+    args = parse_arguments()
+
+    # Initialize builds.
+    pycga_directory = os.path.dirname(os.path.realpath(__file__))
+    cmake_root_dir = os.path.dirname(pycga_directory)
+    output_folder = \
+        args.build_output_folder if args.build_output_folder else os.path.join(pycga_directory, "cga_build")
```
tijyojwad: we can make the default of `build_output_folder` as `pycga_dir/cga_build` so we won't need this check here

pyclaragenomics/install_pyclaragenomics.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Create build folder if does not exist
+if [ ! -e "$CMAKE_BUILD_FOLDER" ]; then
+  mkdir -p "$CMAKE_BUILD_FOLDER"
+elif [ ! -d "$CMAKE_BUILD_FOLDER" ]; then
+    printf "%s already exists, but is not a directory\n" "$CMAKE_BUILD_FOLDER"
+fi
+
+# Build & install Clara Genomics Analysis SDK
+python build_clara_genomics_analysis.py --build_output_folder "${CMAKE_BUILD_FOLDER}/cga_build"
```
tijyojwad: I'm wondering if we need this python script to wrap cmake anymore. Originally we had the cmake building wrapper in python to keep it colocate with the `setup` call. But now that it's separate, we could just call cmake directly in bash.

We would need to add some special handling for checking if the CUDA_TOOLKIT_DIR env var is set, but other than that it should be a single liner really.

pyclaragenomics/install_pyclaragenomics.sh
```diff
@@ -0,0 +1,99 @@
...
+
+# Prepare arugments for pip setuptools
+PYCG_PATH=$(dirname $(realpath "$0"))
```
tijyojwad: let's call this `PYCGA_DIR`

pyclaragenomics/setup.py
```diff
@@ -10,69 +10,19 @@
...
-pycga_directory = os.path.dirname(os.path.realpath(__file__))
-cmake_root_dir = os.path.dirname(pycga_directory)
-cmake_proj = CMakeWrapper(cmake_root_dir,
-                          cmake_build_path=os.path.join(pycga_directory, "cga_build"),
-                          cmake_extra_args="-Dcga_build_shared=ON")
-cmake_proj.build()
+# Must be set before calling pip
+try:
+    pycga_directory = os.environ['PYCG_PATH']
+    cmake_root_dir = os.environ['CGA_SDK_PATH']
```
tijyojwad: We can remove the `cmake_root_dir` as we can grab the header files from the cmake install location

pyclaragenomics/setup.py
```diff
@@ -83,17 +33,14 @@ def get_installed_path(self, component=""):
             os.path.join(cmake_root_dir, "cudapoa/include"),
```
tijyojwad: these can become `cga_install_dir/include/claragenomics/cudaaligner` and `cga_install_dir/include/claragenomics/cudapoa`

pyclaragenomics/install_pyclaragenomics.sh
```diff
@@ -0,0 +1,99 @@
...
+CGA_SDK_PATH=$(dirname "$PYCG_PATH")
+
+# Call setup script using pip
+PYCG_PATH=${PYCG_PATH} \
+CGA_SDK_PATH=${CGA_SDK_PATH} \
+CMAKE_BUILD_FOLDER="${CMAKE_BUILD_FOLDER}/cga_build" \
```
tijyojwad: let's rename this to `CGA_INSTALL_DIR` and pass it `cga_build/install`

pyclaragenomics/install_pyclaragenomics.sh
tijyojwad: Or we get rid of the shell script and wrap it all in another python script which internally calls `pip install ...`

ohadmo: Got rid of the wrapper shell script.

ci/common/test-pyclaragenomics.sh
```diff
@@ -19,7 +19,7 @@ cd $PYCLARAGENOMICS_DIR
 
 #Install external dependencies.
 python -m pip install -r requirements.txt
-python setup.py install
+./install_pyclaragenomics.sh  -b .
```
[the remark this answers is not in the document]
ohadmo: We decided to go with one python setup file, I renamed build_clara_genomics_analysis.py to be setup_pyclaragenomics.py

pyclaragenomics/setup.py
```diff
@@ -10,96 +10,49 @@
...
+        raise RuntimeError("No valid path for requested component exists")
+    return installed_path
 
-        if not os.path.exists(self.build_path):
-            os.makedirs(self.build_path)
 
-        subprocess.check_call(['cmake', self.cmake_root_dir] + cmake_args, cwd=self.build_path)
+# Must be set before calling pip
+try:
+    pycga_dir = os.environ['PYCG_DIR']
```
tijyojwad: rename `PYCG_DIR` to `PYCGA_DIR`

pyclaragenomics/setup_pyclaragenomics.py
```diff
@@ -0,0 +1,105 @@
...
+        if not os.path.exists(self.build_path):
+            os.makedirs(self.build_path)
+
+        subprocess.check_call(['cmake', self.cmake_root_dir] + cmake_args, cwd=self.build_path)
+
+    def run_build_cmd(self):
+        build_args = ['--', '-j16', 'install']
+        subprocess.check_call(['cmake', '--build', '.'] + build_args, cwd=self.build_path)
+
+    def setup_python_binding(self):
```
tijyojwad: Let's move this into a separate function outside this class since this function doesn't have anything to do with make wrapping. And this function can then take in arguments we get from the CMakeWrapper class

tijyojwad: Let's move this into a separate function outside this class since this function doesn't have anything to do with make wrapping. And this function can then take in arguments we get from the CMakeWrapper class

pyclaragenomics/setup.py
tijyojwad: rename PYCG_DIR to PYCGA_DIR

pyclaragenomics/setup_pyclaragenomics.py
```diff
@@ -0,0 +1,105 @@
...
+        self.run_cmake_cmd()
+        self.run_build_cmd()
+        self.setup_python_binding()
+
+
+if __name__ == "__main__":
+
+    args = parse_arguments()
+    current_dir = os.path.dirname(os.path.realpath(__file__))
+    # Build & install Clara Genomics Analysis SDK
```
tijyojwad: Let's add a log statement here saying "Building in install/develop mode"

PR #208, opened 2019-11-08
[evaluate_paf] Updates to PAF evaluation script

tijyojwad:
1. Moved script to `bin` folder
2. Added ability check against multiple overlaps within same read pair
3. Added option to skip self mapped overlaps
4. Some refactor

### DIFF
--- a/pyclaragenomics/bin/evaluate_paf
+++ b/pyclaragenomics/bin/evaluate_paf
@@ -15,13 +15,30 @@ import argparse
 from claragenomics.io import pafio
 from collections import defaultdict
 
+def match_overlaps(query_0, query_1, target_0, target_1, pos_tolerance):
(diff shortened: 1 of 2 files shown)


pyclaragenomics/bin/evaluate_paf
```diff
@@ -0,0 +1,133 @@
...
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+"""Functions and tools for calculating the accuracy of overlap detection"""
+
+import argparse
+from claragenomics.io import pafio
```
vellamike: internal imports should be after external, and separated by a line.

Issue #157, opened 2019-09-26
Graph representation in python and serialization methods


jonn-smith:
The python bindings are very helpful to perform alignments and consensus calls, but there currently isn't a good way to work with the resulting graph structures in python.  The structures are available in C++, but there are some nuances to them.  It would be nice if there were some examples (with documentation) of working with the resulting graphs in C++ and some bindings (or a new interface) to work with them in python as well.

It would also be helpful if there were a method that can be called after performing an alignment or consensus call that would serialize the graph (in DOT or some similar format) so it can be easily inspected / visualized after creation.

tijyojwad:
Hi @jonn-smith ! Thanks a lot for filing this request, and glad to hear you're finding the APIs useful!

Currently our graph structures are all kept in CUDA memory, but we can certainly expose them in C++ and python data structures for further analysis and visualization. 

Do you have a sense of the priority on this request? That'll help us schedule work for this feature appropriately.

Thanks!

jonn-smith:
@tijyojwad It would be pretty helpful for the work I'm doing, so I would like it to be a high priority (but I also don't want to derail your development cycles).

I was toying with the idea of doing some of it myself (either using the MSA sequences to rebuild the graph without edge weights or exposing the `GraphDetails` with a getter and then updating the python bindings), but I started to get concerned that it might be a rabbit hole for me since I'm not familiar with the details of the code.

tijyojwad:
Sounds good @jonn-smith . We'll prioritize this request, and I will get back to you with an ETA in a day or so

jonn-smith:
Great!  Sounds good!

tijyojwad:
Hi @jonn-smith , apologies for the late response. We're dealing with some urgent items right now for an upcoming release, and it's seems difficult to get this request into that release which is slated for next week. I think however that we can get this feature in sometime around the 4th week of October (maybe soon if things go well next week :) ).

tijyojwad:
Hi @jonn-smith - just wanted to update you that we started working on a graph serialization for cudapoa in https://github.com/clara-genomics/ClaraGenomicsAnalysis/pull/209. The C++ graph should be up in a day or two, the python bindings will follow immediately after. Apologies for getting to this so late!

jonn-smith:
@tijyojwad Sounds good!  No prob!  Glad to hear that it'll be in soon!

tijyojwad:
[...]

PR #209, opened 2019-11-09
[cudapoa] add graph data structure

tijyojwad:
1. Adds graph class to C++ API
2. Add graph serialization to dot format

Addresses #157

### DIFF
--- a/common/utils/include/claragenomics/utils/graph.hpp
+++ b/common/utils/include/claragenomics/utils/graph.hpp
@@ -1,0 +1,116 @@
+/*
+* Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+*
+* NVIDIA CORPORATION and its licensors retain all intellectual property
+* and proprietary rights in and to this software, related documentation
+* and any modifications thereto.  Any use, reproduction, disclosure or
+* distribution of this software and related documentation without an express
+* license agreement from NVIDIA CORPORATION is strictly prohibited.
+*/
+
+#pragma once
+
+#include <vector>
+#include <unordered_map>
+#include <unordered_set>
+#include <stdint.h>
+#include <string>
+#include <iostream>
+#include <algorithm>
+
+namespace claragenomics
+{
+
+using node_id_t = int32_t;
+
+/// \struct pair_hasher
+struct PairHash
+{
+public:
+    template <class T1, class T2>
+    size_t operator()(const std::pair<T1, T2>& pair) const
+    {
+        size_t hash_1 = std::hash<T1>()(pair.first);
+        size_t hash_2 = std::hash<T2>()(pair.second);
+        return hash_1 ^ hash_2;
+    }
+};
+
+/// \class DirectedGraph
+/// Object representing a graph structure
+class DirectedGraph
+{
+public:
+    DirectedGraph() = default;
+
+    ~DirectedGraph() = default;
+
+    virtual void add_edge(node_id_t node_id_from, node_id_t node_id_to)
+    {
+        auto edge      = std::pair<node_id_t, node_id_t>(node_id_from, node_id_to);
+        auto find_edge = edges_.find(edge);
+        if (find_edge == edges_.end())
+        {
+            edges_.insert(edge);
+            auto find_node = adjacent_nodes_.find(node_id_from);
+            if (find_node == adjacent_nodes_.end())
+            {
+                adjacent_nodes_.insert({node_id_from, {node_id_to}});
+            }
+            else
+            {
+                find_node->second.push_back(node_id_to);
+            }
+        }
+    }
+
+    virtual void add_label(node_id_t node, const std::string& label)
+    {
+        node_labels_.insert({node, label});
+    }
+
+    virtual const std::vector<node_id_t>& get_adjacent_nodes(node_id_t node)
+    {
+        auto iter = adjacent_nodes_.find(node);
+        if (iter != adjacent_nodes_.end())
+        {
+            return iter->second;
+        }
+        else
+        {
+            return empty_;
+        }
+    }
+
+    virtual const std::vector<node_id_t> get_node_ids()
+    {
+        std::vector<node_id_t> nodes;
+        for (auto iter : adjacent_nodes_)
+        {
+            nodes.push_back(iter.first);
+        }
+
+        return nodes;
+    }
+
+    virtual std::string get_node_label(node_id_t node)
+    {
+        auto found_node = node_labels_.find(node);
+        if (found_node != node_labels_.end())
+        {
+            return found_node->second;
+        }
+        else
+        {
+            throw std::runtime_error("No node found with given ID");
+        }
+    }
+
+private:
+    std::unordered_map<node_id_t, std::vector<node_id_t>> adjacent_nodes_;
+    std::unordered_set<std::pair<node_id_t, node_id_t>, PairHash> edges_;
+    std::unordered_map<node_id_t, std::string> node_labels_;
+    const std::vector<node_id_t> empty_;
+};
+
+} // namespace claragenomics
--- a/common/utils/include/claragenomics/utils/graph.hpp
+++ b/common/utils/include/claragenomics/utils/graph.hpp
@@ -13,7 +13,7 @@
 #include <vector>
 #include <unordered_map>
 #include <unordered_set>
-#include <stdint.h>
+#include <cstdint>
 #include <string>
 #include <iostream>
 #include <sstream>
@@ -21,10 +21,6 @@
 
 namespace claragenomics
 {
-
-using node_id_t     = int32_t;
-using edge_weight_t = int32_t;
-using edge_t        = std::pair<node_id_t, node_id_t>;
 
 /// \struct PairHash
 /// Hash function for a pair
@@ -45,21 +41,18 @@ public:
 class Graph
 {
 public:
-    /// \brief Default dtor
-    ~Graph() = default;
-
-    /// \brief Add edges to a graph
-    ///
-    /// \param node_id_from Source node ID
-    /// \param node_id_to Sink node ID
-    /// \param weight Edge weight
-    virtual void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight) = 0;
+    /// Typedef for node ID
+    using node_id_t = int32_t;
+    /// Tpyedef for edge weight
+    using edge_weight_t = int32_t;
+    /// Typedef for edge
+    using edge_t = std::pair<node_id_t, node_id_t>;
 
     /// \brief Get a list of adjacent nodes to a given node
     ///
     /// \param node Node for which adjacent nodes are requested
     /// \return Vector of adjacent node IDs
-    virtual const std::vector<node_id_t>& get_adjacent_nodes(node_id_t node) const
+    const std::vector<node_id_t>& get_adjacent_nodes(node_id_t node) const
     {
         auto iter = adjacent_nodes_.find(node);
         if (iter != adjacent_nodes_.end())
@@ -68,14 +61,14 @@ public:
         }
         else
         {
-            return empty_;
+            return Graph::empty_;
         }
     }
 
     /// \brief List all node IDs in the graph
     ///
     /// \return A vector of node IDs
-    virtual const std::vector<node_id_t> get_node_ids() const
+    const std::vector<node_id_t> get_node_ids() const
     {
         std::vector<node_id_t> nodes;
         for (auto iter : adjacent_nodes_)
@@ -89,21 +82,16 @@ public:
     /// \brief Get a list of all edges in the graph
     ///
     /// \return A vector of edges
-    virtual const std::vector<std::pair<edge_t, edge_weight_t>> get_edges() const
+    const std::vector<std::pair<edge_t, edge_weight_t>> get_edges() const
     {
-        std::vector<std::pair<edge_t, edge_weight_t>> edges;
-        for (auto iter : edges_)
-        {
-            edges.push_back({iter.first, iter.second});
-        }
-        return edges;
+        return {begin(edges_), end(edges_)};
     }
 
     /// \brief Add string labels to a node ID
     ///
     /// \param node ID of node
     /// \param label Label to attach to that node ID
-    virtual void set_node_label(node_id_t node, const std::string& label)
+    void set_node_label(node_id_t node, const std::string& label)
     {
         node_labels_.insert({node, label});
     }
@@ -113,7 +101,7 @@ public:
     /// \param node node ID for label query
     /// \return String label for associated node. Returns empty string if
     //          no label is associated or node ID doesn't exist.
-    virtual std::string get_node_label(node_id_t node) const
+    std::string get_node_label(node_id_t node) const
     {
         auto found_node = node_labels_.find(node);
         if (found_node != node_labels_.end())
@@ -126,29 +114,14 @@ public:
         }
     }
 
-    /// \brief Serialize graph structure to dot format
-    ///
-    /// \return A string encoding the graph in dot format
-    virtual std::string serialize_to_dot() const = 0;
-
 protected:
-    Graph() = default;
-
     /// \brief Check if a directed edge exists in the grph
     ///
     /// \param edge A directed edge
     /// \return Boolean result of check
     bool directed_edge_exists(edge_t edge)
     {
-        auto find_edge = edges_.find(edge);
-        if (find_edge == edges_.end())
-        {
-            return false;
-        }
-        else
-        {
-            return true;
-        }
+        return edges_.find(edge) != edges_.end();
     }
 
     /// \brief Update the adjacent nodes based on edge information
@@ -173,12 +146,13 @@ protected:
     void node_labels_to_dot(std::ostringstream& dot_str) const
     {
         const std::vector<node_id_t> nodes = get_node_ids();
-        for (auto node : nodes)
+        for (auto iter : adjacent_nodes_)
         {
-            auto label_found = node_labels_.find(node);
+            auto& node_id    = iter.first;
+            auto label_found = node_labels_.find(node_id);
             if (label_found != node_labels_.end())
             {
-                dot_str << node << " [label=\"" << label_found->second << "\"];\n";
+                dot_str << node_id << " [label=\"" << label_found->second << "\"];\n";
             }
         }
     }
@@ -215,16 +189,12 @@ protected:
 class DirectedGraph : public Graph
 {
 public:
-    DirectedGraph() = default;
-
-    ~DirectedGraph() = default;
-
     /// \brief Add directed edges to graph.
     ///
     /// \param node_id_from Source node ID
     /// \param node_id_to Sink node ID
     /// \param weight Edge weight
-    virtual void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight = 0) override
+    void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight = 0)
     {
         auto edge = edge_t(node_id_from, node_id_to);
         if (!directed_edge_exists(edge))
@@ -237,7 +207,7 @@ public:
     /// \brief Serialize graph structure to dot format
     ///
     /// \return A string encoding the graph in dot format
-    virtual std::string serialize_to_dot() const override
+    std::string serialize_to_dot() const
     {
         std::ostringstream dot_str;
         dot_str << "digraph g {\n";
@@ -257,16 +227,12 @@ public:
 class UndirectedGraph : public Graph
 {
 public:
-    UndirectedGraph() = default;
-
-    ~UndirectedGraph() = default;
-
     /// \brief Add undirected edges to graph.
     ///
     /// \param node_id_from Source node ID
     /// \param node_id_to Sink node ID
     /// \param weight Edge weight
-    virtual void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight = 0) override
+    void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight = 0)
     {
         auto edge          = edge_t(node_id_from, node_id_to);
         auto edge_reversed = edge_t(node_id_to, node_id_from);
@@ -281,7 +247,7 @@ public:
     /// \brief Serialize graph structure to dot format
     ///
     /// \return A string encoding the graph in dot format
-    virtual std::string serialize_to_dot() const override
+    std::string serialize_to_dot() const
     {
         std::ostringstream dot_str;
         dot_str << "graph g {\n";
--- a/common/utils/include/claragenomics/utils/graph.hpp
+++ b/common/utils/include/claragenomics/utils/graph.hpp
@@ -41,8 +41,7 @@ public:
     }
 };
 
-/// \class Graph
-/// Object representing a generic graph structure
+/// \brief Object representing a generic graph structure
 class Graph
 {
 public:
@@ -53,6 +52,7 @@ public:
     ///
     /// \param node_id_from Source node ID
     /// \param node_id_to Sink node ID
+    /// \param weight Edge weight
     virtual void add_edge(node_id_t node_id_from, node_id_t node_id_to, edge_weight_t weight) = 0;
 
     /// \brief Get a list of adjacent nodes to a given node
@@ -167,6 +167,9 @@ protected:
         }
     }
 
+    /// \brief Serialize node labels to dot format
+    ///
+    /// \param dot_str Output string stream to serialize labels to
     void node_labels_to_dot(std::ostringstream& dot_str) const
     {
         const std::vector<node_id_t> nodes = get_node_ids();
@@ -180,6 +183,10 @@ protected:
         }
     }
 
+    /// \brief Serialize edges to dot format
+    ///
+    /// \param dot_str Output string stream to serialize labels to
+    /// \param node_separator DOT delimiter for edge description
     void edges_to_dot(std::ostringstream& dot_str, const std::string& node_separator) const
     {
         for (auto iter : edges_)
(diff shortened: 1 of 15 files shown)


tijyojwad:
A sample of a POA graph - 

image

cudapoa/src/cudapoa_batch.hpp
```diff
@@ -66,6 +66,9 @@ class CudapoaBatch : public Batch
     StatusType get_msa(std::vector<std::vector<std::string>>& msa,
                        std::vector<StatusType>& output_status);
 
+    void get_graphs(std::vector<DirectedGraph>& graphs,
```
ahehn-nv: Is there really added value in getting the status for each individual graph retrieval instead of just returning one status?

cudapoa/src/allocate_block.cpp
```diff
@@ -120,7 +120,13 @@ std::tuple<int64_t, int64_t, int64_t, int64_t> BatchBlock::calculate_space_per_p
     device_size_per_poa += (output_mask_ & OutputType::msa) ? poa_count * max_sequences_per_poa_ * sizeof(uint16_t) : 0; // input_details_d_->sequence_begin_nodes_ids
 
     // for graph - host
-    host_size_fixed += sizeof(GraphDetails); // graph_details_d_
+    host_size_fixed += sizeof(GraphDetails);                                                            // graph_details_h_
+    host_size_fixed += sizeof(GraphDetails);                                                            // graph_details_d_
```
ahehn-nv: shouldn't this line move further down to the `// for graph - device` section?

cudapoa/src/allocate_block.cpp
```diff
@@ -259,13 +265,25 @@ void BatchBlock::get_alignment_details(AlignmentDetails** alignment_details_d_p)
...
 {
     GraphDetails* graph_details_d{};
+    GraphDetails* graph_details_h{};
 
     // on host
+    graph_details_h = reinterpret_cast<GraphDetails*>(&block_data_h_[offset_h_]);
+    offset_h_ += sizeof(GraphDetails);
+    graph_details_h->nodes = &block_data_h_[offset_h_];
+    offset_h_ += sizeof(uint8_t) * max_nodes_per_window_ * max_poas_;
+    graph_details_h->incoming_edges = reinterpret_cast<uint16_t*>(&block_data_h_[offset_h_]);
```
ahehn-nv: This may lead to a problem if `offset_h_` happens to be odd here, because `uint16_t` has stricter alignment requirements than `uint8_t` on most platforms.

cudapoa/src/cudapoa_batch.cpp
```diff
@@ -312,6 +312,83 @@ StatusType CudapoaBatch::get_msa(std::vector<std::vector<std::string>>& msa, std
...
+                                     cudaMemcpyDeviceToHost,
+                                     stream_));
+
+    CGA_CU_CHECK_ERR(cudaMemcpyAsync(graph_details_h_->incoming_edge_count,
+                                     graph_details_d_->incoming_edge_count,
+                                     sizeof(uint16_t) * max_nodes_per_window_ * max_poas_,
+                                     cudaMemcpyDeviceToHost,
+                                     stream_));
+
+    CGA_CU_CHECK_ERR(cudaMemcpyAsync(input_details_h_->sequence_lengths,
```
ahehn-nv: not 100% sure, but this might be a bug: do you intend to copy `global_sequence_idx_` many sequence lengths here or just one at `input_details_d_->sequence_lengths+global_sequence_idx_`?

ahehn-nv: or should that be `poa_count_`?

cudapoa/src/cudapoa_batch.cpp
```diff
@@ -312,6 +312,83 @@ StatusType CudapoaBatch::get_msa(std::vector<std::vector<std::string>>& msa, std
...
+                                     stream_));
+
+    CGA_CU_CHECK_ERR(cudaMemcpyAsync(output_details_h_->consensus,
+                                     output_details_d_->consensus,
+                                     CUDAPOA_MAX_CONSENSUS_SIZE * max_poas_ * sizeof(uint8_t),
+                                     cudaMemcpyDeviceToHost,
+                                     stream_));
+
+    CGA_CU_CHECK_ERR(cudaStreamSynchronize(stream_));
+
```
ahehn-nv: Since you know the final number of graphs, and copying the graphs will be rather expensive, you could do a `graphs.reserve(poa_count_);` here.

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: ok, I think I get it now, but please double check.

cudapoa/src/cudapoa_batch.cpp
```diff
@@ -312,6 +312,83 @@ StatusType CudapoaBatch::get_msa(std::vector<std::vector<std::string>>& msa, std
...
+            DirectedGraph& graph = graphs.back();
+            int32_t seq_0_offset = input_details_h_->window_details[poa].seq_len_buffer_offset;
+            int32_t num_nodes    = input_details_h_->sequence_lengths[seq_0_offset];
+            uint8_t* nodes       = &graph_details_h_->nodes[max_nodes_per_window_ * poa];
+            for (int32_t n = 0; n < num_nodes; n++)
+            {
+                // For each node, find it's incoming edges and add the edge to the graph,
+                // along with its label.
+                node_id_t sink = n;
+                graph.set_node_label(sink, std::string(1, static_cast<char>(nodes[n])));
```
ahehn-nv: How about `std::to_string(nodes[n])` here?

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+    size_t operator()(const std::pair<T1, T2>& pair) const
+    {
+        size_t hash_1 = std::hash<T1>()(pair.first);
+        size_t hash_2 = std::hash<T2>()(pair.second);
+        return hash_1 ^ hash_2;
+    }
+};
+
+/// \brief Object representing a generic graph structure
+class Graph
```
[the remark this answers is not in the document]
tijyojwad: Only use the Directed and Undirected classes directly. Sure, will get rid of the ctor and dtor for Graph. But since the the derived classes need the serialize and add_edge functions to be custom, wouldn't they need `virtual`? Or did you mean all except those?

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+        }
+    }
+
+    /// \brief Serialize graph structure to dot format
+    ///
+    /// \return A string encoding the graph in dot format
+    virtual std::string serialize_to_dot() const = 0;
+
+protected:
+    Graph() = default;
```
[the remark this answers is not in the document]
tijyojwad: Oh really? I assumed since it's default it wouldn't have an impact on the auto generation of move constructors. But will remove to be safe.

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+
+    /// \brief Check if a directed edge exists in the grph
+    ///
+    /// \param edge A directed edge
+    /// \return Boolean result of check
+    bool directed_edge_exists(edge_t edge)
+    {
+        auto find_edge = edges_.find(edge);
```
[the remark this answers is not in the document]
tijyojwad: XD good point!

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+
+        return nodes;
+    }
+
+    /// \brief Get a list of all edges in the graph
+    ///
+    /// \return A vector of edges
+    virtual const std::vector<std::pair<edge_t, edge_weight_t>> get_edges() const
+    {
+        std::vector<std::pair<edge_t, edge_weight_t>> edges;
```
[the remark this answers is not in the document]
tijyojwad: Neat! Semantically better to have `begin(edges_)` or `edges_.begin()`?

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+        for (auto iter : edges_)
+        {
+            const edge_t& edge          = iter.first;
+            const edge_weight_t& weight = iter.second;
+            dot_str << edge.first << " " << node_separator << " " << edge.second;
+            dot_str << " [label=\"" << weight << "\"];\n";
+        }
+    }
+
+    /// List of adjacent nodes per node ID
```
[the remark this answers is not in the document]
tijyojwad: I think they're `protected`? doxygen was complaining, so I added the comments (which are a bit unnecessary but helpful still I guess)

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+        {
+            edges_.insert({edge, weight});
+            update_adject_nodes(edge);
+        }
+    }
+
+    /// \brief Serialize graph structure to dot format
+    ///
+    /// \return A string encoding the graph in dot format
+    virtual std::string serialize_to_dot() const override
```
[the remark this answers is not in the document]
tijyojwad: The reason for this was that having a `string` as an output makes the corresponding python bindings easier for this function. I don't even know if cython supports dealing with string streams directly

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+            find_node->second.push_back(edge.second);
+        }
+    }
+
+    /// \brief Serialize node labels to dot format
+    ///
+    /// \param dot_str Output string stream to serialize labels to
+    void node_labels_to_dot(std::ostringstream& dot_str) const
+    {
+        const std::vector<node_id_t> nodes = get_node_ids();
```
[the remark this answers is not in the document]
tijyojwad: Good catch, I intended to get back and clean it up but slipped my mind later

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+        else
+        {
+            find_node->second.push_back(edge.second);
+        }
+    }
+
+    /// \brief Serialize node labels to dot format
+    ///
+    /// \param dot_str Output string stream to serialize labels to
+    void node_labels_to_dot(std::ostringstream& dot_str) const
```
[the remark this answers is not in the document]
tijyojwad: Unlabeled nodes would then get the node ID as a number by default. If we assume unlabaled nodes have an empty string label, then a bunch of nodes would just show up as blank

common/utils/tests/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
...
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
+#
+# NVIDIA CORPORATION and its licensors retain all intellectual property
+# and proprietary rights in and to this software, related documentation
+# and any modifications thereto.  Any use, reproduction, disclosure or
+# distribution of this software and related documentation without an express
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+project(cgautilstests)
```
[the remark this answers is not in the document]
tijyojwad: I believe that's how CMake binaries need to be setup?
e.g.
https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.4.0/cudaaligner/tests/CMakeLists.txt

We also don't have any tests for our other utils. So I intend to grow this for tests around utility functions.

cudapoa/src/cudapoa_batch.hpp
tijyojwad: Some of the POAs don't get generated, and sometimes useful to know what the error is.

cudapoa/src/allocate_block.cpp
tijyojwad: The device allocation for graph is already done. This is the host memory allocation for when parts of the graph are copied back

common/utils/include/claragenomics/utils/graph.hpp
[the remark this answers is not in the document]
ahehn-nv: They are equivalent. `begin(x)` was introduced in C++11 and will call x.begin() in almost all cases.
I made it a habit to use it because it allows you to extend classes which don't have `.begin()` member functions without touching the class itself. You just need to add a `begin(const someclass& x)` function.

cudapoa/src/allocate_block.cpp
tijyojwad: Fair point. Not likely to happen because our max nodes and all are defined to be even numbers, but I'll try to address this as well

common/utils/include/claragenomics/utils/graph.hpp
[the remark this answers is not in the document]
ahehn-nv: Yes, see e.g. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-five. Oh they also have a rule about not declaring the constructor if it isn't needed: http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-zero

cudapoa/src/cudapoa_batch.cpp
tijyojwad: The lengths of the updated graphs are stored in the 0th position of the sequence lengths. When optimizing it seemed the best way to re-use memory without having to introduce yet another buffer. And the sequence lengths are a contiguous array with lengths. So we need the 0th index for each POA within this contiguous buffer (that logic is within the loop).

cudapoa/src/cudapoa_batch.cpp
tijyojwad: Right now it's an `emplace_back`, but I can overlap the reserve call with the cuda memcpies to use that latency

cudapoa/src/cudapoa_batch.cpp
tijyojwad: `nodes[n]` is stored as an `int8_t`. wouldn't `to_string` just stringify the number?

common/utils/tests/TestGraph.cpp
```diff
@@ -0,0 +1,77 @@
...
+*/
+
+#include <claragenomics/utils/graph.hpp>
+
+#include "gtest/gtest.h"
+
+namespace claragenomics
+{
+
+TEST(GraphTest, DirectediGraph)
```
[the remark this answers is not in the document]
tijyojwad: Typo, will fix

common/utils/tests/main.cpp
```diff
@@ -0,0 +1,19 @@
+/*
```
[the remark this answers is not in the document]
tijyojwad: Because I expect more utilities tests to be added to the `cgautilstests` project

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: `emplace_back()` will only place the newly created object directly in the vector (which I think it doesn't even do if you create the object explicitly. I think you should do `v.emplace_back({});`, such that the constructor call gets forwarded).
The problem I see is that the vector will grow in each iteration, which implies the vector will reallocate its memory every now and then (when `v.size()` reached `v.capacity()`). This implies it has to move (or copy if it can't move) all its current elements to the newly allocated memory. So I would like to set the vector's capacity to the expected size right from the start by calling `v.reserve(expected_size)`. The `emplace_back()` and everything else will remain the same.

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: oh sorry, my bad. I thought that's what you were doing here.

common/utils/include/claragenomics/utils/graph.hpp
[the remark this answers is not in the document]
ahehn-nv: fair enough. You could also add a `to_string(DirectedGraph const& g)` free function based on this operator. :P

common/utils/include/claragenomics/utils/graph.hpp
```diff
@@ -0,0 +1,300 @@
...
+    std::unordered_map<node_id_t, std::vector<node_id_t>> adjacent_nodes_;
+
+    /// All edges in the graph
+    std::unordered_map<edge_t, edge_weight_t, PairHash> edges_;
+
+    /// Label per node
+    std::unordered_map<node_id_t, std::string> node_labels_;
+
+    /// An empty list representing no connectivity
+    const std::vector<node_id_t> empty_;
```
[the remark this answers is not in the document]
tijyojwad: After I started making this change realized that to initialize the static member I would need to put it outside the header file, otherwise will get multiple definitions errors... Was hoping to keep this header only

common/utils/tests/CMakeLists.txt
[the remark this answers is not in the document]
ahehn-nv: You can definitely have multiple binaries in a project in CMake. But maybe it is due to our cmake structure... We should look into this at some point.

common/utils/include/claragenomics/utils/graph.hpp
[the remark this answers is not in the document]
ahehn-nv: Since they don't provide any functionality in the base class and you don't intend a runtime interface just remove them and only define them in the derived classes.
Even if you had implemented them, there is no technical reason to make them virtual as long as you don't want to refer to the derived classes by a common `Graph* g = new DirectedGraph();` (or similar).
The only "problem" if you have the same non-virtual function in a base and derived class is function/name hiding, i.e. the DirectedGraph function will hide ALL the base class' function with the same name (unless you do `using Base::function;` explicitly) - which can lead to nasty bugs if you are not careful.

cudapoa/src/allocate_block.cpp
ahehn-nv: ah sorry, yes, you are right. I misread this line. All good.

cudapoa/src/cudapoa_batch.hpp
ahehn-nv: ok, makes sense.

common/utils/tests/main.cpp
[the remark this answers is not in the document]
ahehn-nv: ok

common/utils/tests/TestGraph.cpp
```diff
@@ -0,0 +1,77 @@
...
+    EXPECT_NE(std::find(adjacent_nodes_to_2.begin(), adjacent_nodes_to_2.end(), 3), adjacent_nodes_to_2.end());
+    EXPECT_NE(std::find(adjacent_nodes_to_2.begin(), adjacent_nodes_to_2.end(), 4), adjacent_nodes_to_2.end());
+    EXPECT_NE(std::find(adjacent_nodes_to_2.begin(), adjacent_nodes_to_2.end(), 5), adjacent_nodes_to_2.end());
+    EXPECT_EQ(std::find(adjacent_nodes_to_2.begin(), adjacent_nodes_to_2.end(), 1), adjacent_nodes_to_2.end());
+
+    const auto& adjacent_nodes_to_3 = graph.get_adjacent_nodes(3);
+    EXPECT_EQ(std::find(adjacent_nodes_to_3.begin(), adjacent_nodes_to_3.end(), 2), adjacent_nodes_to_3.end());
+}
+
+TEST(GraphTest, UndirectediGraph)
```
ahehn-nv: same typo as above

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: so you need the 0th and the `global_sequence_idx_-1`th entry?

cudapoa/src/cudapoa_batch.cpp
tijyojwad: Replaced with `resize` which will allocate and initialize together

cudapoa/src/allocate_block.cpp
tijyojwad: Will address in a separate PR - https://github.com/clara-genomics/ClaraGenomicsAnalysis/issues/215

cudapoa/src/cudapoa_batch.cpp
tijyojwad: The global sequence index spans _all_ the sequences passed to device, across all POAs. So for example, if POA 1 had 10 sequences, POA 2 had 15 sequences, POA 3 had 5 sequences, `global_sequence_idx` would be 30. And the indices we would need are 0 (first length for POA 1), 15 (first length for POA 2) and 25 (first length for POA 3)

common/utils/tests/CMakeLists.txt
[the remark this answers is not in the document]
tijyojwad: Yep, can certainly have multiple binaries within a single project. In reality the tests folder doesn't need to be independent project since we set the `project` at the module level once. But that's somehow become the convention within our repo (for single binaries, project and target have become the same)

cudapoa/src/cudapoa_batch.cpp
```diff
@@ -312,6 +312,85 @@ StatusType CudapoaBatch::get_msa(std::vector<std::vector<std::string>>& msa, std
...
+            output_status.emplace_back(claragenomics::cudapoa::StatusType::success);
+            DirectedGraph& graph = graphs[poa];
+            int32_t seq_0_offset = input_details_h_->window_details[poa].seq_len_buffer_offset;
+            int32_t num_nodes    = input_details_h_->sequence_lengths[seq_0_offset];
+            uint8_t* nodes       = &graph_details_h_->nodes[max_nodes_per_window_ * poa];
+            for (int32_t n = 0; n < num_nodes; n++)
+            {
+                // For each node, find it's incoming edges and add the edge to the graph,
+                // along with its label.
+                Graph::node_id_t sink = n;
```
ahehn-nv: Remove the `Graph` class here, and change the types to `DirectedGraph::node_id_t`, ...

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: That's probably even better.

cudapoa/src/cudapoa_batch.cpp
ahehn-nv: ok, got it. Thanks for the explanation.

common/utils/tests/CMakeLists.txt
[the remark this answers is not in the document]
ahehn-nv: Is there a reason why we do that? Otherwise I'd remove them everywhere - these are really not separate projects, but tests for a project.

common/utils/include/claragenomics/utils/graph.hpp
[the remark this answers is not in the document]
tijyojwad: Duh, yes...

common/utils/tests/CMakeLists.txt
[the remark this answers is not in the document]
tijyojwad: No good reason, was going to address in a separate PR. but small enough so I changed it in this one itself.

common/utils/tests/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
-project(cgautilstests)
+set(TARGET_NAME cgautilstests)
```
ahehn-nv: why do you define variable for this?

cudaaligner/tests/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
-project(cudaalignertests)
+set(TARGET_NAME cudaalignertests)
```
ahehn-nv: see above.

cudamapper/tests/CMakeLists.txt
```diff
@@ -8,8 +8,7 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
-cmake_minimum_required(VERSION 3.10.2)
-project(cudamappertests)
+set(TARGET_NAME cudamappertests)
```
ahehn-nv: see above.

cudapoa/tests/CMakeLists.txt
```diff
@@ -8,7 +8,7 @@
 # license agreement from NVIDIA CORPORATION is strictly prohibited.
 #
 
-project(cudapoatests)
+set(TARGET_NAME cudapoatests)
```
ahehn-nv: see above

common/utils/tests/CMakeLists.txt
tijyojwad: It's a future proofing habit I have for cmake. These cmakelist files tend to grow, and it's sets the stage for how to address the target

ahehn-nv: I fail to see how a TARGET_NAME variable helps.
Instead of
```
add_executable(cgautilstest)
target_link_libraries(cgautilstest ...)
somemorestuff(cgautilstest ...)
```
you write
```
add_executable(${TARGET_NAME})
target_link_libraries(${TARGET_NAME} ...)
somemorestuff(${TARGET_NAME} ...)
```
This can never be a list, so why is this more future proof?

ahehn-nv: If you add more targets, you will have to introduce a second TARGET_NAME variable or overwrite the first one, which will only make it more obscure.

The only benefit is that you can do c&p from CMakeLists.txt from other similar test directories.

tijyojwad: 1. It puts the main target of the CMake file up top, making it more readable.
2. Creates a template for future files of the same type. New test CMake files can use this as a template and only replace target_name and sources.
3. Low likelihood (but has happened before in drive works e.g.) when copying string names multiple times can easily make mistake with similarly named targets.

We can dig into each of these if you'd like, but I see diminishing returns...

ahehn-nv: We could. :) But I guess you are right it is beside the point. It is a detail after all.

Issue #206, opened 2019-11-05
[CI] Add support for executing gpuCI tests locally through nvidia-docker


ohadmo:
Follow https://github.com/rapidsai/cudf/tree/branch-0.11/ci/local

PR #210, opened 2019-11-11
[CI] Run CI locally

ohadmo:
Fixes issue #206

### DIFF
--- a/README.md
+++ b/README.md
@@ -167,3 +167,24 @@ To run style check manually, simply run the following from the top level folder.
 ```
 flake8 pyclaragenomics/
 ```
+
+## Running CI Tests Locally
+Please note, your git repository will be mounted to the container, any untracked files will be removed from it.
+Before executing the CI locally, stash or add them to the index.
+
+Requirements:
+1. docker (https://docs.docker.com/install/linux/docker-ce/ubuntu/)
+2. nvidia-docker (https://github.com/NVIDIA/nvidia-docker)
+3. nvidia-container-runtime (https://github.com/NVIDIA/nvidia-container-runtime)
+
+Run the following command to execute the CI build steps inside a container locally:
+```bash
+bash ci/local/build.sh -r <ClaraGenomicsAnalysis repo path>
+```
+ci/local/build.sh script was adapted from [rapidsai/cudf](https://github.com/rapidsai/cudf/tree/branch-0.11/ci/local)
+
+The default docker image is **clara-genomics-base:cuda10.0-ubuntu16.04-gcc5-py3.7**.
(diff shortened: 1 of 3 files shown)


ci/common/prep-init-env.sh
```diff
@@ -57,5 +57,6 @@ fi
 
 # Cleanup local git
 cd $1
-git clean -xdf
+# Exclude folder for running local CI tests
+git clean -xdf -e ci_binded/
```
tijyojwad: It looks like this folder is generated by the local CI run within the repo, but it's empty (ls ci_binded/build shows nothing). Does it get cleaned after the test? Can it be removed?

README.md
```diff
@@ -165,3 +165,29 @@ To run style check manually, simply run the following from the top level folder.
...
+```
+ci/local/build.sh script was adapted from [rapidsai/cudf](https://github.com/rapidsai/cudf/tree/branch-0.11/ci/local)
+
+The default docker image is **clara-genomics-base:cuda10.0-ubuntu16.04-gcc5-py3.7**.
+Other images from [gpuci/clara-genomics-base](https://hub.docker.com/r/gpuci/clara-genomics-base/tags) repository can be used instead, by using -i argument
+```bash
+bash ci/local/build.sh -r <ClaraGenomicsAnalysis repo path> -i clara-genomics-base:cuda10.0-ubuntu18.04-gcc7-py3.6
+```  
+
+Please note, your git repository will be mounted to the container, any untracked files will be removed from it.
```
tijyojwad: Let's move this note to the beginning of this section :)

tijyojwad:
btw @ohadmo This worked beautifully for me

ci/common/prep-init-env.sh
ohadmo: @tijyojwad, <repp>/ci_binded is mounted to <repo>/<docker_image_tag_name>/cpp. Once  <repo>/<docker_image>/cpp content is deleted by the git clean command, the folder remains empty. This mounted directory allows using the same BUILD_SCRIPT content for different images. We do not really need this, because we anyway clean the environment before, Or with could use a symbolic link.

README.md
ohadmo: You are right!

ci/common/prep-init-env.sh
tijyojwad: @ohadmo I see, okay not a big deal if it stays around since it doesn't show up on `git status` and so won't get checked in by mistake.

ohadmo: @tijyojwad I removed the mounted directory creation, we do not need it.

tijyojwad: Sweet!

PR #211, opened 2019-11-12
[cmake] Add an option to disable doxygen documentation generation

### LINKED ISSUE #205: [common] Option to disable doc build
Please add an option `cga_build_documentation` (or similar) to the cmake options, such that you can opt-out of building the documentation.

Right now the documentation is built if cmake finds doxygen and I'm not aware of any way to turn it off.

ohadmo:
Fixes #205

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -28,17 +28,24 @@ option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHE
 option(spoa_accurate "Run cudapoa code in mode that matches spoa" OFF)
 option(cga_enable_cudapoa_nw_print "Enable verbose prints within cudapoa NW kernel" OFF)
 option(cga_profiling "Compile a binary for profiling with NVTX markers." OFF)
+option(cga_generate_docs "Generate Doxygen documentation" ON)
 
+# Must be included before others for options value validation
+include(cmake/Utils.cmake)
+
+validate_boolean(cga_enable_tests)
 if (cga_enable_tests)
     message(STATUS "Enabling ClaraGenomicsAnalysis unit tests")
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -65,8 +66,15 @@ add_subdirectory(cudamapper)
 add_subdirectory(cudaaligner)
 
 # Add documentation generation.
-set_doxygen_mainpage(${CMAKE_CURRENT_SOURCE_DIR}/README.md)
-add_docs_target("ClaraGenomicsAnalysis" "${CGA_VERSION}")
+if (cga_generate_docs STREQUAL "ON")
```
tijyojwad: Since ON/OFF doubles up as a Boolean in cmake, we can directly test for the `if(cga_generate_docs)` and in the else case print the message

ohadmo: As we said, I created an argument validation method under cmake/Utils.cmake

CMakeLists.txt
```diff
@@ -28,17 +28,23 @@ option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHE
 option(spoa_accurate "Run cudapoa code in mode that matches spoa" OFF)
 option(cga_enable_cudapoa_nw_print "Enable verbose prints within cudapoa NW kernel" OFF)
 option(cga_profiling "Compile a binary for profiling with NVTX markers." OFF)
+option(cga_generate_docs "Generate Doxygen documentation" ON)
 
+include(cmake/Utils.cmake)
```
tijyojwad: @ohadmo can you add a comment above this include saying that we have to include it before any of the other includes since it had utilities?

ahehn-nv:
Why do we validate if the `option()` variables are only "OFF" or "ON"? How can they be anything else? If you specify with `-D...`?

PR #217, opened 2019-11-15
[cudamapper] Minimizer hashing

vellamike:
Hashing is applied to minimizer representations to give a more uniform distribution of minimizers across the representation space.


vellamike:
> Actually, wait... We should adapt the unittests before merging it in... I know it's tedious, but it makes catching errors in the future much easier....

What did you have in mind? I was hesitant to just change the existing unit tests to use the new hash values because the existing ones have a very clear interpretation (you can easily see just by looking at them how the minimizer and corresponding value was selected).

Perhaps I can keep the existing unit tests and add some new ones which utilise the hash function during index generation and just check for specific values of minimizers?

mimaric:
> What did you have in mind? I was hesitant to just change the existing unit tests to use the new hash values because the existing ones have a very clear interpretation (you can easily see just by looking at them how the minimizer and corresponding value was selected).
> 
> Perhaps I can keep the existing unit tests and add some new ones which utilise the hash function during index generation and just check for specific values of minimizers?

Yes, that's what comes to my mind, having test for `hash_representations == true` and `hash_representations`= false`. I think it's important to cover this case as well.

vellamike:
> that's what comes to my mind, having test for `hash_representations == true` and `hash_representations`= false`. I think it's important to cover this case as well.

@mimaric  done - please re-review

PR #219, opened 2019-11-19
[cudaaligner] Added pairing string to FormattedAlignment

### LINKED ISSUE #141: [cudaaligner] FormattedAlignment should include `|` and `x` symbols
`FormattedAlignment` in cudaaligner returns aligned strings in the following format:

```
ACCGTCA
ACGC--A
```

This would be preferable:

```
ACCGTCA
||xx  |
ACGC--A
```

ohadmo:
Fixes #141

### DIFF
--- a/cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
+++ b/cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
@@ -24,11 +24,16 @@ namespace cudaaligner
 /// \addtogroup cudaaligner
 /// \{
 
-/// \typedef FormattedAlignment
-/// Holds formatted strings representing an alignment.
-/// FormattedAlignment.first = formatted string for query
-/// FormattedAlignment.second = formatted string for target
-typedef std::pair<std::string, std::string> FormattedAlignment;
+/// FormattedAlignment -Holds formatted strings representing an alignment.
+typedef struct FormattedAlignment
+{
+    /// \brief FormattedAlignment.first = formatted string for query
+    std::string first;
+    /// \brief FormattedAlignment.pairing = formatted pairing string
+    std::string pairing;
+    /// \brief FormattedAlignment.first = formatted string for target
+    std::string second;
+} FormattedAlignment;
 
 /// Alignment - Object encapsulating an alignment between 2 string.
 class Alignment
--- a/cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
+++ b/cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
@@ -27,13 +27,20 @@ namespace cudaaligner
 /// FormattedAlignment -Holds formatted strings representing an alignment.
 typedef struct FormattedAlignment
 {
-    /// \brief FormattedAlignment.first = formatted string for query
-    std::string first;
+    /// \brief FormattedAlignment.query = formatted string for query
+    std::string query;
     /// \brief FormattedAlignment.pairing = formatted pairing string
     std::string pairing;
-    /// \brief FormattedAlignment.first = formatted string for target
-    std::string second;
+    /// \brief FormattedAlignment.target = formatted string for target
+    std::string target;
+    /// \brief Maximal line length when outputting to ostream, default = 80 (default terminal size)
+    uint32_t linebreak_after = 80;
 } FormattedAlignment;
+
+/// \brief Write FormattedAlignment struct content to output stream
+///
+/// \return ostream
+std::ostream& operator<<(std::ostream& os, const FormattedAlignment& formatted_alignment);
 
 /// Alignment - Object encapsulating an alignment between 2 string.
 class Alignment
@@ -72,7 +79,7 @@ public:
     virtual const std::vector<AlignmentState>& get_alignment() const = 0;
 
     /// \brief Print formatted alignment to stderr.
-    virtual FormattedAlignment format_alignment() const = 0;
+    virtual FormattedAlignment format_alignment(int32_t maximal_line_length = 80) const = 0;
 };
 
 /// \}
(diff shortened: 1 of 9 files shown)


cudaaligner/src/alignment_impl.cpp
```diff
@@ -78,31 +78,38 @@ FormattedAlignment AlignmentImpl::format_alignment() const
...
+    FormattedAlignment ret_formatted_alignment;
+
     for (auto const& x : alignment_)
     {
         switch (x)
         {
         case AlignmentState::match:
+            ret_formatted_alignment.second += target_[t_pos++];
+            ret_formatted_alignment.first += query_[q_pos++];
+            ret_formatted_alignment.pairing += "|";
```
ahehn-nv: Since you only append a single character I would I would use `'` instead of `"`.

cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
```diff
@@ -24,11 +24,16 @@ namespace cudaaligner
...
+/// FormattedAlignment -Holds formatted strings representing an alignment.
+typedef struct FormattedAlignment
+{
+    /// \brief FormattedAlignment.first = formatted string for query
+    std::string first;
+    /// \brief FormattedAlignment.pairing = formatted pairing string
+    std::string pairing;
+    /// \brief FormattedAlignment.first = formatted string for target
+    std::string second;
+} FormattedAlignment;
```
ahehn-nv: Would it make sense to use `query` and `target` instead of `first` and `second`? Or is that too specific?

cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
```diff
@@ -24,11 +24,16 @@ namespace cudaaligner
...
```
ahehn-nv: Since we use full struct now, I would suggest to add a
`std::ostream& operator << (std::ostream& os, FormattedAlignment const& formattedalignment)` for output.

There is one problem with output to the terminal: the output is garbled when the width is longer than the terminal width.
I would suggest to add a `linebreak_after` member to FormattedAlignment with default `0`, i.e. no line breaks, and use that in the `<<` operator.

Opinions? @ohadmo @tijyojwad

cudaaligner/tests/Test_AlignmentImpl.cpp
```diff
@@ -161,8 +161,10 @@ TEST_P(TestAlignmentImpl, AlignmentFormatting)
 {
     FormattedAlignment formatted_alignment = alignment_->format_alignment();
     std::string query                      = formatted_alignment.first;
+    std::string pairing                    = formatted_alignment.pairing;
     std::string target                     = formatted_alignment.second;
```
ahehn-nv: remove these local std::strings and use `formatted_alignment` directly. That will save 3 lines of code. :)

ohadmo: Thanks, I will an operator== to the struct!

ahehn-nv: no, an `==` is not necessary. You usually don't want to compare these objects.

cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
ohadmo: I agree

cudaaligner/include/claragenomics/cudaaligner/alignment.hpp
```diff
@@ -24,11 +24,23 @@ namespace cudaaligner
...
+typedef struct FormattedAlignment
+{
+    /// \brief FormattedAlignment.query = formatted string for query
+    std::string query;
+    /// \brief FormattedAlignment.pairing = formatted pairing string
+    std::string pairing;
+    /// \brief FormattedAlignment.target = formatted string for target
+    std::string target;
+    /// \brief Maximal line length when outputting to ostream (default = 0, no line break)
+    std::size_t linebreak_after = 0;
```
ahehn-nv: I would make this an `int32_t`, since it is a count and will never need to be so big that it can address every byte of memory. This would also match the argument of the `format_alignment(...)` method then.

cudaaligner/src/alignment.cpp
```diff
@@ -0,0 +1,45 @@
...
+        os << formatted_alignment.query.substr(last_line_start_idx, last_line_length) << '\n'
+           << formatted_alignment.pairing.substr(last_line_start_idx, last_line_length) << '\n'
+           << formatted_alignment.target.substr(last_line_start_idx, last_line_length) << '\n';
+    }
+    os << std::endl;
+    return os;
+}
+
+} // namespace cudaaligner
+} // namespace claragenomics
```
ahehn-nv: add a newline after `// namespace claragenomics` (I'm astonished that clang-format doesn't take care of that...)

cudaaligner/src/alignment.cpp
```diff
@@ -0,0 +1,45 @@
...
+{
+
+namespace cudaaligner
+{
+
+std::ostream& operator<<(std::ostream& os, const FormattedAlignment& formatted_alignment)
+{
+    std::size_t line_length              = (formatted_alignment.linebreak_after == 0) ? formatted_alignment.query.size() : formatted_alignment.linebreak_after;
+    std::size_t number_of_complete_lines = (std::size_t)(formatted_alignment.query.size() / line_length);
+    for (std::size_t i = 0; i < number_of_complete_lines; i++)
```
ahehn-nv: Good logic, but you should be able to simplify this function:
I looked up `std::string::substr(pos, length)`. It's fine if `length` is beyond the end of the string. It will just return the remainder then.
So how about just one loop
```
for (std::size_t i=0; i < formatted_alignment.query.size(); i+=line_length)
{
...substr(i, line_length)...
}
```
?

cudaaligner/src/alignment.cpp
ohadmo: There doesn't seem to be such an option.
Surprisingly, gcc doesn't throw a warning since C++11 as well.
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=68994

PR #227, opened 2019-11-20
[cudapoa] python API for cudapoa graph

tijyojwad:
Addresses #157


pyclaragenomics/claragenomics/bindings/cudapoa.pyx
```diff
@@ -226,6 +229,43 @@ cdef class CudaPoaBatch:
...
+                weight = edges[e].second
+                nx_digraph.add_edge(edge.first,
+                                    edge.second,
+                                    weight=weight)
+            attributes = {}
+            for n in nx_digraph.nodes:
+                attributes[n] = {'label' : deref(graph).get_node_label(n).decode('utf-8')}
+            nx.set_node_attributes(nx_digraph, attributes)
+            nx_digraphs.append(nx_digraph)
+        return (nx_digraphs, status)
```
ahehn-nv: This is just an idea: If we expose our Graph class to python anyways, why don't we return just that and provide a function to convert it to an nx.DiGraph() separately? At larger scales it might save the user an expensive construction, if the additional features of networkx are not needed.

pyclaragenomics/claragenomics/bindings/graph.pxd
```diff
@@ -0,0 +1,39 @@
...
+# license agreement from NVIDIA CORPORATION is strictly prohibited.
+#
+
+# cython: profile=False
+# distutils: language = c++
+# cython: embedsignature = True
+# cython: language_level = 3
+
+from libcpp.pair cimport pair
+from libcpp.memory cimport unique_ptr
```
ohadmo: Remove the unused cimports: unique_ptr, int8_t, int16_t, uint16_t

pyclaragenomics/claragenomics/bindings/cudapoa.pyx
tijyojwad: @ahehn-nv For python bindings there are two levels of exposing - one is where the C++ class is exposed in the binding, and the other is a custom python class that wraps that binding to expose to normal python usage. Right now I only implemented the first, which let's other bindings use the data type. The reason I didn't want to add another python class wrapping the graph class was mostly a maintainability thing. The primary use case for looking at the graph is for debugging, so the copy overhead won't be much of an issue. And users have more complex debugging functions available out of the box without having to copy things themselves. In time I'd like to maybe even replace our C++ graph API to a cuGraph compatible thing.

P.S. - I chose networkx as the interface coz it seemed widely used, had the dot serialization function, and I saw that used as a common graph lib at the hackathon :).

pyclaragenomics/claragenomics/bindings/graph.pxd
tijyojwad: Done

tijyojwad: And good catch! I forgot that our flake8 doesn't catch cython formatting issues

pyclaragenomics/claragenomics/bindings/cudapoa.pyx
ahehn-nv: Ok, makes sense. Thanks for the explanation. :)

vellamike:
@tijyojwad should this have been merged?

PR #229, opened 2019-11-21
[cudamapper] Using new index/matcher

mimaric:
Switching from old to new indexer and matcher.

@vellamike @ahehn-nv please review
@akkamesh @tijyojwad FYI

See #216

### DIFF
--- a/cudamapper/src/main.cu
+++ b/cudamapper/src/main.cu
@@ -164,11 +164,11 @@ int main(int argc, char* argv[])
     std::chrono::milliseconds matcher_time    = std::chrono::duration_values<std::chrono::milliseconds>::zero();
     std::chrono::milliseconds overlapper_time = std::chrono::duration_values<std::chrono::milliseconds>::zero();
 
-    for (size_t query_start = 0; query_start < queries; query_start += index_size)
+    for (size_t query_start_index = 0; query_start_index < queries; query_start_index += index_size)
     { // outer loop over query
-        size_t query_end = std::min(query_start + index_size, static_cast<size_t>(queries) - 1);
+        size_t query_one_past_the_end_index = std::min(query_start_index + index_size, static_cast<size_t>(queries));
 
-        std::cerr << "Query range: " << query_start << " - " << query_end << std::endl;
+        std::cerr << "Query range: " << query_start_index << " - " << query_one_past_the_end_index - 1 << std::endl;
 
         std::unique_ptr<claragenomics::cudamapper::IndexTwoIndices> query_index(nullptr);
         std::unique_ptr<claragenomics::cudamapper::IndexTwoIndices> target_index(nullptr);
@@ -178,34 +178,34 @@ int main(int argc, char* argv[])
             CGA_NVTX_RANGE(profiler, "generate_query_index");
             auto start_time = std::chrono::high_resolution_clock::now();
             query_index     = claragenomics::cudamapper::IndexTwoIndices::create_index(*query_parser,
-                                                                                   query_start,
-                                                                                   query_end + 1, // <- past the last
+                                                                                   query_start_index,
+                                                                                   query_one_past_the_end_index,
                                                                                    k,
                                                                                    w);
             index_time += std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time);
             std::cerr << "Query index generation time: " << std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time).count() << "ms" << std::endl;
         }
 
-        size_t target_start = 0;
+        size_t target_start_index = 0;
         // If all_to_all mode, then we can optimzie by starting the target sequences from the same index as
         // query because all indices before the current query index are guaranteed to have been processed in
         // a2a mapping.
         if (all_to_all)
         {
-            target_start = query_start;
+            target_start_index = query_start_index;
         }
-        for (; target_start < targets; target_start += target_index_size)
+        for (; target_start_index < targets; target_start_index += target_index_size)
         {
-            size_t target_end = std::min(target_start + target_index_size, static_cast<size_t>(targets) - 1);
+            size_t target_one_past_the_end_index = std::min(target_start_index + target_index_size, static_cast<size_t>(targets));
 
-            std::cerr << "Target range: " << target_start << " - " << target_end << std::endl;
+            std::cerr << "Target range: " << target_start_index << " - " << target_one_past_the_end_index - 1 << std::endl;
 
             {
                 CGA_NVTX_RANGE(profiler, "generate_target_index");
                 auto start_time = std::chrono::high_resolution_clock::now();
                 target_index    = claragenomics::cudamapper::IndexTwoIndices::create_index(*target_parser,
-                                                                                        target_start,
-                                                                                        target_end + 1, // <- past the last
+                                                                                        target_start_index,
+                                                                                        target_one_past_the_end_index,
                                                                                         k,
                                                                                         w);
                 index_time += std::chrono::duration_cast<std::chrono::milliseconds>(std::chrono::high_resolution_clock::now() - start_time);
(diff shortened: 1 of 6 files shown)


mimaric:
rerun tests

mimaric:
@vellamike @ahehn-nv  If you haven't started doing the review yet for a little more, I'd like to remove old index and matcher and rename the new one in this PR

mimaric:
@vellamike @ahehn-nv You can review now

cudamapper/tests/mock_index.cuh
```diff
@@ -12,19 +12,30 @@
...
+                             0,
+                             0,
+                             0,
+                             0,
+                             true)
+    {
+    }
+
+    MOCK_METHOD(const std::string&, read_id_to_read_name, (const read_id_t read_id), (const override));
+    MOCK_METHOD(const std::uint32_t&, read_id_to_read_length, (const read_id_t read_id), (const, override));
```
ahehn-nv: should this really be `(const, override)` with a comma?

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
```diff
@@ -169,17 +169,18 @@ TEST(TestCudamapperOverlapperTriggerred, OneAchorNoOverlaps)
     testv.push_back("READ2");
     std::vector<std::uint32_t> test_read_length(testv.size(), 1000);
 
-    EXPECT_CALL(test_index, read_id_to_read_name)
-        .WillRepeatedly(testing::ReturnRef(testv));
-    EXPECT_CALL(test_index, read_id_to_read_length)
-        .WillRepeatedly(testing::ReturnRef(test_read_length));
+    for (std::size_t i = 0; i < testv.size(); ++i)
```
ahehn-nv: Optional: Just because I would like to establish more consistent style: could you use the following?
```
for (std::int64_t i = 0; i < get_size(testv); ++i)
```
(also in the following loops)

cudamapper/tests/mock_index.cuh
mimaric: Apparently yes:
`Since you're overriding a virtual method, we suggest adding the override keyword. For const methods the 4th parameter becomes (const, override), for non-const methods just (override). This isn't mandatory.`
https://github.com/google/googletest/blob/master/googlemock/docs/for_dummies.md#how-to-define-it

cudamapper/src/main.cu
```diff
@@ -154,33 +157,34 @@ int main(int argc, char* argv[])
...
-        // Add overlaps for All-vs-all for chunk A
-        // Add overlaps for Chunk A vs Chunk B
-        // Add overlaps for Chunk A vs Chunk C
-        // Add overlaps for All-vs-all for chunk B
-        // Add overlaps for Chunk B vs Chunk C
-        // Add overlaps for All-vs-all for chunk C
-        std::pair<std::uint64_t, std::uint64_t> query_range{query_start, query_end};
-
-        auto overlapper = claragenomics::cudamapper::OverlapperTriggered();
+        size_t query_end = std::min(query_start + index_size, static_cast<size_t>(queries) - 1);
```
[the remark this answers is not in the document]
mimaric: Yes, that's a bug. Thanks!

cudamapper/src/main.cu
```diff
@@ -154,33 +157,34 @@ int main(int argc, char* argv[])
...
+        std::unique_ptr<claragenomics::cudamapper::IndexTwoIndices> query_index(nullptr);
+        std::unique_ptr<claragenomics::cudamapper::IndexTwoIndices> target_index(nullptr);
+        std::unique_ptr<claragenomics::cudamapper::MatcherTwoIndices> matcher(nullptr);
+
+        {
+            CGA_NVTX_RANGE(profiler, "generate_query_index");
+            auto start_time = std::chrono::high_resolution_clock::now();
+            query_index     = claragenomics::cudamapper::IndexTwoIndices::create_index(*query_parser,
+                                                                                   query_start,
+                                                                                   query_end + 1, // <- past the last
```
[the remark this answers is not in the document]
mimaric: I agree, mixing two styles caused the bug above

cudamapper/src/main.cu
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/main.cu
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/main.cu
```diff
@@ -191,65 +195,37 @@ int main(int argc, char* argv[])
...
-
-            start_time = std::chrono::high_resolution_clock::now();
-
-            std::vector<std::pair<std::uint64_t, std::uint64_t>> ranges;
-            std::vector<claragenomics::io::FastaParser*> parsers;
-
-            ranges.push_back(query_range);
-            parsers.push_back(query_parser.get());
+        {
+            size_t target_end = std::min(target_start + target_index_size, static_cast<size_t>(targets) - 1);
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/main.cu
```diff
@@ -191,65 +195,37 @@ int main(int argc, char* argv[])
...
             {
-                // Only add a new range if it is not the case that mode is all_to_all and ranges between target and query match.
-                std::pair<std::uint64_t, std::uint64_t> target_range{target_start, target_end};
-                ranges.push_back(target_range);
-                parsers.push_back(target_parser.get());
+                CGA_NVTX_RANGE(profiler, "generate_target_index");
+                auto start_time = std::chrono::high_resolution_clock::now();
+                target_index    = claragenomics::cudamapper::IndexTwoIndices::create_index(*target_parser,
+                                                                                        target_start,
+                                                                                        target_end + 1, // <- past the last
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
mimaric: I generally agree with the idea of using signed values for loops, but let's do this in a separate PR (there's a lot of signed-unsigned things we should address)

cudamapper/src/main.cu
```diff
@@ -154,102 +157,75 @@ int main(int argc, char* argv[])
...
-        // Add overlaps for Chunk A vs Chunk C
-        // Add overlaps for All-vs-all for chunk B
-        // Add overlaps for Chunk B vs Chunk C
-        // Add overlaps for All-vs-all for chunk C
-        std::pair<std::uint64_t, std::uint64_t> query_range{query_start, query_end};
-
-        auto overlapper = claragenomics::cudamapper::OverlapperTriggered();
-
-        size_t target_start = 0;
+        size_t query_one_past_the_end_index = std::min(query_start_index + index_size, static_cast<size_t>(queries));
```
ahehn-nv: just call it `query_end_index` please. This is such a common pattern in C++, that the verbosity of _one_past_the_end should really not be necessary.

I've seen people use `_last` very often if they meant to include the element.

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
ahehn-nv: I just meant because you edit this loop anyways. I didn't want you to fix the other loops, but ok, let's do a separate PR.

cudamapper/tests/mock_index.cuh
ahehn-nv: Ok, but then the line above should have a comma too.

mimaric: So they do support both, but only document one...
Done

cudamapper/src/main.cu
mimaric: This one is a really tricky to decide. I agree that to every C++ programmer it should be clear that container::end() is one_past_the_end, but `query_end_index` can mean both things, probably even leaning towards it being actual last valid index...

mimaric: done

PR #231, opened 2019-11-22
[cudamapper] Remover old index and matcher implementations

mimaric:
See #216 

@vellamike @tijyojwad FYI


cudamapper/src/main.cu
```diff
@@ -72,11 +70,11 @@ int main(int argc, char* argv[])
         }
     }
 
-    if (k > claragenomics::cudamapper::Index::maximum_kmer_size())
+    /*if (k > claragenomics::cudamapper::Index::maximum_kmer_size())
```
ahehn-nv: Do you plan to reintroduce this commented lines? If note, just remove them completely.

mimaric: Yes, they'll be reintroduced in the next PR (had to split them in two because GitHub gets confused when some a filed get removed and another gets renamed to that name)

ahehn-nv: ok

PR #232, opened 2019-11-22
[cudamapper] Renamed new index and matcher

mimaric:
Renamed:
`IndexTwoIndices` -> `Index`
`IndexGPUTwoIndices` -> `IndexGPU`
`MatcherTwoIndices` -> `Matcher`

It also makes sure that public header libraries are included correctly

@vellamike @tijyojwad FYI

Closes #216


mimaric:
> you still have `index_gpu_two_indices` namespaces, I guess they should be renamed as well.

Whoops, missed that one....

PR #238, opened 2019-11-25
[pyclaragenomics] Create wheel package from pyclaragenomics

ohadmo:


pyclaragenomics/setup_pyclaragenomics.py
```diff
@@ -104,9 +134,13 @@ def setup_python_binding(is_develop_mode, pycga_dir, cga_install_dir):
                               cga_install_dir=cga_installation_directory,
                               cmake_extra_args="-Dcga_build_shared=ON")
     cmake_proj.build()
+    # Copyies shared libraries into clargenomics package
```
vellamike: typo - copies

pyclaragenomics/setup.py
```diff
@@ -11,6 +11,7 @@
 #
 
 import os
+import glob
```
vellamike: Imports should be ordered with stdlib first and 3rd party after, alphabetical within groups like so:

```
import glob
import os
from setuptools import setup, find_packages, Extension

from Cython.Build import cythonize
```

pyclaragenomics/setup_pyclaragenomics.py
```diff
@@ -75,14 +95,33 @@ def build(self):
...
+    if wheel_output_folder:
+        # Copies shared libraries into clargenomics package
+        copy_all_files_in_directory(
+            os.path.join(cga_install_dir, "lib"),
+            os.path.join(pycga_dir, "claragenomics/shared_libs/"),
+        )
+        setup_command = ['python', 'setup.py', 'bdist_wheel', '-d', wheel_output_folder]
+        completion_message = \
+            "A wheel file was create for pyclaragenomics under {}".format(wheel_output_folder)
+        cga_runtime_lib_dir = os.path.join('$ORIGIN', os.pardir, 'shared_libs')
```
tijyojwad: I'm wondering now, for the case of simplicity, why don't we use this setup for both cases? i.e. copy the shared libs over into the shared_libs folder, and then call either pip or python setup bdist. the way we have one less path setting to maintain. and we can add a .gitignore file which masks the shared_libs folder under pyclaragenomics so those copied files don't show up in the git index. what do you think?

pyclaragenomics/setup_pyclaragenomics.py
```diff
@@ -14,6 +14,22 @@
 import os.path
 import os
 import subprocess
+import shutil
+import glob
```
tijyojwad: reorder glob position to maintain alphabetical order

pyclaragenomics/setup_pyclaragenomics.py
tijyojwad: this will of course require the user to copy over the libs into the shared_libs folder location if they wanted to call `setup.py` by itself (without using our wrapper). I think if that case comes up (I expect it won't anytime soon) we can re-evaluate. We should just push for our users to run the wrapper script and not worry about calling things directly

tijyojwad: or can the shared lib copying be moved into the `setup.py` script which copes files from he environment variable location?

ohadmo: @tijyojwad, the reason I didn't copy the files when installing through 'pip install' is that setuptools setup() ignores the 'package_data'  field when a package is not created (I can try to use data_files instead). Also, I already distinguish between 'pip install' and 'setup bdist_wheel' in setup_pyclaragenomics, so I did not want to do this again in setup.py

ohadmo: @tijyojwad  You are right about copying the shared libraries copy setup.py, it makes more sense.

ohadmo: @tijyojwad, @vellamike, I'm also considering using 'pip wheel' instead of 'python setup.py bdist_wheel'.
The wheel package leftovers- 'build/'  & 'pyclaragenomics.egg-info/'  folders under pyclaragenomics cause an unexpected behaviour if not deleted in the following scenario:
1. run `python setup_pyclaragenomics.py  --create_wheel_only`
2. run `python setup_pyclaragenomics.py `
now examine the `rpath` of the binding so files using `objdump` command and you can see they point to `$ORIGIN/../shared_libs` which is the `rpath` of the wheel creation int he previous command. pip perform this in a temporary folder under `/tmp` which does not leave leftovers under the repository folder.

pyclaragenomics/setup.py
```diff
@@ -30,29 +54,61 @@ def get_verified_path(path):
...
+    'Programming Language :: Python :: 3.7',
+    'Programming Language :: Python :: 3.8',
+    'Programming Language :: Python :: 3.9',
+],
 
 extensions = [
     Extension(
         "*",
-        sources=[os.path.join(pycga_dir, "claragenomics/**/*.pyx")],
+        sources=[os.path.join("claragenomics/**/*.pyx")],
```
ohadmo: Can not use an absolute path along with `include_package_data`

pyclaragenomics/setup.py
```diff
@@ -30,29 +54,61 @@ def get_verified_path(path):
...
       packages=find_packages(where=pycga_dir),
+      python_requires='>=3.6',
+      license='Apache License 2.0',
+      long_description='Python libraries and utilities for manipulating genomics data',
+      classifiers=pycga_classifiers,
       ext_modules=cythonize(extensions, compiler_directives={'embedsignature': True}),
-      scripts=[get_verified_path(os.path.join(pycga_dir, 'bin', 'genome_simulator')),
-               get_verified_path(os.path.join(pycga_dir, 'bin', 'assembly_evaluator'))],
+      scripts=[os.path.join('bin', 'genome_simulator'),
+               os.path.join('bin', 'assembly_evaluator')],
```
ohadmo: Can not use an absolute path along with `include_package_data`

PR #239, opened 2019-11-26
[pyclaragenomics] `paf_eval` bugfix and lint

vellamike:
* Fixed bug whereby if query and target were reversed in order overlaps
were not being evaluated
* Linted for PEP8 compliance
* Added printing of TP,FP and FN absolute numbers

### DIFF
--- a/pyclaragenomics/bin/evaluate_paf
+++ b/pyclaragenomics/bin/evaluate_paf
@@ -34,6 +34,16 @@ def match_overlaps(query_0, query_1, target_0, target_1, pos_tolerance):
         abs(target_0[0] - target_1[0]) < pos_tolerance and \
         abs(target_0[1] - target_1[1]) < pos_tolerance
 
+def generate_key(name_1, name_2):
+    """Given two read names, return key for indexing overlaps.
+
+    Args:
+        name_1 (str) : Name of first read
+        name_2 (str): Name of second read
+
+    Returns: a key of concatenated names.
+    """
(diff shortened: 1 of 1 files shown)


pyclaragenomics/bin/evaluate_paf
```diff
@@ -62,6 +64,10 @@ def evaluate_paf(truth_paf_filepath, test_paf_filepath, pos_tolerance=400, skip_
     false_positive_count = 0
     false_negative_count = 0
 
+    print("Counted {} true overlaps".format(num_true_overlaps))
+
+    seen_test_overlap_keys = set()
```
vellamike: @tijyojwad this is added in case there are duplicate overlaps in the PAF file, which would mess up the statistics otherwise.

pyclaragenomics/bin/evaluate_paf
```diff
@@ -53,7 +55,7 @@ def evaluate_paf(truth_paf_filepath, test_paf_filepath, pos_tolerance=400, skip_
                 (truth_overlap.query_sequence_name == truth_overlap.target_sequence_name):
             continue
 
-        key = truth_overlap.query_sequence_name + truth_overlap.target_sequence_name
+        key = truth_overlap.query_sequence_name + "_" + truth_overlap.target_sequence_name
```
tijyojwad: Can you move the key generation to a function that appends name1 and name1 with an underscore? Should have done it last time to avoid this bug in the first place

PR #251, opened 2019-12-02
[cudamapper] Perf improvements for cudamapper

tijyojwad:
PR from @vellamike . Grabbing highlights from commit messages.

* Multi-GPU support
* Index caching
* Improved multi-threaded overlap writing


tijyojwad:
rerun tests

cudamapper/src/main.cu
```diff
@@ -105,148 +118,177 @@ int main(int argc, char* argv[])
...
+            std::int32_t target_end_index = std::min(target_start_index + target_index_size,
+                                                     static_cast<size_t>(targets));
+            q.target_ranges.push_back(std::make_pair(target_start_index, target_end_index));
+        }
+
+        query_target_ranges.push_back(q);
+    }
+
+    // This is a per-device cache, if it has the index it will return it, if not it will generate it, store and return it.
+    std::vector<std::map<std::pair<uint64_t, uint64_t>, std::shared_ptr<std::unique_ptr<claragenomics::cudamapper::Index>>>> index_cache(num_devices);
```
ahehn-nv: `std::shared_ptr<std::unique_ptr<...>>` seems redundant, do you need that? Otherwise you can construct a shared pointer directly from an unique pointer via `std::shared_ptr<...> p(std::move(unique_p));`.

vellamike: The reason I am using `std::shared_ptr` is that I was unable to populate the vector with unique_ptrs, (since they don't implement a copy constructor being unique). 

Could I just use `std::unique_ptr<...>` instead by using std::move ?

mimaric: You could, but you'd end up with a similar problem. When you try to use an `Index` instance you could not just say "std::unique_ptr<Index> index = foo_map[bar]" as you'd try to copy an `unique_ptr`. In this case you're trying to share index between several matchers and the caching container, so `shared_ptr` is the right choice. @ahehn-nv is right in saying that you should just move that `unique_ptr` into initial `shared_ptr`

vellamike: OK that makes sense

vellamike: Done

common/io/src/hts_fasta_parser.cpp
```diff
@@ -60,29 +60,37 @@ int32_t FastaParserHTS::get_num_seqences() const
     return num_seqequences_;
 }
 
-FastaSequence FastaParserHTS::get_sequence_by_id(int32_t i) const
```
mimaric: Why can't this method be `const`? Mutex? In that case you can use `mutable`

cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -88,7 +88,7 @@ class Index
     /// \param hash_representations - if true, hash kmer representations
     /// \return instance of Index
     static std::unique_ptr<Index>
-    create_index(const io::FastaParser& parser,
```
mimaric: Same question about `const`

common/io/src/hts_fasta_parser.cpp
vellamike: This feels like more a matter of opinion :)

cudamapper/include/claragenomics/cudamapper/index.hpp
vellamike: As above, I don't think this is right or wrong, my preference is to not use `mutable`

cudamapper/src/main.cu
ahehn-nv: I thought you really want shared_ptr semantics (i.e. shared ownership of the object), if there will always be just a single pointer that has the ownership, I would use `std::unique_ptr` and construct the vector with `std::move`.

common/io/src/hts_fasta_parser.cpp
mimaric: Partially, but const-correctness is a very important thing and should be done right whenever possible. Currently we don't mind passing a non-const `FastaParser` to `Index`, but this could change in the future. Also, using ``mutable` for mutex is a rather common practice in this case in C++, it's not considered hacky.
@ahehn-nv, what do you think?

cudamapper/src/main.cu
```diff
@@ -105,148 +118,171 @@ int main(int argc, char* argv[])
...
-        std::chrono::high_resolution_clock::now() - start_time);
-
-    std::cerr << "\n\n"
-              << std::endl;
-    std::cerr << "Index execution time: " << index_time.count() << "ms" << std::endl;
-    std::cerr << "Matcher execution time: " << matcher_time.count() << "ms" << std::endl;
-    std::cerr << "Overlap detection execution time: " << overlapper_time.count() << "ms" << std::endl;
-    std::cerr << "PAF detection execution time: " << paf_time.count() << "ms" << std::endl;
+    // create thread pool to compute overlaps. One worker thread per device.
+    ThreadPool overlap_pool(num_devices);
```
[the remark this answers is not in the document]
vellamike: Yes, that's correct, I tried that but performance did not improve and memory consumption went up a lot. The cache would also need to be guarded by a mutex at that point.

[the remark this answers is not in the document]
mimaric: ok, thought so

common/io/src/hts_fasta_parser.cpp
ahehn-nv: I agree with @mimaric. If you look at the abstract logic of the class, this method leaves the object in the same state, except for an internal implementation detail. From the outside the object is unchanged.

I would expect a `const` method to be thread-safe though - but that's exactly what this detail is supposed to do.

tijyojwad: @ahehn-nv @mimaric reverted the methods to be `const` again. Reasons above convinced me that `mutable` is the way to go. abstract API indeed remains const

mimaric: Looks good, marking as resolved

cudamapper/include/claragenomics/cudamapper/index.hpp
mimaric: @tijyojwad fixed this, marking as resolved

cudamapper/src/main.cu
```diff
@@ -105,148 +118,171 @@ int main(int argc, char* argv[])
...
+        std::vector<std::pair<std::int32_t, int32_t>> target_ranges;
+    };
+
+    //First generate all the ranges independently, then loop over them.
+    std::vector<query_target_range> query_target_ranges;
+
+    for (std::int32_t query_start_index = 0; query_start_index < queries; query_start_index += index_size)
+    {
+
+        std::int32_t query_end_index = std::min(query_start_index + index_size, static_cast<size_t>(queries));
```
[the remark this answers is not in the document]
vellamike: done

cudamapper/src/main.cu
```diff
@@ -105,148 +118,171 @@ int main(int argc, char* argv[])
...
+                                                     int device_id) {
+        std::pair<uint64_t, uint64_t> key;
+        key.first  = query_start_index;
+        key.second = query_end_index;
+
+        std::shared_ptr<claragenomics::cudamapper::Index> index;
+
+        if (index_cache[device_id].count(key))
+        {
+            index = index_cache[device_id][key];
```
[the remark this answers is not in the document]
vellamike: Very good point. Caching doesn't even make sense int he query-target situation. I have added a check so that entries are only added to the cache in the case of all_to_all.

cudamapper/src/main.cu
```diff
@@ -105,148 +118,171 @@ int main(int argc, char* argv[])
...
-            std::this_thread::yield();
         }
+        return index;
     };
-    std::future<void> overlap_result(std::async(std::launch::async, overlaps_writer_func));
 
-    claragenomics::cudamapper::OverlapperTriggered overlapper;
+    auto evict_index = [&index_cache](
+                           const claragenomics::cudamapper::read_id_t query_start_index,
+                           const claragenomics::cudamapper::read_id_t query_end_index,
```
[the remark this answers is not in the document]
vellamike: No, given our caching strategy only a query range is ever evicted from the cache (although a cache entry that is currently be being used as a query could have been previously a target). The code could probably be tidied up quite a bit as I don't like how the logic of cache is tied up with the logic of main, I will clarify in a comment.

vellamike:
> I spotted two potential bugs (not sure if the second one is actually a bug):
> 
> 1. read_ids of both query and target start from 0. This means that if we're not doing all-to-all cache could return query as target and target as query. I've been thinking if we should fix this in general (by setting a starting read_id for each index), but I'm still not sure if that's necessary

Agree. I have disabled caching for non-all-to-all (it isn't useful for this anyway)

> 2. `get_index` and `evict_index` are not thread-safe. I have a feeling you made sure that only one thread is active for each device at a time with the way you're dealing with `ThreadPool` at the end of the file, but that part is poorly documented so I'm not sure if I got it right. I think we should think about it a bit more, especially if we want to ready for using streams/multiple thread per device

I have documented the thread-per-device behaviour and also added mutexes to guard the cache.

> 
> Other comments are mainly about const-correctness and namings, so nothing crucial.

all addressed 
> 
> I would suggest documenting the code better, especially the multi-threading part

mimaric:
rerun tests

cudamapper/src/main.cu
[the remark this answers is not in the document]
mimaric: Why does it not make sense? Let's say query is split into chunks/indices q1, q2, q3 and target in t1, t2, t3, t4, t5. Without chunking t1, t2, t3, t4, t5 would be generated three times (once for each query chunk)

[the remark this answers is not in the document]
vellamike: You are right, the caching strategy we are using here however wouldn't make sense, - in a "standard" query vs target matching we should only ever cache the target reads and there should be no eviction from the cache at all (since all query indices are used only once and all target indices are used the same number of times).

cudamapper/src/main.cu
```diff
@@ -116,152 +129,200 @@ int main(int argc, char* argv[])
...
+    ThreadPool overlap_pool(num_devices);
+
+    // 2. Enqueue the query-target ranges which need to be computed, each thread returns a vector of futures for the threads it launches
+    std::vector<std::future<std::vector<std::shared_ptr<std::future<void>>>>> overlap_futures;
+    for (int i = 0; i < query_target_ranges.size(); i++)
+    {
+        // assign chunk pairs (Query, target) to device
+        auto query_target_range = query_target_ranges[i];
+        auto device_id          = i % num_devices;
+        overlap_futures.push_back(overlap_pool.enqueue(compute_overlaps, query_target_range, device_id));
```
mimaric: I'm still not sure if this works as expected. Imagine you have 2 GPUs and 2 threads in the pool and 4 query ranges.
You'd assigned them in the following way:
range_0 -> device_0
range_1 -> device_1
range_2 -> device_0
range_3 -> device_1
Threads are taking ranges in the order they entered the queue. If we have two threads that means most likely you'd have:
thread_0 -> range_0 -> device_0
thread_1 -> range_1 -> device_1
Now imagine thread_1 finishes before thread_0 and picks up the next task, which is range_2. Now you'd have:
thread_0 -> range_0 -> device_0
thread_1 -> range_2 -> device_0
meaning there are two threads working on device_0 and none on device_1.
 Am I missing some property of `ThreadPool` or something else?

vellamike: This is true,in general I am making the assumption that it only has quite a small effect, do you have any suggestions for a clean way to make a worker thread assign its work to whichever GPU is available?

cudamapper/src/main.cu
```diff
@@ -116,152 +129,200 @@ int main(int argc, char* argv[])
...
+
+            // If in all-to-all mode, put this query in the cache for later use.
+            // Cache eviction is handled later on by the calling thread
+            // using the evict_index function.
+            if (index_cache[device_id].size() < max_cache_size && allow_cache_index)
             {
-                break;
+                index_cache_mtx.lock();
+                index_cache[device_id][key] = index;
+                index_cache_mtx.unlock();
```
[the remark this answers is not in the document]
vellamike: I will use `std::unique_lock` since it has (IMO) nicer semantics

cudamapper/src/main.cu
mimaric: I have a different idea. How about creating equal number of threads amd GPUs and assigning each thread to a device. Then a thread whenever it's done with its current query_index could pick up the next available query_index and work on it (this pool of query_indices should obviously be implemented in a thread-safe manner).
This essentially means assigning each thread a GPU instead of assigning each query_index a GPU

vellamike: That's a good idea - I will implement it this way, then I can also remove all the mutexes around the cache since I can guarantee that there will only be one thread per device.

cudamapper/src/main.cu
```diff
@@ -116,152 +128,206 @@ int main(int argc, char* argv[])
...
+    {
+        workers.push_back(std::thread(
+            [&, device_id]() {
+                while (ranges_idx < query_target_ranges.size())
+                {
+                    int range_idx = ranges_idx.fetch_add(1);
+                    //Need to perform this check again for thread-safety
+                    if (range_idx < query_target_ranges.size())
+                    {
+                        overlap_futures.push_back(compute_overlaps(query_target_ranges[range_idx], device_id));
```
mimaric: I think this is not thread-safe. Multiple threads are pushing to the same vector.

vellamike: Agreed, resolved

PR #256, opened 2019-12-04
Make pafio.py preserve SAM style tags

alexomics:
This is a complete re-write of this module that I am proposing. It is an extended version of my own `readpaf.py` that I wrote a while back. See https://github.com/alexomics/read-paf/blob/master/readpaf.py

Notably, this introduces the ability to preserve the SAM tags that minimap2 includes. It also handles type conversion of the tags and will write them back out again.

I have tried to keep the functionality as close to what it was before, but have changed a few things:
 - namedtuple changed from `Overlaps` to `PAF`
 - namedtuple fields are configurable


GPUtester:

tijyojwad:
Hi @alexomics thanks for the awesome changes. We'll get to reviewing it shortly!

vellamike:
Hi @alexomics this looks great, thank you! I haven't done a full review yet but one thing that I noticed is that your docstrings are not the [google-style](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) that we use. Would it be possible to change these?

alexomics:
@vellamike I've updated the docstrings to google style. 👍

tijyojwad:
add to whitelist

pyclaragenomics/claragenomics/io/pafio.py
```diff
@@ -1,117 +1,169 @@
...
+            "alignment_block_length",
+            "mapping_quality",
+            "tags",
+        ]
+
+    with open(filepath, "r") as fh:
+        return list(_paf_generator(fh, fields=fields))
+
+
+def write_paf(records, filepath):
```
tijyojwad: @alexomics one of the use cases is to pass in a list of custom overlaps that can be written out to a paf file. In this API, it looks like we need the PAF namedtuple that is only returned by the `read_paf` function. Could we add the capability to instantiate a PAF namedtuple outside of the read_paf function?

e.g. https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/pyclaragenomics/claragenomics/simulators/readsim.py#L81

tijyojwad:
@alexomics our tests currently don't cover all the places where `pafio` is used, so there are a couple of places where the namedtuple members have to be updated to work with the new pafio. Is it okay if I push those changes to your branch?

alexomics:
Yes, go ahead

pyclaragenomics/claragenomics/io/pafio.py
alexomics: Yes!

alexomics: My latest commits (9a7a4ae, fbbc131) should clear this up.

alexomics: I have rolled back to using the same (hard coded) fields from before my pull req. I have only added a `tags` field at the end. The namedtuple has changed back to `Overlap` and is importable again.

vellamike:
rerun tests

PR #260, opened 2019-12-06
[env] update cmake to look at conda env when available

tijyojwad:
Once this goes in, we can take out this line from our custom dockerfiles - https://github.com/clara-genomics/gpuci-build-environment/blob/master/Dockerfile#L29

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -62,6 +62,7 @@ include(cmake/Tests.cmake)
 include(cmake/Benchmarks.cmake)
 include(cmake/Format.cmake)
 include(cmake/Packaging.cmake)
+include(cmake/CondaEnv.cmake)
 
 # Add ClaraGenomicsAnalysis projects.
 add_subdirectory(common/logging)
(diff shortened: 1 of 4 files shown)


tijyojwad:
@ohadmo good point. I've removed specific instructions for installation right now, but since we don't bundle htslib with the repo right now still keeping it in the README. Eventually want to just make it a submodule of CGA

ohadmo:
@tijyojwad  after rethinking about this, perhaps we should not change the README file at all?
Just as we refer the user to the official site for downloading clang-format/Doxygen (even though they are installed through anaconda when preparing the CI environment). Unlike, racon, minimap2 & miniasm which are pyclaragenomics dependencies. What do you think?

tijyojwad:
@ohadmo I think what's their currently is okay. For doxygen we don't specify where to get it from (we get it from conda, but many may want to install it from source). clang-format is a bit special because it was really hard to find a proper package from conda. Source was the best wya to do it. but for htslib users can choose what they want, as long as the version needs are met.

PR #263, opened 2019-12-13
Create intermediate Fasta index and bgzip files in a separate directory...

rilango:
The directory containing ".fasta" file, is the default location for creating(htslib) Fasta index and bgzip files. This is a problem in environments such as Clara deploy, where the directory containing the input files is read-only.

These change circumvent this issue by allowing creation of these files in a temp/output directory. Temp directory is an optional input parameter(-o).

### DIFF
--- a/common/io/CMakeLists.txt
+++ b/common/io/CMakeLists.txt
@@ -16,17 +16,19 @@ add_library(${PROJECT_NAME} ${cga_library_type}
         src/hts_fasta_parser.cpp)
 target_link_libraries(${PROJECT_NAME} PUBLIC hts)
 
+include_directories(../utils/include)
+
 add_doxygen_source_dir(${CMAKE_CURRENT_SOURCE_DIR}/include/claragenomics/io)
 
 target_include_directories(${PROJECT_NAME}
-    PUBLIC 
-        $<INSTALL_INTERFACE:include>    
+    PUBLIC
+        $<INSTALL_INTERFACE:include>
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
 )
 
 target_compile_options(${PROJECT_NAME} PRIVATE -Werror)
 
-install(TARGETS ${PROJECT_NAME} 
+install(TARGETS ${PROJECT_NAME}
     EXPORT ${PROJECT_NAME}
     DESTINATION lib
     INCLUDES DESTINATION include
--- a/common/io/src/hts_fasta_parser.cpp
+++ b/common/io/src/hts_fasta_parser.cpp
@@ -36,16 +36,17 @@ namespace io
 {
 
 FastaParserHTS::FastaParserHTS(const std::string& fasta_file,
-        const std::string& output_dir)
+                               const std::string& output_dir)
 {
     std::string fai_file = fasta_file + ".fai";
     std::string gzi_file = fasta_file + ".gzi";
     if (output_dir.length() != 0)
     {
         // resolve idx and gzi file path
-        if (!claragenomics::filesystem::dirExists(output_dir)) {
+        if (!claragenomics::filesystem::dirExists(output_dir))
+        {
             throw std::runtime_error("Output dir " + output_dir +
-                    " not found or not a directory!");
+                                     " not found or not a directory!");
         }
 
         std::string file_name =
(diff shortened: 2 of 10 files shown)


cudamapper/src/main.cu
```diff
@@ -363,6 +369,9 @@ void help(int32_t exit_code = 0)
               << R"(
         -F --filtering-parameter
             filter all representations for which sketch_elements_with_that_representation/total_sketch_elements >= filtering_parameter), filtering disabled if filtering_parameter == 1.0 [1'000'000'001] (Min = 0.0, Max = 1.0))"
+              << R"(
+        -o --output-dir
```
vellamike: This is confusing, the user would usually expect `-o --output-dir` to be the output location of all files not just intermediate ones. Why don't we output all files into this directory?

common/io/src/hts_fasta_parser.cpp
```diff
@@ -34,9 +35,30 @@ namespace claragenomics
...
+        // resolve idx and gzi file path
+        if (!claragenomics::filesystem::dirExists(output_dir))
+        {
+            throw std::runtime_error("Output dir " + output_dir +
+                                     " not found or not a directory!");
+        }
+
+        std::string file_name =
+            claragenomics::filesystem::resolveFileName(fasta_file);
+        fai_file = output_dir + "/" + file_name + ".fai";
```
vellamike: Is this a good way to be constructing file paths? We should be using `<filesystem>` like [described here](https://stackoverflow.com/a/36848326)

common/utils/include/claragenomics/utils/filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+#include <sys/stat.h>
+#include <string>
+
+namespace claragenomics
+{
+
+namespace filesystem
+{
+
+bool dirExists(const std::string path)
```
vellamike: We don't use camelCase anywhere in CGA, @tijyojwad  is this part of our official style guide?

common/utils/include/claragenomics/utils/filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+    int response = stat(path.c_str(), &info);
+    if (response != 0)
+    {
+        return false;
+    }
+
+    return (info.st_mode & S_IFDIR) ? true : false;
+}
+
+std::string resolveFileName(const std::string file_path)
```
vellamike: as above (re camelCase)

common/utils/include/claragenomics/utils/filesystem.hpp
vellamike: Why not just use [filesystem_exists](https://en.cppreference.com/w/cpp/filesystem/exists) ?

common/utils/include/claragenomics/utils/filesystem.hpp
vellamike: These functions require doxygen-style documentation (see other parts of CGA)

common/utils/include/claragenomics/utils/filesystem.hpp
```diff
@@ -0,0 +1,45 @@
...
+{
+    size_t found;
+    found = file_path.find_last_of("/\\");
```
vellamike: I don't understand this. Can you clarify in the documentation

common/io/CMakeLists.txt
```diff
@@ -16,17 +16,19 @@ add_library(${PROJECT_NAME} ${cga_library_type}
         src/hts_fasta_parser.cpp)
 target_link_libraries(${PROJECT_NAME} PUBLIC hts)
 
+include_directories(../utils/include)
```
mimaric: If it turns out we cannot replace `claragenomics/utils/filesystem.hpp` with `std::filesystem`:

`claragenomics/utils/filesystem.hpp` should be added to `utils` library (or turned into a separate library) and included properly

tijyojwad:
@rilango Since these changes are mostly for supporting deploy SDK, we can add a requirement right now that the index needs to be generated before running the deploy pipeline. It's likely that we'll move away from htslib soon in which case we won't need a tmpdir for the index files anyway. So let's defer these changes for now till we stabilize on the fasta IO library.

rilango:
This change may be not required as we are moving away from hts based parser.

PR #264, opened 2019-12-18
[cudamapper] fix read skipping index

tijyojwad:
Properly handle skipping of reads in index.

Ran into this when I was experimenting with skipping reads greater than some number in the index.

### DIFF
--- a/cudamapper/src/index_gpu.cuh
+++ b/cudamapper/src/index_gpu.cuh
@@ -602,7 +602,7 @@ void IndexGPU<SketchElementImpl>::generate_index(const io::FastaParser& parser,
         return;
     }
 
-    number_of_reads_ = past_the_last_read_id - first_read_id;
+    number_of_reads_ = 0;
 
     std::uint64_t total_basepairs = 0;
     std::vector<ArrayBlock> read_id_to_basepairs_section_h;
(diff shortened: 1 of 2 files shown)


mimaric:
Was this causing some bug or you decided to fix because of that comment?
I believe that (currently commented out) test `TestCudamapperIndexGPU, CATCAAG_AAGCTA_3_5` tests this. Can you uncomment it and see if it works?
This part was causing some problems due to the way main.cu works so we should be rather careful with it. I think that one of the cases I had in mind here was being future-proof for the same where we'd on-the-fly change index sizes to prevent OOM.

tijyojwad:
@mimaric it was causing a problem when I was attempting to filter our reads longer than a certain length in the indexer. Basically in the if check I was adding `&& read length < some_num)`, which caused a segfault and other issues later.

I uncommented the test, and it works except I had to change one thing. In the `past_the_last_read_id` argument we were previously passing, but since the read is now filtered the internal vectors in indexer only contain 1 read. this was causing a seg fault in the test itself. So I had to change that value to 1. Let me know if it shouldn't be that way.

mimaric:
> I uncommented the test, and it works except I had to change one thing. In the `past_the_last_read_id` argument we were previously passing, but since the read is now filtered the internal vectors in indexer only contain 1 read. this was causing a seg fault in the test itself. So I had to change that value to 1. Let me know if it shouldn't be that way.

That's something we should discuss, that part of the API is not the best. I think `past_the_last_read_id` should still be 2, i.e. the number of reads we want to process. We don't know up front how many reads are going to be skipped. The reason I added that skipping is that Minimizer does not handle the empty reads properly.
We should sit down and think about read_id <-> actual ordinal number of that read in the input file mapping. This gets exceptionally tricky once we start chunking actual reads into smaller ones in order to prevent OOMs, as you're doing right now

tijyojwad:
> I think past_the_last_read_id should still be 2, i.e. the number of reads we want to proces

Yeah, makes sense. It's this check that needs to be rewritten in the test function though
```
for (read_id_t read_id = first_read_id; read_id < past_the_last_read_id; ++read_id)
    {
        ASSERT_EQ(index.read_id_to_read_length(read_id), expected_read_id_to_read_length[read_id - first_read_id]) << "read_id: " << read_id;
        ASSERT_EQ(index.read_id_to_read_name(read_id), expected_read_id_to_read_name[read_id - first_read_id]) << "read_id: " << read_id;
    }
```
because after skipping/filtering of reads, indexing by read id directly doesn't make sense because it assumes every read id was processed by the indexer.

> once we start chunking actual reads into smaller ones in order to prevent OOMs, as you're doing right now

Absolutely right. My current experiments just filtered reads out altogether, but it definitely holds true for chunking within single reads.

tijyojwad:
Closing for now as this isn't the proper fix

PR #269, opened 2019-12-31
Additional flag to skip native optimized during cudapoa compilation

rilango:
This will allow building cudapoa for machines with older CPUs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -18,7 +18,7 @@ option(cga_enable_tests "Build ClaraGenomicsAnalysis unit tests" OFF)
 option(cga_enable_benchmarks "Build ClaraGenomicsAnalysis benchmarks" OFF)
 option(cga_build_shared "Build ClaraGenomicsAnalysis libraries as shared objects" OFF)
 option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHECK_ERR calls" OFF)
-option(optimize_for_native "Build with march=native" ON)
+option(cga_optimize_for_native_cpu "Build with march=native" ON)
 # The spoa_accurate option runs a different (and slower) version of
 # the topological sort in cudapoa which exactly matches the output
 # of the topological sort implementation in the SPOA library. This is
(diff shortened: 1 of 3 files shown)


cudapoa/CMakeLists.txt
```diff
@@ -26,8 +26,13 @@ cuda_add_library(cudapoa ${cga_library_type}
     )
 
 target_link_libraries(cudapoa utils logging)
+target_compile_options(cudapoa PRIVATE -Werror -Wall -Wextra)
 
-target_compile_options(cudapoa PRIVATE -Werror -Wall -Wextra -march=native)
+option(optimize_for_native "Build with march=native" ON)
+if (optimize_for_native)
+    target_compile_options(cudapoa PRIVATE -march=native)
+    # set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -march=native")
```
tijyojwad: let's remove the commented line

cudapoa/CMakeLists.txt
```diff
@@ -26,8 +26,12 @@ cuda_add_library(cudapoa ${cga_library_type}
     )
 
 target_link_libraries(cudapoa utils logging)
+target_compile_options(cudapoa PRIVATE -Werror -Wall -Wextra)
 
-target_compile_options(cudapoa PRIVATE -Werror -Wall -Wextra -march=native)
+option(optimize_for_native "Build with march=native" ON)
```
tijyojwad: I think it would be better to move this option up to the top level CMakeLists.txt, and call it `cga_optimize_for_native_cpu` (following convention here https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/CMakeLists.txt#L17 ). Can you also see if other parts of the system use `march=native` and if so guard them with this option as well?

CMakeLists.txt
```diff
@@ -18,6 +18,7 @@ option(cga_enable_tests "Build ClaraGenomicsAnalysis unit tests" OFF)
 option(cga_enable_benchmarks "Build ClaraGenomicsAnalysis benchmarks" OFF)
 option(cga_build_shared "Build ClaraGenomicsAnalysis libraries as shared objects" OFF)
 option(cga_device_synchronize_kernels "Run cudaDeviceSynchronize() in CGA_CU_CHECK_ERR calls" OFF)
+option(optimize_for_native "Build with march=native" ON)
```
tijyojwad: rename to `cga_optimize_for_native_cpu`

PR #270, opened 2020-01-03
[ci] add python doc generation check

tijyojwad:
Use sphinx to generate documentation for python code

### DIFF
--- a/ci/checks/style.sh
+++ b/ci/checks/style.sh
@@ -47,11 +47,8 @@ logger "Run Python documentation generation..."
 mkdir -p pyclaragenomics/docs; cd pyclaragenomics/docs
 sphinx-quickstart -p "Clara Genomics Analysis SDK" -v "x.y.z" -a "NVIDIA Corportation" -q --ext-autodoc --sep
 ##### Update sphinx conf.py with path for modules
-echo "import os" >> temp.py
-echo "import sys" >> temp.py
-echo "sys.path.insert(0, os.path.abspath('../..'))" >> temp.py
-cat source/conf.py >> temp.py
-mv temp.py source/conf.py
+sed -i '1s@^@import sys \n@' source/conf.py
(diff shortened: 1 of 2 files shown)


ci/checks/style.sh
```diff
@@ -38,16 +38,31 @@ source ci/common/prep-init-env.sh ${WORKSPACE}
...
 ./ci/checks/check_copyright.py
 
+# Python
 logger "Run Python/Cython formatting check..."
 python -m pip install -r ./ci/checks/python-style-requirements.txt
 source pyclaragenomics/style_check
 
+logger "Run Python documentation generation..."
+mkdir -p pyclaragenomics/docs; cd pyclaragenomics/docs
+sphinx-quickstart -p "Clara Genomics Analysis SDK" -v "x.y.z" -a "NVIDIA Corportation" -q --ext-autodoc --sep
```
ohadmo: We should extract the version from setup.py for now (until we create a setup.cfg file).

ci/checks/style.sh
```diff
@@ -38,16 +38,31 @@ source ci/common/prep-init-env.sh ${WORKSPACE}
...
+##### Update sphinx conf.py with path for modules
+echo "import os" >> temp.py
```
ohadmo: Maybe we should use `sed` in order to append commands to the beginning of the `conf.py` file instead of creating a temp file?  What do you think?
https://stackoverflow.com/a/9533736

tijyojwad: Neat! Updated

ci/checks/style.sh
tijyojwad: I specified a dummy version in this line mainly because this doesn't create the official documentation. It's just a test to ensure that documentation generation runs successfully. Working on a separate release script that will do proper version reading and so on.

Issue #265, opened 2019-12-19
[cudamapper] Optimize sorting in overlapper


mimaric:
cudamapper spends most of its execution time doing sorting at the beginning of the `Overlapper`. This is because `Matcher` groups anchors by representation and `Overlapper` needs them to be grouped by pairs of query and target read_ids

Think of a way to avoid this sort and harmonize the interface between `Matcher` and `Overlapper`.

Alternatively look for ways to reduce the sorting time.

PR #279, opened 2020-01-20
[cudamapper] Get the length of the longest read

mimaric:
Implemented `Index::number_of_basepairs_in_longest_read()` which returns the length of the longest read in that index.
Also switched return type of `Index::number_of_reads` from `std::uint64_t` to `read_id_t`.

These methods will be needed in a future PR for Anchor sorting

See #265

### DIFF
--- a/cudamapper/src/index_gpu.cuh
+++ b/cudamapper/src/index_gpu.cuh
@@ -10,6 +10,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <vector>
 
 #include <thrust/adjacent_difference.h>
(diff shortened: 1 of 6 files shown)


mimaric:
rerun tests

cudamapper/src/index_gpu.cuh
```diff
@@ -616,10 +631,12 @@ void IndexGPU<SketchElementImpl>::generate_index(const io::FastaParser& parser,
         const std::string& read_name      = fasta_reads.back().name;
         if (read_basepairs.length() >= window_size_ + kmer_size_ - 1)
         {
+            // TODO: make sure that no read is longer than what fits into position_in_read_t
```
vellamike: no DNA is this long, but still good to check I guess

PR #283, opened 2020-01-28
[cudaaligner] Changes for ukkonen_cpu

ahehn-nv:
* Moved implementation to a .cpp file
* Replaced old implementation by a naive fool-proof implementation for testing


tijyojwad:
@ahehn-nv I see the declaration and implementation, but this PR doesn't have any change where the `naive` version is being used/tested. is that in a separate PR?

ahehn-nv:
Yes, I might add a few sanity tests in a separate PR.
I added this version mostly to verify more complicated banding implementations in tests. This version is the most straight forward implementation of the Ukkonen's banding idea that I could think of - it also still uses the full NW matrix.
Might make sense to move the whole thing from `cudaaligner/src` to `cudaaligner/tests` instead.

PR #288, opened 2020-02-05
[cudamapper] Index provides helper functions for smallest and largest read_id in it

mimaric:
These functions will be needed for new anchors sorting


vellamike:
Approved (on phone) pls merge

PR #294, opened 2020-02-18
[cudamapper] No temporary representations in Index

mimaric:
Due to the way old `device_buffer` was implemented a temporary `representation_d` was needed. This is not the case anymore.

Old `device_buffer` also didn't have `begin()` and `end()`, new one has it. Adapted to these changes.

Added a few comments on how to improve efficiency and code design even more, not a bottleneck right now so leaving it for later


cudamapper/src/index_gpu.cuh
```diff
@@ -747,19 +750,11 @@ void IndexGPU<SketchElementImpl>::generate_index(const io::FastaParser& parser,
...
-
-    // copy the data to member functions (depending on the interface desing these copies might not be needed)
-    representations_d_.resize(representations_d.size());
-    representations_d_.shrink_to_fit();
-    thrust::copy(thrust::device,
-                 representations_d.data(),
-                 representations_d.data() + representations_d.size(),
-                 representations_d_.begin());
-    representations_d.free();
+                               std::begin(representations_d_),
```
ahehn-nv: Does it make sense to sort `sketch_elements.representations_d` + `sketch_elements.rest_d` and then move them to `representations_d_`? That way the object would remain in a valid state when an exception is thrown. It might have already destroyed `sketch_elements.representations_d` though...

mimaric: Yes, I'll implement it. But the code is a exception-safety minefield anyway...

mimaric: done

ahehn-nv: yes, I know. :boom: So I am unsure if I should care about such things.

PR #295, opened 2020-02-18
Host cache for indexer

r-mafi:
The current PR contains changes to implement host caching for indexer.
it can be controlled with -C parameters, e.g.
./cudamapper chrx300k.fasta chrx300k.fasta -F 1e-5 -i 1000 -C 1000
Default -C parameter is ~~1000~~ 0 indices. So by default host cache is not used, unless defined as input param by user


GPUtester:

mimaric:
rerun tests

cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -72,6 +72,16 @@ class Index
     /// \return number of reads in input data
     virtual read_id_t number_of_reads() const = 0;
 
+    /// ToDo this is to allow IndexCache to access to read_id_to_read_name_, may need to change if overloading is not desired
```
mimaric: I would suggest naming it `read_ids_to_read_names()`, not just to avoid overloading but because this function returns something that maps multiple read_ids to multiples read_names.

cudamapper/src/index.cu
```diff
@@ -40,3 +40,135 @@ std::unique_ptr<Index> Index::create_index(std::shared_ptr<DeviceAllocator> allo
...
+            : first_read_id(first_read_id_in),
+              kmer_size(kmer_size_in),
+              window_size(window_size_in)
+    {
+        CGA_NVTX_RANGE(profiler, "cache_index");
+        {
+            auto const & src = index.representations();
+            auto const sz = src.size();
+            representations_.resize(sz);
+            cudautils::device_copy_n(src.data(), sz, representations_.data());
```
mimaric: This is a personal preference, I find
```
representations_.resize(index.representations().size());
cudautils::device_copy_n(index.representations().data(), index.representations().size(), representations_.data());
```
to be more readable, especially as in that case there is no need for additional scopes (but again it's a personal preference)

cudamapper/src/index.cu
```diff
@@ -40,3 +40,135 @@ std::unique_ptr<Index> Index::create_index(std::shared_ptr<DeviceAllocator> allo
...
+
+
+
+namespace claragenomics
+{
+namespace cudamapper
+{
+
+    IndexCache::IndexCache(const Index& index,
+                           const read_id_t first_read_id_in,
```
mimaric: Personal preference: We generally do not use `_in` to indicate parameter direction, `const` should make it clear enough

cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -117,3 +127,99 @@ class Index
...
+
+    std::vector<std::string> read_id_to_read_name_;
+    std::vector<std::uint32_t> read_id_to_read_length_;
+
+    read_id_t number_of_reads_;
+    position_in_read_t number_of_basepairs_in_longest_read_;
+
+public:
+
+    const read_id_t first_read_id = 0;
```
mimaric: Do we really want these to be `public`?

mimaric:
@r-mafi FYI: Before submitting a PR you can execute `make docs` and `make check-format` to make sure PR won't fail for trivial things. You can also execute `make format` to format automatically. I usually either format manually or `git add` files and then execute `make format` to make sure I know what was changed by it.

r-mafi:
> @r-mafi FYI: Before submitting a PR you can execute `make docs` and `make check-format` to make sure PR won't fail for trivial things. You can also execute `make format` to format automatically. I usually either format manually or `git add` files and then execute `make format` to make sure I know what was changed by it.

Thanks, that was a useful tip Milos.

r-mafi:
> @r-mafi FYI: Before submitting a PR you can execute `make docs` and `make check-format` to make sure PR won't fail for trivial things. You can also execute `make format` to format automatically. I usually either format manually or `git add` files and then execute `make format` to make sure I know what was changed by it.

Thanks, that was a useful tip Milos.

r-mafi:
> I'm not sure if `IndexCache` should really be a completely separate class or it should inherit from `Index`. @r-mafi @ahehn-nv let's discuss this later today

I was initially thinking to inherit it from Index. But soon noticed all those methods have a different return type in IndexCache than in Index (std::vector<> instead of device_buffer<>). So basically overriding becomes irrelevant.

cudamapper/include/claragenomics/cudamapper/index.hpp
r-mafi: changed to private.

cudamapper/include/claragenomics/cudamapper/index.hpp
r-mafi: done.

cudamapper/src/index.cu
r-mafi: fixed.

cudamapper/src/index.cu
r-mafi: the reason is I keep using the same name auto const& src to represent different data types, therefore I used extra scope.

r-mafi:
rerun tests

mimaric:
> > I'm not sure if `IndexCache` should really be a completely separate class or it should inherit from `Index`. @r-mafi @ahehn-nv let's discuss this later today
> 
> I was initially thinking to inherit it from Index. But soon noticed all those methods have a different return type in IndexCache than in Index (std::vector<> instead of device_buffer<>). So basically overriding becomes irrelevant.

I partially agree. Inheriting from Index would ensure long-term compatibility of the two. I was also thinking of having `IndexGPU(const IndexCache&)` and `IndexCache(const IndexGPU&)` copy constructors (having something like does not require inheritance though).

Anyway, if you decide to keep it as it is (which probably makes sense) make it compatible with general design where we have an abstract class in public header and a `generate_xyz` function which creates an object of implementing class which is declared and defined in `src`

tijyojwad:
add to whitelist

cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -72,6 +72,14 @@ class Index
     /// \return number of reads in input data
     virtual read_id_t number_of_reads() const = 0;
 
+    /// \brief returns look up table array mapping read id to read name
+    /// \return the array mapping read id to read name
+    virtual const std::vector<std::string>& read_id_to_read_names() const = 0;
```
mimaric: Really nitpicky: shouldn't this be `read_ids_to_read_names`?

cudamapper/src/index_gpu.cuh
```diff
@@ -563,6 +577,57 @@ IndexGPU<SketchElementImpl>::IndexGPU(std::shared_ptr<DeviceAllocator> allocator
...
+    , directions_of_reads_d_(allocator)
+    , unique_representations_d_(allocator)
+    , first_occurrence_of_representations_d_(allocator)
+{
+    number_of_reads_                     = host_cache.number_of_reads();                     //H2D
+    number_of_basepairs_in_longest_read_ = host_cache.number_of_basepairs_in_longest_read(); //H2D
+
+    //H2D- representations_d_ = host_cache.representations();
+    representations_d_.resize(host_cache.representations().size());
+    representations_d_.shrink_to_fit();
```
mimaric: FYI: Currently `device_buffer::resize()` allocates exactly the needed amount of memory: https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/b5ea4f4d9d11ccf2bd8e8b4a571168ba6a996547/common/utils/include/claragenomics/utils/buffer.hpp#L148
`device_buffer::shrink_to_fit()` performs the check and only shrinks if need, so in this case it does not have any effect: https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/b5ea4f4d9d11ccf2bd8e8b4a571168ba6a996547/common/utils/include/claragenomics/utils/buffer.hpp#L204

I do agree this is a good practice as it future proofs for the case that `device_buffer::resize()` changes it's behavior and the price of calling `device_buffer::shrink_to_fit()` is negligible in this case

(so all good, just discussing)

cudamapper/src/index_gpu.cuh
```diff
@@ -563,6 +577,57 @@ IndexGPU<SketchElementImpl>::IndexGPU(std::shared_ptr<DeviceAllocator> allocator
...
+    , window_size_(host_cache.window_size())
+    , allocator_(allocator)
+    , representations_d_(allocator)
+    , read_ids_d_(allocator)
+    , positions_in_reads_d_(allocator)
+    , directions_of_reads_d_(allocator)
+    , unique_representations_d_(allocator)
+    , first_occurrence_of_representations_d_(allocator)
+{
+    number_of_reads_                     = host_cache.number_of_reads();                     //H2D
```
mimaric: Is this really H2D? these values are on host in both cases

cudamapper/src/index_gpu.cuh
```diff
@@ -563,6 +577,57 @@ IndexGPU<SketchElementImpl>::IndexGPU(std::shared_ptr<DeviceAllocator> allocator
                    filtering_parameter);
 }
 
+template <typename SketchElementImpl>
+IndexGPU<SketchElementImpl>::IndexGPU(std::shared_ptr<DeviceAllocator> allocator,
```
mimaric: Nitpicky: I'm not sure, but I think we generally put allocator as the last parameter

cudamapper/src/main.cu
```diff
@@ -177,38 +182,54 @@ int main(int argc, char* argv[])
...
-                                                          const claragenomics::cudamapper::read_id_t end_index,
-                                                          const std::uint64_t k,
-                                                          const std::uint64_t w,
-                                                          const int device_id,
-                                                          const bool allow_cache_index) {
+    std::vector<std::map<std::pair<uint64_t, uint64_t>, std::shared_ptr<claragenomics::cudamapper::Index>>> device_index_cache(num_devices);
+
+    auto get_index = [&device_index_cache, &host_index_cache, max_index_cache_size_on_device, max_index_cache_size_on_host](std::shared_ptr<claragenomics::DeviceAllocator> allocator,
+                                                                                                                            claragenomics::io::FastaParser& parser,
+                                                                                                                            const claragenomics::cudamapper::read_id_t start_index,
```
mimaric: What happened to formatting here? (might be something silly `make format` did)

cudamapper/src/main.cu
```diff
@@ -177,38 +182,54 @@ int main(int argc, char* argv[])
...
             // using the evict_index function.
-            if (get_size<int32_t>(index_cache[device_id]) < max_index_cache_size && allow_cache_index)
+            if (get_size<int32_t>(device_index_cache[device_id]) < max_index_cache_size_on_device && allow_cache_index)
             {
-                index_cache[device_id][key] = index;
+                device_index_cache[device_id][key] = index;
+            }
+
+            // update host cache
+            // ToDo: make it thread-safe when there are multiple GPUs
```
mimaric: Can you do this before the merge? We are trying to keep the code as multi-GPU-ready as possile

cudamapper/src/main.cu
```diff
@@ -226,13 +247,17 @@ int main(int argc, char* argv[])
...
+    auto evict_index = [&device_index_cache, &host_index_cache](const claragenomics::cudamapper::read_id_t query_start_index,
+                                                                const claragenomics::cudamapper::read_id_t query_end_index,
+                                                                const int device_id,
+                                                                const int num_devices) {
         std::pair<uint64_t, uint64_t> key;
         key.first  = query_start_index;
         key.second = query_end_index;
-        index_cache[device_id].erase(key);
+        device_index_cache[device_id].erase(key);
+        // ToDo: freeing host memory is limited to 1 GPU, need to change to work with multiple GPUs
```
mimaric: Same as above (if not too much work please enable multi-GPU)

cudamapper/src/index.cu
mimaric: I agree, but doing it in two lines like I suggested does not require you to explicitly specify the type and if anyone is unsure of what's input and what's output they can check the signature (like with any other function)

cudamapper/src/main.cu
r-mafi: yes, make format pushed everything to the right. will change and see if check-format complains about it.

cudamapper/src/index_gpu.cuh
r-mafi: actually I found everywhere it's passed as the first parameter, wrote it the same way to be consistent with the rest of the code.

cudamapper/src/index_gpu.cuh
r-mafi: that's right, fixed.

cudamapper/src/index_gpu.cuh
r-mafi: thanks for clarifying, I was not sure to add shrink_to_fit(), asked @akkamesh about it, he also suggested no harm to keep it.

cudamapper/include/claragenomics/cudamapper/index.hpp
r-mafi: I had factored last 's' for the whole string! :D will change.

cudamapper/src/index.cu
r-mafi: this was a minor case of conflict between personal preferences. I changed to your suggested way. The reason I liked the other way was the code looked more concise and easier to read.

cudamapper/src/index_gpu.cuh
mimaric: Oh, right, sorry, allocators are a new addition, still didn't get used to them

cudamapper/src/main.cu
mimaric: Ok. There are a few cases where `make format` does strange things, especially with lambdas. If it insists on having certain formatting just let it change whatever is needed

cudamapper/src/index.cu
```diff
@@ -40,3 +40,123 @@ std::unique_ptr<Index> Index::create_index(std::shared_ptr<DeviceAllocator> allo
...
+            thrust::copy(src.begin(), src.end(), read_id_to_read_length_.begin());  //H2H, may replace with shared_ptr
+        }
+    }
+
+    std::unique_ptr<Index> IndexCache::copy_index_to_device(std::shared_ptr<claragenomics::DeviceAllocator> allocator)
+    {
+        return std::make_unique<IndexGPU<Minimizer>>(allocator, *this);
+    }
+
+    const std::vector<representation_t>& IndexCache::representations() const
```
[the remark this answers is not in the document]
mimaric: @ahehn-nv My comment about this was that this class should be abstract, if we want to stick to abstract public classes policy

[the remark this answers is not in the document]
ahehn-nv: oh, yes... sorry, makes sense in this case.

cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -117,3 +127,88 @@ class Index
 } // namespace cudamapper
 
 } // namespace claragenomics
+
+
+
+namespace claragenomics
+{
+namespace cudamapper
```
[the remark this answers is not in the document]
r-mafi: At the time I was not sure if we want to keep it in index.hpp. Will make it abstract and place in the existing namespace block.

cudamapper/src/index.cu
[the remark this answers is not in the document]
r-mafi: revised, now it's rewritten as abstract class.

r-mafi:
> @r-mafi FYI: Before submitting a PR you can execute `make docs` and `make check-format` to make sure PR won't fail for trivial things. You can also execute `make format` to format automatically. I usually either format manually or `git add` files and then execute `make format` to make sure I know what was changed by it.

r-mafi:
> Anyway, if you decide to keep it as it is (which probably makes sense) make it compatible with general design where we have an abstract class in public header and a `generate_xyz` function which creates an object of implementing class which is declared and defined in `src`

modified as per your suggestion.

cudamapper/src/main.cu
r-mafi: the simplest way to keep it thread-safe was to limit write-access to only device 0. Alternatively could be done with mutex, but I doubt they make any significant difference performance-wise.

cudamapper/src/main.cu
r-mafi: the RAM eventually will be freed at the end, but this line frees up memory gradually as computation progresses forward. Since freeing-up memory depends on access pattern by different GPUs (preferrably we don't want to earse a key where may be accessed by another GPU sometime later), I kept it as it is and added this comment:
`// host memory can be freed by removing (key) when working with 1 GPU`
`// in multiple GPUs we keep (key), as it may be accessed by other GPUs depending on access pattern`

r-mafi:
> As mentioned in a previous comment, I agree that `IndexCache` should not inherit from `Index` but I suggest using the abstract class -> implementation class structure.

Done
> It might just be me, but to me `IndexCache` sound like a class that manages cached indices. We should think of a better name here

I was not sure about an attentive. Please share of you have any suggestions.

cudamapper/src/main.cu
r-mafi: Yes, tried again, but check-format didn't like it and make format address those spaces again.

mimaric:
> > It might just be me, but to me `IndexCache` sound like a class that manages cached indices. We should think of a better name here
> 
> I was not sure about an attentive. Please share of you have any suggestions.

How about `IndexHostCopy`? That way it is clear that it is `Index`-like, that it is saved on host and that it is a copy, i.e. it cannot create a new index, it's just used for storage

mimaric:
Regarding multi-GPU, ok, let's leave it as it is for now. We should anyway extract index caching into a separate class once we make sure everything works as expected and we can take care of multi-GPU then

r-mafi:
> Considering renaming `IndexCache` to `IndexHostCopy`, other than that looks good to me.

Changed the name, will merge after tests are done. Thanks for the helpful comments.

PR #297, opened 2020-02-24
[cudamapper] Overlapper gaps

r-mafi:
- detached thread to execute filter_and_print_overlaps is launched more efficiently, as overlaps_to_add vector is passed to std::thread as a shared pointer.

- updating read names for overlap vector is moved outside of get_overlaps and now is performed on detached thread.

- filter_overlaps which takes place on CPU is slightly modified to work with overlap read_ids rather than read_names. This allows to move the omp for loop used for updating read names work on only filtered set of overlaps.

### DIFF
--- a/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
@@ -32,13 +32,9 @@ public:
 
     /// \brief returns overlaps for a set of reads
     /// \param overlaps Output vector into which generated overlaps will be placed
-    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    /// \param index_query
-    /// \param index_target
-    virtual void get_overlaps(std::vector<Overlap>& overlaps,
-                              device_buffer<Anchor>& anchors,
-                              const Index& index_query,
-                              const Index& index_target) = 0;
+    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read
+    /// (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
+    virtual void get_overlaps(std::vector<Overlap> &overlaps, device_buffer <Anchor> &anchors) = 0;
 
     /// \brief prints overlaps to stdout in <a href="https://github.com/lh3/miniasm/blob/master/PAF.md">PAF format</a>
     static void print_paf(const std::vector<Overlap>& overlaps);
--- a/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
@@ -32,10 +32,9 @@ public:
     virtual ~Overlapper() = default;
 
     /// \brief returns overlaps for a set of reads
-    /// \param overlaps Output vector into which generated overlaps will be placed
-    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read
-    /// (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    virtual void get_overlaps(std::vector<Overlap>& overlaps, device_buffer<Anchor>& anchors) = 0;
+    /// \param fused_overlaps Output vector into which generated overlaps will be placed
+    /// \param d_anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
+    virtual void get_overlaps(std::vector<Overlap>& fused_overlaps, device_buffer<Anchor>& d_anchors) = 0;
 
     /// \brief prints overlaps to stdout in <a href="https://github.com/lh3/miniasm/blob/master/PAF.md">PAF format</a>
     static void print_paf(const std::vector<Overlap>& overlaps);
--- a/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
@@ -52,6 +52,14 @@ public:
                                 const std::vector<Overlap>& overlaps,
                                 size_t min_residues    = 20,
                                 size_t min_overlap_len = 50);
+
+    /// \brief updates read names for vector of overlaps output from get_overlaps
+    /// \param overlaps input vector of overlaps generated in get_overlaps
+    /// \param index_query
+    /// \param index_target
+    static void update_read_names(std::vector<Overlap>& overlaps,
+                                  const Index& index_query,
+                                  const Index& index_target);
 };
 //}
 } // namespace cudamapper
--- a/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/overlapper.hpp
@@ -34,7 +34,7 @@ public:
     /// \param overlaps Output vector into which generated overlaps will be placed
     /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read
     /// (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    virtual void get_overlaps(std::vector<Overlap> &overlaps, device_buffer <Anchor> &anchors) = 0;
+    virtual void get_overlaps(std::vector<Overlap>& overlaps, device_buffer<Anchor>& anchors) = 0;
 
     /// \brief prints overlaps to stdout in <a href="https://github.com/lh3/miniasm/blob/master/PAF.md">PAF format</a>
     static void print_paf(const std::vector<Overlap>& overlaps);
--- a/cudamapper/src/overlapper.cpp
+++ b/cudamapper/src/overlapper.cpp
@@ -39,14 +39,14 @@ void Overlapper::update_read_names(std::vector<Overlap>& overlaps,
 #pragma omp parallel for
     for (size_t i = 0; i < overlaps.size(); i++)
     {
-        auto& o                      = overlaps[i];
-        std::string query_read_name  = index_query.read_id_to_read_name(o.query_read_id_);
-        std::string target_read_name = index_target.read_id_to_read_name(o.target_read_id_);
+        auto& o                             = overlaps[i];
+        const std::string& query_read_name  = index_query.read_id_to_read_name(o.query_read_id_);
+        const std::string& target_read_name = index_target.read_id_to_read_name(o.target_read_id_);
 
-        o.query_read_name_ = new char[query_read_name.length()];
+        o.query_read_name_ = new char[query_read_name.length() + 1];
         strcpy(o.query_read_name_, query_read_name.c_str());
 
-        o.target_read_name_ = new char[target_read_name.length()];
+        o.target_read_name_ = new char[target_read_name.length() + 1];
         strcpy(o.target_read_name_, target_read_name.c_str());
 
         o.query_length_  = index_query.read_id_to_read_length(o.query_read_id_);
--- a/cudamapper/src/overlapper.cpp
+++ b/cudamapper/src/overlapper.cpp
@@ -32,6 +32,28 @@ void Overlapper::filter_overlaps(std::vector<Overlap>& filtered_overlaps, const 
                  valid_overlap);
 }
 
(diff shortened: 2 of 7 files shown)


mimaric:
@r-mafi Here's how to turn on test compilation so you can check them before submitting: https://github.com/clara-genomics/ClaraGenomicsAnalysis#enable-unit-tests

r-mafi:
> @r-mafi Here's how to turn on test compilation so you can check them before submitting: https://github.com/clara-genomics/ClaraGenomicsAnalysis#enable-unit-tests

@mimaric thanks, looking at details of the error figured I've missed changing few `get_overlaps` in a test file. I should not have fully relied on CLion automatic refactoring which was used to change signature of `get_overlaps`.

r-mafi:
This PR may need to be changed based on PR #298 . Since this PR makes some modifications based on assumption where filtering is on the host.

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
```diff
@@ -188,7 +188,7 @@ TEST(TestCudamapperOverlapperTriggerred, OneAchorNoOverlaps)
     cudautils::device_copy_n(anchors.data(), anchors.size(), anchors_d.data()); //H2D
 
     std::vector<Overlap> overlaps;
-    overlapper.get_overlaps(overlaps, anchors_d, test_index, test_index);
```
mimaric: Can you add tests for `Overlapper::update_read_names`?

cudamapper/src/overlapper_triggered.hpp
```diff
@@ -33,12 +33,10 @@ class OverlapperTriggered : public Overlapper
     /// Uses a dynamic programming approach where an overlap is "triggered" when a run of
     /// Anchors (e.g 3) with a score above a threshold is encountered and untriggerred
     /// when a single anchor with a threshold below the value is encountered.
-    /// \param overlaps Output vector into which generated overlaps will be placed
-    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    /// \param index_query
-    /// \param index_target
+    /// \param fused_overlaps Output vector into which generated overlaps will be placed
```
mimaric: Maybe specify that names will not be updated and that `Overlapper::update_read_names()` has to be called? Not sure if this should be specified in `param` or general functions description

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -32,13 +32,9 @@ class Overlapper
...
-    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    /// \param index_query
-    /// \param index_target
-    virtual void get_overlaps(std::vector<Overlap>& overlaps,
-                              device_buffer<Anchor>& anchors,
-                              const Index& index_query,
-                              const Index& index_target) = 0;
+    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read
+    /// (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
+    virtual void get_overlaps(std::vector<Overlap>& overlaps, device_buffer<Anchor>& anchors) = 0;
```
mimaric: In `overlapper_trigered` `anchors` was renamed to `d_anchors` and `overlaps` to `fused_overlaps`. I'd rename them here as well for consistency

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -32,13 +32,9 @@ class Overlapper
...
     /// \brief returns overlaps for a set of reads
     /// \param overlaps Output vector into which generated overlaps will be placed
-    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
-    /// \param index_query
-    /// \param index_target
-    virtual void get_overlaps(std::vector<Overlap>& overlaps,
-                              device_buffer<Anchor>& anchors,
-                              const Index& index_query,
-                              const Index& index_target) = 0;
+    /// \param anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read
```
mimaric: Please keep comments consistent with the ones in `overlapper_trigered`

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -52,6 +48,14 @@ class Overlapper
                                 const std::vector<Overlap>& overlaps,
                                 size_t min_residues    = 20,
                                 size_t min_overlap_len = 50);
+
+    /// \brief updates read names for vector of overlaps output from get_overlaps
+    /// \param overlaps input vector of overlaps generated in get_overlaps
+    /// \param index_query
+    /// \param index_target
+    static void update_read_names(std::vector<Overlap>& overlaps,
```
mimaric: Is this function the same for all possible `overlapper` implementations? I think it is, but want to make sure I understand it correctly (otherwise implementation should go to `OverlapperTriggered`)

cudamapper/src/overlapper.cpp
```diff
@@ -32,6 +32,28 @@ void Overlapper::filter_overlaps(std::vector<Overlap>& filtered_overlaps, const
...
+{
+#pragma omp parallel for
+    for (size_t i = 0; i < overlaps.size(); i++)
+    {
+        auto& o                      = overlaps[i];
+        std::string query_read_name  = index_query.read_id_to_read_name(o.query_read_id_);
+        std::string target_read_name = index_target.read_id_to_read_name(o.target_read_id_);
+
+        o.query_read_name_ = new char[query_read_name.length()];
+        strcpy(o.query_read_name_, query_read_name.c_str());
```
mimaric: Is copy to `query_read_name` actually necessary? Can't we simply do `strcpy(o.query_read_name_, index_query.read_id_to_read_name(o.query_read_id_).c_str());`
(make have `const std::string& query_read_name` if we want a separate variable for readability reasons)
@vellamike for this question

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -52,6 +48,14 @@ class Overlapper
...
+                                  const Index& index_query,
```
mimaric: I have a more general and a bit controversial suggestion. `update_read_names()` actually only requires `read_id_to_read_name()` and `read_id_to_read_length()`, which are both host arrays. As `update_read_names()` is supposed to run in parallel with another anchor generating iteration this means that these indices would take up (scarce) device memory.
My suggestion: only pass `read_id_to_read_name` and `read_id_to_read_length` vectors for both indices. That way the user (author of `main.cu`) can decide if they want to pay the price of making copies of those vectors in order to save some device space or they are fine with keeping the whole `Index` object alive (which might make sense in device index caching scenario).
@vellamike for opinion

cudamapper/src/overlapper.cpp
vellamike: yes, this is just the way it is for readability.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
vellamike: @mimaric I see two sides to this:

1. It is cleaner how it is at the moment, the `Index` will only in practice live for a very short extra time anyway
2. In principle, this not only wastes device memory but also can delay eviction from device cache, leading to confusing behaviour. 

My gut instinct is that the extra H2H copy will be expensive and the code a little uglier so we shouldn't do your suggestion. What do you think?

cudamapper/src/overlapper.cpp
ahehn-nv: Isn't this a write beyond bounds?
`query_read_name.length()` returns the number of chars in `query_read_name` without the null terminator, right?
`query_read_name.c_str()` returns a pointer to a c-string with a null terminator and `strcpy` copies the string until, but including, the null terminator. So it will write `query_read_name.length()+1` characters.

We should really get rid of handling raw arrays via pointers in the long run.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
mimaric: I agree. Let's keep it as it is for now and do some more analysis afterwards. That's why I said it's a bit controversial.

cudamapper/src/overlapper.cpp
mimaric: Hmm, yes, that looks bad.

Regarding raw pointers vs string, I fully agree. The problem is that Thrust had problems working with structs that have string in them so that's why we decided to stick with raw pointers.
Mike, Kamesh and I all had a quick look at it and couldn't find a solution, but we should invest some more time in it.

ahehn-nv: If we don't require null-terminated c-strings, we could just not use `strcpy` here and do `std::copy_n(query_read_name.data(), query_read_name.size(), o.query_read_name_);` instead.
Note that writing strings, e.g. to stdout, might require them to be null terminated though.

mimaric: Apparently we use `printf` for printing out the results (https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/bb77507ec301e03b48539d336bcdc6eb4ce9c078/cudamapper/src/overlapper.cpp#L40), but that still requires the null-terminated string

mimaric: I just tried switching to `const std::string& query_read_name`, but there was no performance difference. I assume compiler was able to avoid that copy to `std::string query_read_name` as `index_query` is `const`.
I'd say we should anyway switch to `const std::string& query_read_name` and not rely on compiler to do it

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
vellamike: +1

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
r-mafi: done.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
r-mafi: done.

cudamapper/src/overlapper.cpp
r-mafi: changed it to `const std::string&` and also to be safe, increased the size by one char `new char[query_read_name.length()+1]`.
Also should note in addition to possible compiler optimizations, there are 2 additional reasons in this PR not to see any performance difference by changing to  `const std::string&`: 1- it runs async 2- it runs on a rather small set of filtered overlaps, rather than looping on large arrays as in previous versions

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
r-mafi: I assumed it will be the same.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
r-mafi: left as it is.

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
r-mafi: added a couple of tests.

cudamapper/src/overlapper_triggered.hpp
r-mafi: added some description to both function description and in `param`

cudamapper/tests/Test_CudamapperOverlapperTriggered.cu
mimaric: Can you please add them to all places where `overlapper.get_overlaps(overlaps, anchors_d, test_index, test_index);` got replaced by `overlapper.get_overlaps(overlaps, anchors_d);`?

r-mafi: since update_read_names loops through overlaps, it made sense to add the test for it only to the cases where `overlaps.size() > 0`. Therefore you don't see everywhere.
Also, this update_read_names which before used to take place inside `overlapper.get_overlaps(overlaps, anchors_d, test_index, test_index)`, its output was never verifies, I mean before this PR we did not have any checks for updating query and target read names.

r-mafi:
> Conditionally approving. Please address the missing tests comment (shouldn't be a big one) and merge

since update_read_names loops through overlaps, it is added to the cases where `overlaps.size() > 0`. Before updating read names which took place withing get_overlaps was not verified at all, now it has been added to the tests, wherever applicable.

PR #298, opened 2020-02-24
[cudamapper] move filtering to GPU

tijyojwad:


cudamapper/src/overlapper_triggered.cu
```diff
@@ -371,14 +395,22 @@ void OverlapperTriggered::get_overlaps(std::vector<Overlap>& fused_overlaps,
                       d_fusedoverlaps_args.data() + n_fused_overlap,
                       d_fused_overlaps.data(), fuse_op);
 
-    // memcpyD2H - move fused overlaps to host
-    fused_overlaps.resize(n_fused_overlap);
-    cudautils::device_copy_n(d_fused_overlaps.data(), n_fused_overlap, fused_overlaps.data(), stream);
+    device_buffer<Overlap> d_filtered_overlaps(n_fused_overlap, _allocator, stream);
+
+    FilterOverlapOp filterOp(min_residues, min_overlap_len);
+    auto filtered_overlaps_end =
```
tijyojwad: @akkamesh @mimaric this line is seg faulting. I tried earlier to figure out what's messing it up, but couldn't figure it out yet. The addresses appear to be valid, and I can copy over individual elements in the d_fused_overlap vector into host properly. I wasn't sure if the pointer arithmetic was being done correctly, so I also tried `d_fused_overlaps.data() + n_fused_overlap * sizeof(Overlap)` for the end iterator argument, but that didn't work either.

I'll continue to debug, but maybe something obvious is jumping out at you?

P.S. - my previous implementation was using the thrust device vectors and that implementation was working alright. so I think logically this is setup okay, but some buffer pointer is not being handled correctly by me

akkamesh: @tijyojwad - You are missing the execution policy for thrust copy_if. Can you update the call to `thrust::copy_if(thrust_exec_policy, ...)` and see if you still see the seg fault?

cudamapper/src/overlapper_triggered.cu
```diff
@@ -133,6 +133,28 @@ struct FuseOverlapOp
     }
 };
 
+struct FilterOverlapOp
```
mimaric: Can this be a lambda?

cudamapper/src/overlapper.cpp
```diff
@@ -16,22 +16,6 @@ namespace claragenomics
 namespace cudamapper
 {
 
-void Overlapper::filter_overlaps(std::vector<Overlap>& filtered_overlaps, const std::vector<Overlap>& overlaps,
```
mimaric: @tijyojwad @vellamike
Should this function be in `overlapper.cpp` or `overlapper_triggered.cu`? I have a feeling it is not `overlapper_triggerd` specific

cudamapper/src/overlapper_triggered.cu
mimaric: Ok, I see this is a widely used pattern in `overlapper_triggered`, no need to take care of it in this PR

cudamapper/src/overlapper.cpp
tijyojwad: @mimaric agreed, it's not overlapper_triggered specific. I mostly followed the example of the FusedOp (since that didn't look overlapper_triggered specific either). Do you think it makes sense to move both of them to a common overlapper_ops.hpp file? (don't think it should really belong to the public overlapped.hpp either)

cudamapper/src/overlapper_triggered.cu
tijyojwad: Yep that fixed it! Thanks @akkamesh ! Just curious - why did it work without the execution policy when I was using thrust device vectors?

cudamapper/src/overlapper.cpp
mimaric: Ok, let's keep it as it is. Overlapper fill need a separate refactoring anyway

mimaric:
@tijyojwad Some unittests are failing. Can you please look into that?

cudamapper/src/overlapper_triggered.cu
akkamesh: > Yep that fixed it! Thanks @akkamesh ! Just curious - why did it work without the execution policy when I was using thrust device vectors?

The default execution policy gets picked based on the supplied iterators - calls host implementation for thrust::host_vectors and device implementation for thrust::device_vectors. On the other hand, when we pass raw pointers (defined outside of thrust) without explicit execution policy then by default thrust dispatches the call to host path of the algorithm (which is what was happening here, it was calling host path with device pointer and that results in segfault).

tijyojwad: Aha got it, thanks for the explanation @akkamesh !

tijyojwad:
@mimaric addressed failing tests

cudamapper/src/overlapper_triggered.cu
mimaric: @tijyojwad More on that here: https://github.com/thrust/thrust/wiki/Quick-Start-Guide#iterators-and-static-dispatching

PR #301, opened 2020-03-02
fixed memory leak

r-mafi:
updating read names was applied to filtered set of overlaps, but the call for freeing memory was not changed accordingly. fixed!


r-mafi:
this issues was addressed in PR #298 , as it won't be necessary any more, will close this PR.

PR #305, opened 2020-03-03
improving host cache memory usage

r-mafi:
host cache is used to store an index only if that index is not already cached on GPU. This should reduce RAM usage in presence of using GPU memory for caching.

### DIFF
--- a/cudamapper/src/main.cu
+++ b/cudamapper/src/main.cu
@@ -233,10 +233,9 @@ int main(int argc, char* argv[])
             {
                 device_index_cache[device_id][key] = index;
             }
-
-            // update host cache, only done by device 0 to avoid any race conditions in updating the host cache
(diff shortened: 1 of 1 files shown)


mimaric:
Are we sure that this is what we want?
Assuming host index in larger than device (which is the point) let's say we create an index and there is enough space for in on device. We cache it on device, use it for some time, evict it, need it again and then have to generate it from scratch instead of loading it from host.
Anyway, I think we will have to rethink Index cache in general if we want to overlap it with matcher+overlapper, so let's leave it as it is for now

r-mafi:
The main motive for this change was based on Mike's system that has low RAM but high-end GPU. In case of single GPU I don't think there is any significant impact on performance, since even considering what you described, re-generating a new index would be done only once as it will be cached again.

PR #310, opened 2020-03-04
[cudamapper] remove HTS parser, add minimum sequence length to kseqpp parser

vellamike:


cudamapper/src/main.cu
```diff
@@ -134,7 +134,7 @@ int main(int argc, char* argv[])
         std::cerr << "NOTE - Since query and target files are same, activating all_to_all mode. Query index size used for both files." << std::endl;
     }
 
-    std::shared_ptr<claragenomics::io::FastaParser> query_parser = claragenomics::io::create_kseq_fasta_parser(query_filepath);
+    std::shared_ptr<claragenomics::io::FastaParser> query_parser = claragenomics::io::create_kseq_fasta_parser(query_filepath, k + w);
```
mimaric: Shouldn't this be `k + w - 1`?
For `k = 5, w = 1` 5 basepairs are needed, for `k = 5, w = 2` 6 and so on

cudamapper/src/main.cu
```diff
@@ -144,7 +144,7 @@ int main(int argc, char* argv[])
     }
     else
     {
-        target_parser = claragenomics::io::create_kseq_fasta_parser(target_filepath);
+        target_parser = claragenomics::io::create_kseq_fasta_parser(target_filepath, k + w);
```
mimaric: also `k + w - 1`

common/io/src/kseqpp_fasta_parser.cpp
```diff
@@ -23,17 +23,21 @@ namespace claragenomics
...
     {
-        FastaSequence seq = {record.name, record.seq};
-        total_len += record.seq.size();
-        reads_.emplace_back(seq);
+        FastaSequence seq    = {record.name, record.seq};
+        auto sequence_length = record.seq.size();
+        if (sequence_length >= min_sequencece_length)
+        {
+            total_len += sequence_length;
+            reads_.emplace_back(seq);
```
ahehn-nv: Consider to replace `reads_.emplace_back(seq);` by `reads_.push_back(std::move(seq));` to save a copy from the `seq` object.

The idea of `emplace_back(list of arguments)` is to construct an object directly in the container, by forwarding the `list of arguments` to the constructor of the element class internally.
Since you pass the`seq` object (or more precisely an lvalue reference to it), the constructor used internally will be the copy constructor of FastaSequence. This means it is equivalent to `reads_.push_back(seq)`.
`emplace_back()` would be helpful if FastaSequence had a constructor `FastaSequence(std::string const& name, std::string const& seq)`. Then you could call `reads_.emplace_back(record.name, record.seq);` and avoid creating the `seq` object altogether.

With the temporary `seq` object, the only way to remove the copy into `reads_` is explicitly tell the compiler to move from the object in the call with either `push_back(std::move(seq));` or `emplace_back(std::move(seq));`. All STL containers have an overload `push_back(T&& x)`, so there is again no difference between `push_back` and `emplace_back` here.

vellamike: Thanks for this. I applied the change.

vellamike:
rerun tests

common/io/src/kseqpp_fasta_parser.cpp
```diff
@@ -23,17 +23,21 @@ namespace claragenomics
 namespace io
 {
 
-FastaParserKseqpp::FastaParserKseqpp(const std::string& fasta_file)
+FastaParserKseqpp::FastaParserKseqpp(const std::string& fasta_file, int min_sequencece_length)
 {
     klibpp::KSeq record;
     klibpp::SeqStreamIn iss(fasta_file.data());
     std::vector<FastaSequence> seqs;
     int total_len = 0;
```
mimaric: `total_len` is not actually being used anywhere, right?

common/io/src/fasta_parser.cpp
```diff
@@ -20,14 +19,9 @@ namespace claragenomics
...
-std::unique_ptr<FastaParser> create_hts_fasta_parser(const std::string& fasta_file)
+std::unique_ptr<FastaParser> create_kseq_fasta_parser(const std::string& fasta_file, int min_sequence_length)
 {
-    return std::make_unique<FastaParserHTS>(fasta_file);
-}
-
-std::unique_ptr<FastaParser> create_kseq_fasta_parser(const std::string& fasta_file)
-{
-    return std::make_unique<FastaParserKseqpp>(FastaParserKseqpp(fasta_file));
+    return std::make_unique<FastaParserKseqpp>(FastaParserKseqpp(fasta_file, min_sequence_length));
```
ahehn-nv: Optional: No need to create a temporary `FastaParserKseqpp`, just call
`std::make_unique<FastaParserKseqpp>(fasta_file, min_sequence_length)`.

vellamike: done

common/io/src/kseqpp_fasta_parser.cpp
vellamike: good point. done

PR #313, opened 2020-03-04
[cudamapper] Call cudaSetDevice() for writer thread

### LINKED ISSUE #306: [cudamapper] `CachingDeviceAllocator` causing segfault when running with multi-GPU and cache eviction.
When running with multiple GPUs, a segfault is resulting if CGA is compiled with the flag `-Dcga_enable_allocator=ON`

Command I am running with: `cudamapper /data/20k_reads_shuffled-chrX.fasta /data/20k_reads_shuffled-chrX.fasta -i 30 -w 5 -k 15 -c 30 -F 1e-5 -d 2 >out`

Interestingly, if [device-side cache eviction](https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudamapper/src/main.cu#L351) is disabled, the segfault disappears.

When compiling with  `-Dcga_enable_allocator=OFF` there is no segfault.

Stderr:

```
 time /home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamapper /data/20k_reads_shuffled-chrX.fasta /data/20k_reads_shuffled-chrX.fasta -i 30 -w 5 -k 15 -c 30 -F 1e-5 -d 2 > out
Query /data/20k_reads_shuffled-chrX.fasta index 20000
Target /data/20k_reads_shuffled-chrX.fasta index 20000
Processing query range: (0 - 1116)
Processing query range: (1117 - 2230)
Processing query range: (2231 - 3257)
Processing query range: (3258 - 4370)
Processing query range: (4371 - 5444)
[2]    28589 segmentation fault (core dumped)  /home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamappe
/home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamappe  21.83s user 2.21s system 353% cpu 6.806 total
```

mimaric:
When a new host thread is created it does not inherit `current device` from its parent, it simply gets assigned `device 0`. This can cause problems because two threads use different device/contexts.

This PR sets device in writer threads to same device as their launching worker thread.

This PR also introduces `CGA_CU_ABORT_ON_ERR`

Closes #306

### DIFF
--- a/common/utils/include/claragenomics/utils/allocator.hpp
+++ b/common/utils/include/claragenomics/utils/allocator.hpp
@@ -215,7 +215,7 @@ public:
     pointer allocate(std::size_t n, cudaStream_t stream = 0)
     {
         void* ptr = 0;
-        CGA_CU_ABORT_ON_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
+        CGA_CU_CHECK_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
         return static_cast<pointer>(ptr);
     }
 
--- a/common/utils/include/claragenomics/utils/allocator.hpp
+++ b/common/utils/include/claragenomics/utils/allocator.hpp
@@ -215,7 +215,7 @@ public:
     pointer allocate(std::size_t n, cudaStream_t stream = 0)
     {
         void* ptr = 0;
-        cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream);
+        CGA_CU_CHECK_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
         return static_cast<pointer>(ptr);
     }
 
@@ -226,7 +226,7 @@ public:
     void deallocate(pointer p, std::size_t n, cudaStream_t stream = 0)
     {
         // deallocate should not throw execeptions which is why CGA_CU_CHECK_ERR is not used.
-        cub_allocator_->DeviceFree(p);
+        CGA_CU_CHECK_ERR(cub_allocator_->DeviceFree(p));
     }
 
     /// @brief returns a shared pointer to internally used cub::CachingDeviceAllocator
--- a/common/utils/include/claragenomics/utils/allocator.hpp
+++ b/common/utils/include/claragenomics/utils/allocator.hpp
@@ -215,7 +215,7 @@ public:
     pointer allocate(std::size_t n, cudaStream_t stream = 0)
     {
         void* ptr = 0;
-        CGA_CU_CHECK_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
+        CGA_CU_ABORT_ON_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
         return static_cast<pointer>(ptr);
     }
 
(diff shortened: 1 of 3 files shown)


common/utils/include/claragenomics/utils/allocator.hpp
```diff
@@ -226,7 +226,7 @@ class CachingDeviceAllocator
     void deallocate(pointer p, std::size_t n, cudaStream_t stream = 0)
     {
         // deallocate should not throw execeptions which is why CGA_CU_CHECK_ERR is not used.
```
r-mafi: this comment could be removed now.

mimaric: @ahehn-nv @akkamesh:
@r-mafi raised a good point here. We probably should not throw here, but on the other hand we need a way to let the developer/user know that something went wrong. Any idea how to solve this? Having the application crash does sound a extreme, but if deallocation does actually fail...

r-mafi: @mimaric in the article you posted (https://akrzemi1.wordpress.com/2011/09/21/destructors-that-throw/), it mentions: 
"...For the end, note that while no destructor in C++ Standard Library is allowed to throw exceptions, destructors are allowed to std::terminate the program in case of an error..."
CGA_CU_CHECK_ERR calls `assert(false)` and `std::abort();` so IMO it should be OK.

ahehn-nv: Yes, you are right - given our current "fail hard" error handling this is no problem.
However, what if we try to change that sometime in the future? Then we need to go through every destructor and fix that again.
I'm not sure if this is a purely hypothetical problem, since it is unlikely that we will change this. Nevertheless, an easy future-prove method would be to introduce a `CGA_CU_ABORT_ON_ERR()`, which does the same as `CGA_CU_CHECK_ERR()` but allows us to distinguish the two if we ever need to.
Any opinions on that?

mimaric: Yes, I like the idea of having `CGA_CU_ABORT_ON_ERR()`. I'll implement it and update this PR

mimaric: done

common/utils/include/claragenomics/utils/cudautils.hpp
```diff
@@ -32,6 +32,16 @@
     {                                                                    \
         claragenomics::cudautils::gpu_assert((ans), __FILE__, __LINE__); \
     }
+// ^^^^ CGA_CU_CHECK_ERR currently has the same implementation as CGA_CU_ABORT_ON_ERR.
+//      The idea is that in the future CGA_CU_CHECK_ERR could have a "softer" error reporting (= not calling std::abort)
+
+/// \ingroup cudautils
+/// \def CGA_CU_ABORT_ON_ERR
+/// \brief Log on CUDA error in enclosed expression
```
ahehn-nv: maybe add "and abort the program."?

common/utils/include/claragenomics/utils/allocator.hpp
```diff
@@ -215,7 +215,7 @@ class CachingDeviceAllocator
     pointer allocate(std::size_t n, cudaStream_t stream = 0)
     {
         void* ptr = 0;
-        cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream);
+        CGA_CU_ABORT_ON_ERR(cub_allocator_->DeviceAllocate(&ptr, n * sizeof(T), stream));
```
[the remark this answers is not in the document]
mimaric: yes, done

common/utils/include/claragenomics/utils/cudautils.hpp
mimaric: done (specified both release and debug modes)

PR #320, opened 2020-03-09
[cudamapper] Use custom stream

### LINKED ISSUE #317: [cudamapper] Have all work for one iteration in one stream
Currently different kernels of one query - target read pair (iteration) use different streams without a specific need for that.

Make all kernels of one iteration use the same stream. This will
- prevent possible bugs due to incorrectly synchronized streams
- enable us to to use stream sync deallocations in caching allocators
- make it easier to overlap fetching indices from host cache with matcher-overlapper calculations
- make the profiles cleaner

mimaric:
Indexer, Matcher and Overlapper now all use (the same) custom stream. This will allow us to overlap things more easily.

@tijyojwad CIGAR (PR #307) uses its own custom stream. We should revisit this and make sure that is what we want to have

Closes #317


vellamike:
@mimaric  did you run on nsight compute and verify that the default steram is not engaged at all?

mimaric:
> @mimaric did you run on nsight compute and verify that the default stream is not engaged at all?
Yes, everything was in one (not default) stream

PR #325, opened 2020-03-18
fix #309 validity checks for query and target files

### LINKED ISSUE #309: [cudamapper] no useful error message when cudamapper pointed to non-existent file.
A useful error is needed. Right now application does not crash but ends with:

```
Processing query range (0 - -1)
```

rached-ab:


vellamike:
@ohadmo any idea why CI is not executing?

ohadmo:
@vellamike, seems like there is a long queue caused by cudf

cudamapper/src/main.cu
```diff
@@ -135,8 +136,21 @@ int main(int argc, char* argv[])
     }
 
     std::string query_filepath  = std::string(argv[optind++]);
-    std::string target_filepath = std::string(argv[optind++]);
+    std::ifstream query_file(query_filepath);
+    if (query_file.fail())
```
tijyojwad: @rached-ab what do you think about moving this check into the `create_kseq_fasta_parser` function and throwing an exception if the file doesn't exist? @ahehn-nv would a `runtime_error` be appropriate for that?

tijyojwad:
@rached-ab for the style check failures, take a look at the code formatting (and auto formatting) option we have for the repo - https://github.com/clara-genomics/ClaraGenomicsAnalysis#code-formatting

cudamapper/src/main.cu
ahehn-nv: Yes, I think would make sense.
I.e. put the check into the fasta parser class itself
https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/common/io/src/kseqpp_fasta_parser.cpp#L29
maybe SeqStreamIn also has a `.fail()` method, and in case of failure throw an std::runtime_exception there.

tijyojwad: @rached-ab if you have any clarifying questions let us know!

rached-ab: Thank you! 
SeqStreamIn does not seem to have any useful method to check. I am considering throwing invalid_argument exception from fasta parser class using ifstream check. Please let me know if you think of a better way.

PR #327, opened 2020-03-20
allowing custom size sequence, graph etc. in cudaPOA

r-mafi:
removed some of hard-coded macros defined in cudapoa_kernels

```
 ( CUDAPOA_MAX_NODES_PER_WINDOW, 
  CUDAPOA_MAX_NODES_PER_WINDOW_BANDED,
  CUDAPOA_MAX_SEQUENCE_SIZE,
  CUDAPOA_MAX_CONSENSUS_SIZE,
  CUDAPOA_MAX_MATRIX_GRAPH_DIMENSION,
  CUDAPOA_MAX_MATRIX_GRAPH_DIMENSION_BANDED)
```
and replaced them with a structure defined in batch.hpp `struct BatchSize`
__________________________________________________________________________________
also added a new sample for running POA on a few long reads. This can be enabled by input arg `-l`
POA graph can be printed with input `-g`


r-mafi:
@tijyojwad replacing macros with kernel input in a first glance by running cudapoa_benchmarks does not seem to be any different. May require more through testing.

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -44,6 +44,43 @@ struct Entry
...
+    /// Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+    uint32_t max_matrix_graph_dimension = max_nodes_per_window + 4;
+    /// Maximum vertical dimension of scoring matrix, which stores graph
+    /// Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+    uint32_t max_matrix_graph_dimension_banded = max_nodes_per_window_banded + 4;
+    /// Maximum horizontal dimension of scoring matrix, which stores sequences
+    /// Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+    uint32_t max_matrix_sequence_dimension = max_sequence_size + 4;
+    /// ToDO add banded alignment score matrix dimension parameters (maybe?)
+    //uint32_t max_matrix_sequence_dimension_banded;
```
tijyojwad: this isn't necessary because one of the dimensions of the banded matrix is just the bandwidth

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -44,6 +44,43 @@ struct Entry
 /// A type defining the set and order of Entry's in which a POA is processed.
 typedef std::vector<Entry> Group;
 
+/// A structure to hold  upper limits for data size processed in POA batches
+struct UpperLimits
+{
```
tijyojwad: we can also add the `max_sequences_per_poa` into this struct (the first param to the `create_batch` call) and rename this struct more appropriately. I'll keep thinking on it :)

cudapoa/src/cudapoa_generate_msa.cu
```diff
@@ -180,14 +184,15 @@ __global__ void generateMSAKernel(uint8_t* nodes_d,
                                        node_marks,
                                        check_aligned_nodes,
                                        nodes_to_visit,
-                                       cuda_banded_alignment);
+                                       cuda_banded_alignment,
+                                       (uint16_t)max_nodes_per_window);
```
tijyojwad: this is going to get tricky. I think we need to check for limits on the size of the types, or convert them all to `uint32_t`. The casting may get problematic (although unlikely in this case) if 65k+ number gets rolled over to uint16_t

cudapoa/tests/Test_CudapoaNW.cpp
```diff
@@ -197,26 +197,28 @@ NWAnswer testNW(const BasicNW& obj)
     int16_t mismatch_score;
     int16_t match_score;
     uint16_t* aligned_nodes; //local; to store num of nodes aligned (length of alignment_graph and alignment_read)
+    UpperLimits max_limits;
+    max_limits.setLimits(1024);
```
tijyojwad: do we need this here if the default is `1024`?

r-mafi: you're right, considering that the default is 1024, it's not needed. At the time of adding this line, I thought just in case the default value changes later, this will ensure the Test file behaves as before. Will remove and add a comment.

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -44,6 +44,43 @@ struct Entry
...
+
+    /// set upper limit parameters based on max_sequence_size
+    void setLimits(const uint32_t max_seq_sz)
+    {
+        max_sequence_size           = max_seq_sz;
+        max_concensus_size          = max_sequence_size;
+        max_nodes_per_window        = 3 * max_sequence_size;
+        max_nodes_per_window_banded = 4 * max_sequence_size;
+        // Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+        max_matrix_graph_dimension        = max_nodes_per_window + 4;
```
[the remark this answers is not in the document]
r-mafi: done.

cudapoa/include/claragenomics/cudapoa/batch.hpp
r-mafi: I agree, removed those comments.

common/utils/include/claragenomics/utils/genomeutils.hpp
```diff
@@ -86,6 +86,78 @@ inline std::string generate_random_sequence(const std::string& backbone, std::mi
     return sequence;
 }
 
+// max_insertions, max_deletions and max_mutations are capped to range length
+inline std::string generate_random_sequence_within_range(const std::string& backbone, std::minstd_rand& rng, std::vector<std::pair<int, int>> ranges, int max_mutations, int max_insertions, int max_deletions)
```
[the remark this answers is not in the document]
r-mafi: yes, I can merged them into a same function, and pass the range vector as pointer (default null, not to change any previous usages.)

cudapoa/include/claragenomics/cudapoa/batch.hpp
r-mafi: yes, good idea. May be something like POASize, POADimensions, BatchSize, BatchDimensions...?

cudapoa/src/cudapoa_generate_msa.cu
r-mafi: I agree, this is not safe for cases where number of nodes per window exceeds INT16_MAX = 32767. However, if we want to make it template, we got to template all data structures in cudapoa_kernels.cuh (WindowDetails, InputDetails, GraphDetails...).

cudapoa/src/cudapoa_batch.hpp
```diff
@@ -162,12 +604,19 @@ class CudapoaBatch : public Batch
...
 
     // Maximum POAs to process in batch.
     int32_t max_poas_ = 0;
+
+public:
+    // Static batch count used to generate batch IDs.
+    static int32_t batches;
 };
 
+template <typename ScoreT>
```
tijyojwad: this is just a counter for how many cudapoa batch objects are created. this doesn't need to be templated by `ScoreT`. it's more like an ID of the batch, not the nodes in the graph

cudapoa/src/batch.cpp
```diff
@@ -18,25 +18,43 @@ namespace claragenomics
...
-                                          device_id,
-                                          stream,
-                                          max_mem,
-                                          output_mask,
-                                          gap_score,
-                                          mismatch_score,
-                                          match_score,
-                                          cuda_banded_alignment);
+    // a decision flag to determine proper type definition for ScoreT
+    const bool use_32_bit_for_ScoreT = use32bitInt(batch_size.max_nodes_per_window, match_score);
```
tijyojwad: since this file is using `use32bitInt` from `cudapoa_kernels.cuh`, let's explicitly include that header into this file

cudapoa/include/claragenomics/cudapoa/batch.hpp
r-mafi: added `max_sequences_per_poa` and changed struct name to `BatchSize`

common/utils/include/claragenomics/utils/genomeutils.hpp
```diff
@@ -46,43 +46,105 @@ inline std::string generate_random_sequence(const std::string& backbone, std::mi
 
     std::string sequence = backbone;
 
-    std::uniform_real_distribution<double> random_prob(0, 1);
-
-    for (int j = 0; j < max_deletions; j++)
+    if (ranges == nullptr)
```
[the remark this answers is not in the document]
r-mafi: the reason that I had it in a separate function in the first place was that `for loop` bounds are defined slightly different. I did not want to define an array of max  deletions, insertions and mutations, so they remain scalar. To keep changes within bound, the upper limit is capped to min(max_in/del/mut , range_length). I can change the original code to be consistent with mine, then can merge them into one.

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -44,6 +45,60 @@ struct Entry
...
+    /// Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+    uint32_t max_matrix_graph_dimension_banded = max_nodes_per_window_banded + 4;
+    /// Maximum horizontal dimension of scoring matrix, which stores sequences
+    /// Adding 4 elements more to ensure a 4byte boundary alignment for any allocated buffer
+    uint32_t max_matrix_sequence_dimension = max_sequence_size + 4;
+    /// Maximum number of equences per POA group
+    uint32_t max_sequences_per_poa = 100;
+
+    /// set upper limit parameters based on max_sequence_size
+    void setSize(const uint32_t max_seq_sz, const uint32_t max_num_seqs)
```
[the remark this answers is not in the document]
r-mafi: yes, before adding `setParams`, I was also thinking to use constructor instead of `setSize()`. will change it.

cudapoa/src/batch.cpp
r-mafi: done.

cudapoa/src/cudapoa_batch.hpp
r-mafi: right, but it is defined inside CudapoaBatch class (a couple of lines above it) which this class is now templated. So can't call it like `CudapoaBatch::batches = 0;` you want to take it out from the class, or just leave it as it is.

tijyojwad: Ah that's right! My bad, I got confused. Yes let's leave this as it is.

common/utils/include/claragenomics/utils/genomeutils.hpp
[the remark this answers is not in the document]
tijyojwad: Yes, let's make the original code to be consistent with your new version. This function is used in some cudaaligner unit tests as well, so would be good to run them and make sure it's all backwards compatible

cudapoa/include/claragenomics/cudapoa/batch.hpp
tijyojwad: Thanks!

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -155,11 +252,18 @@ int main(int argc, char** argv)
         std::cout << "Usage:" << std::endl;
         std::cout << "./sample_cudapoa [-m] [-h]" << std::endl;
         std::cout << "-m : Generate MSA (if not provided, generates consensus by default)" << std::endl;
+        std::cout << "-l : Perform long-read sample (if not provided, will run window-based sample by default)" << std::endl;
         std::cout << "-p : Print the MSA or consensus output to stdout" << std::endl;
         std::cout << "-h : Print help message" << std::endl;
         std::exit(0);
     }
 
+    if (long_read)
```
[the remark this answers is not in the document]
r-mafi: good idea, done.

cudapoa/include/claragenomics/cudapoa/batch.hpp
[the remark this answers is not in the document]
r-mafi: done.

common/utils/include/claragenomics/utils/genomeutils.hpp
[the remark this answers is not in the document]
r-mafi: done.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -203,25 +288,23 @@ int main(int argc, char** argv)
...
         if (status != StatusType::exceeded_maximum_poas && status != StatusType::success)
         {
             std::cerr << "Could not add POA group to batch. Error code " << status << std::endl;
+            error_count++;
+            if (error_count > get_size(windows))
+                break;
         }
     }
 
+    if (print && long_read)
```
tijyojwad: we should add a `print_graph` option, and print it for both long and short read

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -155,24 +195,39 @@ int main(int argc, char** argv)
...
+    // data for short reads has many such POA groups to process, hence the data is loaded into a vector
+    // of vector of strings. Long read sample creates one POA group.
     std::vector<std::vector<std::string>> windows;
-    parse_window_data_file(windows, input_data, 1000); // Generate windows.
-    assert(get_size(windows) > 0);
+
+    // Define upper limits for sequence size, graph size ....
+    BatchSize batch_size;
+
+    if (!long_read)
```
tijyojwad: optional : can you switch this to `if (long_read) else` instead of `if (!long_read) else`?

cudapoa/src/batch.cpp
```diff
@@ -18,25 +19,43 @@ namespace claragenomics
 namespace cudapoa
 {
 
-std::unique_ptr<Batch> create_batch(int32_t max_sequences_per_poa,
-                                    int32_t device_id,
+std::unique_ptr<Batch> create_batch(int32_t device_id,
                                     cudaStream_t stream,
                                     size_t max_mem,
                                     int8_t output_mask,
+                                    BatchSize batch_size,
```
tijyojwad: can we make `BatchSize` the first parameter? And take it in as a const ref? `const BatchSize&`

cudapoa/src/cudapoa_kernels.cu
```diff
@@ -649,12 +685,75 @@ void generatePOA(claragenomics::cudapoa::OutputDetails* output_details_d,
...
+                                                              max_sequences_per_poa,
+                                                              output_mask,
+                                                              batch_size);
+    }
+}
+
+bool use32bitInt(int32_t max_sequence_size, int32_t match_score)
+{
+    // a decision flag to determine proper type definition for ScoreT, factor 2 is selected ad-hoc
+    return (max_sequence_size * match_score * 2 > INT16_MAX);
```
tijyojwad: We can probably update the limits for this to be the following - 

max - `max_sequence_size * match_score`. This will only happen when sequence and graph match exactly. any other scenario will cause gap scores to be added which (as long as they're negative) will drive down the score.

min - no match whatsoever, so basically all gaps. and this would be `(max_sequence_size + max_graph_size) * gap_score`. 

what do you think?

cudapoa/samples/sample_cudapoa.cpp
r-mafi: do we really need to add print graph option for short reads? the current sample generates 1000 poa groups, that would be 1000 poa graphs.

cudapoa/samples/sample_cudapoa.cpp

cudapoa/src/batch.cpp
r-mafi: done.

cudapoa/src/cudapoa_kernels.cu
r-mafi: good point, I changed the condition to make decision based on both min and max, and removed the ad-hoc factor. may need to revise the min condition for banded.

r-mafi: I replaced with the following logic which is more accurate and less conservative: if `d1 = max_sequence_size`, and `d2 = max_graph_size (assuming d2>=d1)`, `n = mismatch_score` and `g = gap_score`, where `n` and `g` are negative. Then minimum negative score computed in NW matrix should be bound to:
`d1*(max(g,n)) + (d2-d1)*g`

cudapoa/include/claragenomics/cudapoa/batch.hpp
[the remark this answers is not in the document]
r-mafi: I think it's better we add something like `throw std::invalid_argument` in case of using the constructors that takes all parameters is used and it violates some assumptions, like `max_sequence_size <= max_graph_size`, etc. what do you think @tijyojwad ?

cudapoa/samples/sample_cudapoa.cpp
r-mafi: although keeping symmetry between short-read and long-read samples is preferred, but considering large number of graphs for short-read sample, I modified the help message to better reflect behavior of -p for short-read and long-read.

r-mafi: alternatively we can add this print-graph option and allow its use only along -l flag. I think this is better, will change it that way.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -155,24 +199,40 @@ int main(int argc, char** argv)
         std::cout << "Usage:" << std::endl;
         std::cout << "./sample_cudapoa [-m] [-h]" << std::endl;
         std::cout << "-m : Generate MSA (if not provided, generates consensus by default)" << std::endl;
+        std::cout << "-l : Perform long-read sample (if not provided, will run window-based sample by default)" << std::endl;
```
tijyojwad: `will run short-read sample by default`

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -155,24 +199,40 @@ int main(int argc, char** argv)
...
+    // Define upper limits for sequence size, graph size ....
+    BatchSize batch_size;
+
+    if (long_read)
+    {
+        generate_simulated_long_reads(windows, batch_size);
+    }
+    else
+    {
+        const std::string input_data = std::string(CUDAPOA_BENCHMARK_DATA_DIR) + "/sample-windows.txt";
```
tijyojwad: optional : for symmetry would be nice to move this section to a `generate_short_reads` function which parses the file and gets the `BatchSize` limits from the data in `windows`, just like the long read version does. This will be another test that our heuristic limits for BatchSize work well

cudapoa/src/batch.cpp
tijyojwad: I see the `const ref` change but not the BatchSize as the first param. I am little confused what we decided on now for it :)

cudapoa/include/claragenomics/cudapoa/batch.hpp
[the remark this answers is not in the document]
tijyojwad: Yes that would be really good!

cudapoa/src/cudapoa_kernels.cu
tijyojwad: Looks good to me! Can you add a comment explaining the min and max score calculations in the function?

tijyojwad:
@r-mafi also seeing some `cudamapper` files in this PR. maybe they got added by mistake?

cudapoa/src/batch.cpp
r-mafi: I thought it was agreed to keep device and stream ids as first parameters for now, since in many other constructors they come first. To keep consistency, I left the order as it is for ow and only changed the type to const ref. sorry for confusion, it looks we concluded our discussion with different takes!

cudapoa/src/cudapoa_kernels.cu
r-mafi: done.

cudapoa/samples/sample_cudapoa.cpp
r-mafi: good catch, I kept to skip reading it all this time! changed.

cudapoa/samples/sample_cudapoa.cpp
r-mafi: done

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -44,6 +45,58 @@ struct Entry
 /// A type defining the set and order of Entry's in which a POA is processed.
 typedef std::vector<Entry> Group;
 
+/// A structure to hold  upper limits for data size processed in POA batches
+struct BatchSize
+{
+    /// Maximum number of elements in a sequence
+    uint32_t max_sequence_size;
```
tijyojwad: there's one thing I forgot to mention @r-mafi - earlier in our development we decided to use `int32_t` for integer data types, and then use the `throw_on_negative` from https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/common/utils/include/claragenomics/utils/signed_integer_utils.hpp#L39 to catch negative numbers. this helps catch cases where use inputs a negative numbers to the API with type as `uint32_t`

tijyojwad: e.g. https://github.com/clara-genomics/ClaraGenomicsAnalysis/pull/327/files#diff-c4fc2db9ecbc3d4544f4422682e91668R59

r-mafi:
> @r-mafi also seeing some `cudamapper` files in this PR. maybe they got added by mistake?

somehow git merge had reverted some changes on `cudamapper`. fixed!

cudapoa/include/claragenomics/cudapoa/batch.hpp
r-mafi: sure, changed to int32_t. This may require updating the python binding again.

cudapoa/include/claragenomics/cudapoa/batch.hpp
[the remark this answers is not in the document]
r-mafi: I added few checks, but it's difficult to cover all invalid scenarios.

PR #331, opened 2020-03-24
[common] minor fixes

ahehn-nv:
* [common] allocator added missing include + fixed unused variable warnings
* [common] buffer: avoid buffer(Allocator, cudaStream = 0) constructor for buffer(n) calls.
* [common] preallocated allocator: minor changes:
  * initialize stream member of first memory block
  * make create_buffer(buffer_size) use argument instead of buffer_size_ member

### DIFF
--- a/common/utils/include/claragenomics/utils/allocator.hpp
+++ b/common/utils/include/claragenomics/utils/allocator.hpp
@@ -13,6 +13,7 @@
 #include <memory>
 #include <type_traits>
 
+#include <cuda_runtime_api.h>
 #include <cub/util_allocator.cuh>
 #include <claragenomics/utils/device_preallocated_allocator.cuh>
 
@@ -46,6 +47,7 @@ public:
     template <typename U>
     CudaMallocAllocator(const CudaMallocAllocator<U>& rhs)
     {
+        static_cast<void>(rhs);
     }
 
     /// @brief copy assignment operator
@@ -102,6 +104,7 @@ public:
     /// @return pointer to allocated array
     pointer allocate(std::size_t n, cudaStream_t stream = 0)
     {
+        static_cast<void>(stream);
         void* ptr = 0;
         CGA_CU_CHECK_ERR(cudaMalloc(&ptr, n * sizeof(T)));
         return static_cast<pointer>(ptr);
@@ -113,6 +116,8 @@ public:
     /// @param stream CUDA stream to be associated with this method.
     void deallocate(pointer p, std::size_t n, cudaStream_t stream = 0)
     {
+        static_cast<void>(n);
+        static_cast<void>(stream);
         CGA_CU_ABORT_ON_ERR(cudaFree(p));
     }
 };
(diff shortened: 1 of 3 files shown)


common/utils/include/claragenomics/utils/buffer.hpp
```diff
@@ -70,7 +70,7 @@ class buffer
     /// @param allocator The allocator to use by this buffer.
     /// @param stream The CUDA stream to be associated with this allocation. Default is stream 0.
     /// @tparam AllocatorIn Type of input allocator. If AllocatorIn::value_type is different than T AllocatorIn will be converted to Allocator<T> if possible, compilation will fail otherwise
-    template <typename AllocatorIn>
+    template <typename AllocatorIn, std::enable_if_t<std::is_class<AllocatorIn>::value, int> = 0>
```
mimaric: Why cannot it be union? A comment would help

common/utils/include/claragenomics/utils/allocator.hpp
```diff
@@ -46,6 +47,7 @@ class CudaMallocAllocator
     template <typename U>
     CudaMallocAllocator(const CudaMallocAllocator<U>& rhs)
     {
+        static_cast<void>(rhs);
```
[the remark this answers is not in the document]
ahehn-nv: Yes, exactly.
I don't think the comment is much help. It is not an idea that requires explanation, but merely a C++ expression. Since the only use of `static_cast<void>(x)` is to avoid unused variable warnings (to my knowledge), and it can be easily found by searching on the internet, this should be fine without a comment.

common/utils/include/claragenomics/utils/buffer.hpp
ahehn-nv: Yes, agreed, a comment would help here.
I don't think this can be a union, we assume that an Allocator provides `allocate()` and `deallocate()` functions and so on...

ahehn-nv: done

mimaric: yes, makes sense

common/utils/include/claragenomics/utils/allocator.hpp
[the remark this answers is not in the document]
mimaric: I would still add a short comment `static_cast<void>(rhs); // has no effect, here to suspend unused variable warning`. It simply reduces the number of WTFs_per_line_of_code for less experienced developers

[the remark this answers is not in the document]
ahehn-nv: Yes, I agree that it is not the most common C++ expression, but it's really just a C++ language thing, nothing elaborate.
The reason I'm hesitant to add the comment is that I will have to add it to every static_cast<void>() then.
It might be a bit awkward the first time you see it, but once you've understood it, all comments are only noise.
If a curious developer stumbles over it, they will look it up, in my opinion.

Issue #318, opened 2020-03-05
[cudamapper] Overlap fetching indices from host memory with matcher+overlapper


mimaric:
Fetch new target index from host cache while matcher+overlapper work on current target index.

On current benchmarks fetching index from device cache takes approximately the same amount of time as doing mathcer + overlapper for it, so it makes sense to overlap those two activities.

Work on this issue can start after issue #317 has been done

PR #337, opened 2020-03-26
[cudamapper] Renamed HostCache to IndexHostCopy

mimaric:
`HostCache` class name is a bit misleading. `IndexHostCopy` reflects the purpose better as the object is an actual copy of an index in host memory.
I'm not completely happy with reaming base `IndexHostCopy` into `IndexHostCopyBase`, but I do feel that `IndexHostCopy` is something we should expose externally.

This PR is part of preparations for the main PR for issue #318

### DIFF
--- a/cudamapper/include/claragenomics/cudamapper/index.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/index.hpp
@@ -134,6 +134,9 @@ public:
     virtual std::unique_ptr<Index> copy_index_to_device(DefaultDeviceAllocator allocator,
                                                         const cudaStream_t cuda_stream = 0) = 0;
 
+    /// \brief virtual destructor
+    virtual ~IndexHostCopyBase() = delete;
+
     /// \brief returns an array of representations of sketch elements (stored on host)
     /// \return an array of representations of sketch elements
     virtual const std::vector<representation_t>& representations() const = 0;
--- a/cudamapper/include/claragenomics/cudamapper/index.hpp
+++ b/cudamapper/include/claragenomics/cudamapper/index.hpp
@@ -121,10 +121,10 @@ public:
                  const cudaStream_t cuda_stream   = 0);
 };
 
-/// IndexHostCopy - Creates and maintains a copy of computed IndexGPU elements on the host, then allows to retrieve target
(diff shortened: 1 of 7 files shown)


cudamapper/include/claragenomics/cudamapper/index.hpp
```diff
@@ -121,10 +121,10 @@ class Index
                  const cudaStream_t cuda_stream   = 0);
 };
 
-/// IndexHostCopy - Creates and maintains a copy of computed IndexGPU elements on the host, then allows to retrieve target
+/// IndexHostCopyBase - Creates and maintains a copy of computed IndexGPU elements on the host, then allows to retrieve target
 /// indices from host instead of recomputing them again
 ///
-class IndexHostCopy
+class IndexHostCopyBase
```
ahehn-nv: Not your mistake, but while you are at it: could you add a virtual destructor to the base class?

cudamapper/src/index_host_copy.cuh
```diff
@@ -30,11 +30,11 @@ public:
     /// \param window_size the number of adjacent k-mers in a window, adjacent = shifted by one basepair
     /// \param cuda_stream D2H copy is done on this stream
     /// \return - pointer to claragenomics::cudamapper::IndexCache
-    explicit HostCache(const Index& index,
-                       const read_id_t first_read_id,
-                       const std::uint64_t kmer_size,
-                       const std::uint64_t window_size,
-                       const cudaStream_t cuda_stream);
+    explicit IndexHostCopy(const Index& index,
```
ahehn-nv: this constructor does not need to be `explicit`.

cudamapper/include/claragenomics/cudamapper/index.hpp
mimaric: done

cudamapper/src/index_host_copy.cuh
mimaric: done

cudamapper/include/claragenomics/cudamapper/index.hpp
ahehn-nv: it needs to be `= default;` not `= delete;` :)

mimaric: whoops :)

PR #338, opened 2020-03-26
[cmake] fixes default cmake install path setting

tijyojwad:
Fix for https://github.com/lbcb-sci/racon/issues/17

Add check to only update default install path if CGA is being built as the main project


tijyojwad:
@ohadmo we should, but no rush on doing that. we can upload to pypi once the nightly auto release functionality is available

PR #342, opened 2020-03-30
[common] some utils improvements

ahehn-nv:
* Move find_largest_contiguous_device_memory_section() from cudamapper to utils
* Remove utils dependency from logging library
* CachingDeviceAllocator: Change the default-construction behavior


ahehn-nv:
Ok, I broke it for CUDA <=9.2.... Will fix it.

ahehn-nv:
I will file a new updated PR.

PR #344, opened 2020-03-31
Cudamapper check input files fix309

rached-ab:


common/io/src/kseqpp_fasta_parser.cpp
```diff
@@ -28,16 +30,34 @@ FastaParserKseqpp::FastaParserKseqpp(const std::string& fasta_file, int min_sequ
 {
     klibpp::KSeq record;
     klibpp::SeqStreamIn iss(fasta_file.data());
+
+    /*    std::ifstream query_file(fasta_file);
```
tijyojwad: @rached-ab let's remove the commented out code

rached-ab: @tijyojwad do you think it is needed to add try/catch in non regression tests when parser in created ?

cudamapper/src/main.cu
```diff
@@ -289,17 +289,33 @@ int main(int argc, char* argv[])
     claragenomics::logging::Init();
 
     const ApplicationParameteres parameters = read_input(argc, argv);
+    std::shared_ptr<claragenomics::io::FastaParser> query_parser;
 
-    std::shared_ptr<claragenomics::io::FastaParser> query_parser = claragenomics::io::create_kseq_fasta_parser(parameters.query_filepath, parameters.k + parameters.w - 1);
-
+    try
```
ahehn-nv: While we should probably catch all exceptions in main, I would not do this for individual function calls.
For now I would rely on the C++ implementation to have a good handler for all uncaught exceptions which prints the error for us and ends the program.

rached-ab: done.

tijyojwad:
Closes #309

PR #348, opened 2020-04-01
[common] CachingDeviceAllocator: Change the default-construction behavior

ahehn-nv:


common/utils/include/claragenomics/utils/allocator.hpp
```diff
@@ -287,12 +306,21 @@ using DefaultDeviceAllocator = CachingDeviceAllocator<char, DevicePreallocatedAl
 using DefaultDeviceAllocator = CachingDeviceAllocator<char, cub::CachingDeviceAllocator>;
 #endif
 
+inline DefaultDeviceAllocator create_default_device_allocator(std::size_t max_caching_size = 2ull * 1024 * 1024 * 1024 /* 2GiB */)
```
mimaric, line 309: I try not to use `/* */` in comments as it gets messy if you try to comment a larger block of code that already contains `/* */`

ahehn-nv: I see we have a different way of commenting out larger blocks of code. :D
I'll fix it. Not very fond of `/* */` myself.
Actually, I should provide full documentation comments here...

mimaric: I usually add `//` in front of each line, but sometimes it's just easier to enclose it with `/* */` when trying out something quickly.
Yes, a comment would be useful there

ahehn-nv: Ok, so we do have the same commenting style... :)
Anyhow, fixed it.

PR #351, opened 2020-04-08
adding bonito data to cudaPOA

r-mafi:
A new function `generate_bonito_long_reads` is added to run long-read sample in cudapoa on the data extracted from bonito.
in this PR, the whole data set extracted from bonito is added in a text file `sample-bonito.txt`
this text file consists of 1000 windows (each consists of 6 sequences). The sample_cudapoa uses only first few windows, this file can be truncated to a much smaller size.


cudapoa/CMakeLists.txt
```diff
@@ -23,7 +23,7 @@ cuda_add_library(cudapoa ${cga_library_type}
     src/batch.cpp
     )
 
-target_link_libraries(cudapoa cgautils cgalogging)
+target_link_libraries(cudapoa cgautils cgalogging cgaio)
```
r-mafi, line 26: this was added in order to use fasta reader. But later the 6 fasta files were packed in a single text file, therefore `cgaio` is no longer needed and this change can be reverted.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -227,7 +246,7 @@ int main(int argc, char** argv)
 
     if (long_read)
     {
-        generate_simulated_long_reads(windows, batch_size);
+        generate_bonito_long_reads(windows, batch_size);
     }
```
r-mafi, line 250: I was not sure whether to keep or remove `generate_simulated_long_reads`, since now it's been replaced with `generate_bonito_long_reads`, there's no use for it other than debugging and internal benchmarking.

PR #353, opened 2020-04-09
adding bonito data to cudaPOA

r-mafi:
A new function generate_bonito_long_reads is added to run long-read sample in cudapoa on the data extracted from bonito.
in this PR, the whole data set extracted from bonito is added in a text file sample-bonito.txt
this text file consists of 1000 windows (each consists of 6 sequences). The sample_cudapoa uses only first few windows, this file can be truncated to a much smaller size.


r-mafi:
the previous branch got corrupted since by mistake changes for benchmarking were pushed to it, therefore I resent this PR.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -133,43 +133,23 @@ void generate_short_reads(std::vector<std::vector<std::string>>& windows, BatchS
...
-    variation_ranges.push_back(std::pair<int, int>(3000, 3500));
-    variation_ranges.push_back(std::pair<int, int>(4000, 4200));
-    variation_ranges.push_back(std::pair<int, int>(5000, 5400));
-    variation_ranges.push_back(std::pair<int, int>(6000, 6200));
-    variation_ranges.push_back(std::pair<int, int>(8000, 8300));
-
-    std::vector<std::string> long_reads(number_of_reads);
-
-    for (int w = 0; w < number_of_windows; w++)
+    const std::string input_data = std::string(CUDAPOA_BENCHMARK_DATA_DIR) + "/sample-bonito.txt";
```
tijyojwad, line 138: since this logic is now almost the same for short reads as well, let's move it to a single function and instead just change the name of file used to generate the windows vector based on long/short read option.

```
void generate_window_data(filename, ...)
{
    parse_window_file(...)
    // for loop to calculate longest sequence, etc.
}

if (long_read)
   generate_window_data(bonito_file,...)
else
   generate_window_data(short_read_file,...)
```

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -277,10 +257,17 @@ int main(int argc, char** argv)
                 }
             }
 
-            // After MSA is generated for batch, reset batch to make room for next set of POA groups.
+            // After MSA/consensus is generated for batch, reset batch to make room for next set of POA groups.
             batch->reset();
 
-            std::cout << "Processed windows " << window_count << " - " << i << std::endl;
+            if (status == StatusType::success)
```
tijyojwad, line 263: Could you elaborate a bit on this change? Not sure I'm quite following this

common/io/include/claragenomics/io/fasta_parser.hpp
```diff
@@ -59,9 +59,10 @@ class FastaParser
 ///
 /// \param fasta_file Path to FASTA(.gz) file. If .gz, it must be zipped with bgzip.
 /// \param min_sequence_length Minimum length a sequence needs to be to be parsed. Shorter sequences are ignored.
+/// \param shuffle Enables shuffling reads
 ///
 /// \return A unique pointer to a constructed parser object.
-std::unique_ptr<FastaParser> create_kseq_fasta_parser(const std::string& fasta_file, int min_sequence_length = 0);
+std::unique_ptr<FastaParser> create_kseq_fasta_parser(const std::string& fasta_file, int min_sequence_length = 0, bool shuffle = true);
```
tijyojwad, line 65: @r-mafi since this change isn't directly used by this PR, let's create a separate PR with this `shuffle` option so cudamapper folks can review it

cudapoa/samples/sample_cudapoa.cpp
r-mafi: in case than number of windows is more than the capacity available on GPU, the for loop breaks into smaller number of windows. If adding window `i` in `batch->add_poa_group` turns out to exceed the limit, it wont be processed, therefore we print `i-1` to account for the fact that window `i` was excluded at that round. 
I can add this explanation as a comment to make this part of code more clear.
```
            if (status == StatusType::success)
            {
                std::cout << "Processed windows " << window_count << " - " << i << std::endl;
            }
            else
            {
                std::cout << "Processed windows " << window_count << " - " << i - 1 << std::endl;
            }
```

cudapoa/samples/sample_cudapoa.cpp
r-mafi: yes, the reason it is remained separate was initially they were quite different, but you're right, at this point, we can compact them into one function.

common/io/include/claragenomics/io/fasta_parser.hpp
r-mafi: since it was minor without any impact on the current code (...,`bool shuffle = true`), I found it harmless to be here. But OK, will submit in a new PR and ask Mike to review.

cudapoa/samples/sample_cudapoa.cpp

cudapoa/samples/sample_cudapoa.cpp
r-mafi: added comment to clarify.

common/io/include/claragenomics/io/fasta_parser.hpp
r-mafi: done.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -227,11 +199,11 @@ int main(int argc, char** argv)
 
     if (long_read)
     {
-        generate_simulated_long_reads(windows, batch_size);
+        generate_window_data("/sample-bonito.txt", 8, 6, windows, batch_size);
```
tijyojwad, line 202: let's create the filename outside of the function, and pass in the full path to the file into the function

i.e. move the `const std::string input_data = std::string(CUDAPOA_BENCHMARK_DATA_DIR) + file_name;` from the function definition into this conditional

PR #357, opened 2020-04-10
Fasta parser shuffle

r-mafi:
there are cases where shuffling reads is not desired. An example is parsing fasta files from Bonito where correspondence between reads of different fasta files need to be preserved. This PR adds a flag to allow disabling shuffling in fasta parser. By default, this flag is set to true, i.e. won't impact previous codes in anyway.


mimaric:
Great! I was thinking about implementing this as well. Shuffling can also cause problems with our testing

PR #359, opened 2020-04-13
[cga] Renamed cgautils -> cgabase and introduced SDK-wide types

mimaric:
Although `read_id_t` and `position_in_read_t` are currently only used in cudamapper they are actually SDK-wide concepts so they can be moved to `common`.

The background of this PR is that I'd like to make types in `FastaParser` more verbose.
For example `get_sequence_by_id()` accepts `int32_t`. Having is accept `read_id_t` would make this way more expressive.

This PR is a part of issue #318


common/defs/CMakeLists.txt
```diff
@@ -0,0 +1,44 @@
...
+#
+
+project(cgadefs)
+
+find_package(CUDA 9.0 QUIET REQUIRED)
+
+add_library(${PROJECT_NAME} INTERFACE)
+target_link_libraries(${PROJECT_NAME} INTERFACE ${CUDA_LIBRARIES})
+
+if (cga_profiling)
```
tijyojwad, line 18: I believe all the CUDA libs, profiling macros and debugging macros are taken care of in the `cgaconfig` project, so they don't need to be added again to new ones. Is that right @ahehn-nv ?

mimaric: Probably, I largely copied it from `cgaconfig`. Is it the necessary to include `cgaconfig` somehow? I'm still having problems what gets included and how in our structure.

ahehn-nv: Just do a `target_link_libraries(${PROJECT_NAME} INTERFACE cgaconfig)`.

In general, you should only need
```
add_library(${PROJECT_NAME} INTERFACE)
target_link_libraries(${PROJECT_NAME} INTERFACE cgaconfig)target_include_directories(${PROJECT_NAME}
    INTERFACE 
        $<INSTALL_INTERFACE:include>    
        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
)

install(TARGETS ${PROJECT_NAME}
    EXPORT ${PROJECT_NAME}
    INCLUDES DESTINATION include
)
install(DIRECTORY include/ DESTINATION include)
install(EXPORT ${PROJECT_NAME} DESTINATION cmake)
# Adding formatting
cga_enable_auto_formatting("${CMAKE_CURRENT_SOURCE_DIR}")
```
here.

mimaric:
Yes, I'm also not sure about the best place to put this. The reason I created another library is that I already added a new file for `IndexDescriptor` as it is not a simple `using` but a full class, it even has a unittest folder

mimaric:
Sorry, hit "Close and comment" by accident

mimaric:
See https://github.com/mimaric/ClaraGenomicsAnalysis/tree/mimaric/better-types-in-fasta-parser/common/defs for `IndexDescriptor` in `defs`.
I'd like to keep it in a separate PR in order to be able to focus on project structure in this PR

ahehn-nv:
Is `IndexDescriptor` also used outside of `libcudamapper`?

mimaric:
For now no, but `FasraParser` has `get_chunks()` which essentially returns reads that belong to one index. On the other hand, one could argue that that method does not belong to `FastaParser` and should be internal to cudamapper

common/defs/CMakeLists.txt
mimaric: We decided not to introduce `defs` library. Marking this as resolved

mimaric:
@ahehn-nv and I agreed to not introduce `defs` library but to rename `cgautils` into `cgabase` and have `utils` be a subdirectory in it

common/base/include/claragenomics/types.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+namespace claragenomics
+{
+
+/// unique ID of a read
+using read_id_t = std::uint32_t;
+/// position of a basepair/kmer in a read
+using position_in_read_t = std::uint32_t;
+
+} // namespace claragenomics
```
vellamike, line 23: pedantic: missing newline

mimaric: Yup, will be changed in the next PR

tijyojwad:
discussed with @ahehn-nv and @mimaric - we'll move `config` and `logging` into `base` as well in another PR, and make `base` be the common dependence across modules. Approving this PR as the first step towards that.

PR #360, opened 2020-04-13
[cudamapper] Removed BioParser dependency

mimaric:
BioParser is not used anymore


mimaric:
I just realized that `3rdparty/spoa` uses BioParser.
@tijyojwad @vellamike can you check is that is still needed?
For now I'll only remove references to it from cudamapper

tijyojwad:
@mimaric `spoa` should have `bioparser` in it's own submodule list, and removing it from the CGA submodule list should not have any impact on what `spoa` uses internally. are you seeing otherwise?

we use `spoa` for some `cudapoa` MSA tests

mimaric:
I was some tests failing. I'll investigate a bit more

mimaric:
@tijyojwad After I remove BIoParser from `.gitmodules` and `cmake/3rdparty.cmake` the tests start failing. As far as I see there is no mention of BioParser in any directory outside of `3rdparty`, so I assume that spoa expects us to download it.
Can you please have a look?
I wouldn't mind keeping BioParser in the SDK, I just wanted to remove it from cudamapper as we don't use it there anymore

tijyojwad:
@mimaric - the `bioparser` used by `spoa` is listed under the `3rdparty/spoa/.gitmodules` file. It's the `.gitmodules` file that ships with the `spoa` repo itself.

I'll take a look to see what's going on. we don't use bioparser anywhere in CGA directly, so we should just remove it.

tijyojwad:
@mimaric I think the submodule was not removed correctly in your branch. I created #361 with instructions from https://gist.github.com/myusuf3/7f645819ded92bda6677 . Okay if I close this one?

mimaric:
I suggest we do it this way: I'll only remove references to BioParser from cudamapper in this PR. Once that is merged you can removed all other references to it in #361

tijyojwad:
Since `cudamapper` is the only module that used `bioparser`, I removed all references in #361 as well... Let's just merge that since tests have been kicked off already?

mimaric:
Sure, closing this PR, we'll continue on #361

PR #363, opened 2020-04-13
[common] Bugfix issue-346: unexpected behavior in cudautils::align

rilango:
The function was re-aligning the input to the next aligned value if the
input was already aligned. With this change an already aligned input is
not modified.


r-mafi:
the behavior of the function looks good after this fix, just gotta check why CI fails.

common/utils/include/claragenomics/utils/cudautils.hpp
```diff
@@ -96,7 +96,8 @@ __host__ __device__ __forceinline__
     align(const IntType& value)
 {
     static_assert((boundary & (boundary - 1)) == 0, "Boundary for align must be power of 2");
-    return (value + boundary) & ~(boundary - 1);
+    // value + mod(value -1) ^ boundary -1
+    return value + (((value - 1) & (boundary - 1)) ^ (boundary - 1));
```
tijyojwad, line 100: I think this can be simplified as `(value + boundary - 1) & ~(boundary - 1)`?

PR #364, opened 2020-04-14
[cudaaligner]  Banded Myers algorithm + misc improvements

ahehn-nv:
* Added Myers with Ukkonen banding
* Fixed bugs in min3
* Fixed illegal instruction error on compute_70


cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -312,44 +314,51 @@ myers_compute_scores(
...
         int32_t warp_carry = 0;
         if (threadIdx.x == 0)
             warp_carry = 1; // for global alignment the (implicit) first row has to be 0,1,2,3,... -> carry 1
-        for (int32_t idx = threadIdx.x; idx < n_words; idx += warp_size)
+        for (int32_t idx = threadIdx.x; idx < n_warp_iterations; idx += warp_size)
         {
-            const uint32_t warp_mask = idx / warp_size < n_words / warp_size ? 0xffff'ffffu : (1u << (n_words % warp_size)) - 1;
+            if (idx < n_words)
+            {
+                const uint32_t warp_mask = idx / warp_size < n_words / warp_size ? 0xffff'ffffu : (1u << (n_words % warp_size)) - 1;
```
tijyojwad, line 331: pedantic comment - safer to put `(...)` around `(idx / warp_size)` and `(n_words / warp_size)`?

cudaaligner/src/hirschberg_myers_gpu.cu
```diff
@@ -312,44 +314,51 @@ myers_compute_scores(
...
-                if (idx + 1 == n_words)
-                    score(t, reverse ? 1 : 0) = score(t - 1, reverse ? 1 : 0) + warp_carry;
+                warp_carry = myers_advance_block(warp_mask, highest_bit, eq, pv_local, mv_local, warp_carry);
+                if (full_score_matrix)
+                {
+                    score(idx, t) = score(idx, t - 1) + warp_carry;
+                }
+                else
+                {
+                    if (idx + 1 == n_words)
```
tijyojwad, line 347: Pedantic comment - I'm a bit surprised our style check doesn't change this to 
```
if (...)
{
...
}
```

safer to change this to have explicit scoping with `{}`?

cudaaligner/src/myers_gpu.cu
```diff
@@ -28,27 +28,51 @@ namespace claragenomics
 namespace cudaaligner
 {
 
+constexpr int32_t warp_size = 32;
```
tijyojwad, line 31: `CGA_CONSTEXPR` from here https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/89c411633de49dcdafe34492259d1d14722fe808/common/config/include/claragenomics/cga_config.hpp#L27 for compatibility with old CUDA versions

cudaaligner/src/aligner_global_myers_banded.cpp
```diff
@@ -0,0 +1,63 @@
...
+    : AlignerGlobal(max_query_length, max_target_length, max_alignments, stream, device_id)
+    , workspace_()
+{
+    scoped_device_switch dev(device_id);
+    workspace_ = std::make_unique<Workspace>(max_alignments, ceiling_divide<int32_t>(max_query_length, sizeof(myers::WordType)), max_target_length, stream);
+}
+
+AlignerGlobalMyersBanded::~AlignerGlobalMyersBanded()
+{
+    // Keep empty destructor to keep Workspace type incomplete in the .hpp file.
```
tijyojwad, line 48: My knowledge gap is showing here - can you explain this a bit more for me?

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: there are more instances of this on line 350 and 356

cudaaligner/src/myers_gpu.cu
ahehn-nv: `constexpr` variables should be ok. The problem are `constexpr` functions. The CI CUDA 9.0 passed, too.

cudaaligner/src/aligner_global_myers_banded.cpp
ahehn-nv: AlignerGlobalMyersBanded has a `unique_ptr<Workspace>`, which means the AlignerGlobalMyersBanded's dtor has to call delete on `Workspace`.
Now `Workspace` is defined in the .cpp file only. In the .hpp file, `Workspace` is only declared as "some" struct, so the compiler does not know the size of the Workspace object by looking at the .hpp file (nor does it know if it has to call destructors for data members). Therefore, it cannot delete it if it only has the .hpp file, but has to go to the .cpp file.

If I write `~AlignerGlobalMyersBanded = default;` in the .hpp file, it will generate a destructor inline (for all files that include the .hpp file), which will then try to delete Workspace inline. This fails for the reason mentioned above.

So the solution is to prevent the compiler from generating a destructor inline, but to explicitly tell the compiler there is a destructor in the .cpp file and it has to call that. In the .cpp file the compiler can then generate the delete Workspace, since it sees its definition.

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: To me operator precedence of `/` vs. `<` seems rather obvious - it's like in math.
Unless you feel strongly about it, I would like to leave it as it is.

cudaaligner/src/hirschberg_myers_gpu.cu
ahehn-nv: Old habits die hard. fixed.

cudaaligner/src/hirschberg_myers_gpu.cu
tijyojwad: Nah don't feel strongly enough. Just a personal preference. Resolving

cudaaligner/src/myers_gpu.cu
tijyojwad: Ah I see. Yes missed that the tests all passed

cudaaligner/src/aligner_global_myers_banded.cpp
tijyojwad: Cool, thanks for the explanation!

ahehn-nv:
rerun test

ahehn-nv:
rerun tests

ahehn-nv:
rerun tests

PR #367, opened 2020-04-14
[cudapoa] Batch size constructor

r-mafi:
a minor change to keep consistency with the old code. Adding +4, not sure if really needed though!


r-mafi:
these changes don't seem necessary. closing the PR.

PR #368, opened 2020-04-14
[pyclaragenomics] [cudapoa] Issus-356 - Update recent changes to python binding

rilango:
Refect BatchSize contructor change in python binding.


tijyojwad:
@rilango thanks for making these changes! in another PR we can make the changes for the python API to accept the other options exposed as part of the new `BatchSize` struct

PR #372, opened 2020-04-16
[pyclaragenomics] Respect CUDA_TOOLKIT_ROOT_DIR env variable in setup.py

ahehn-nv:


pyclaragenomics/setup.py
```diff
@@ -79,6 +79,10 @@ def copy_all_files_in_directory(src, dest, file_ext="*.so"):
 cga_root_dir = os.environ['CGA_ROOT_DIR']
 cga_install_dir = os.environ['CGA_INSTALL_DIR']
 cga_version = os.environ['CGA_VERSION']
+cuda_root = '/usr/local/cuda' if 'CUDA_TOOLKIT_ROOT_DIR' not in os.environ.keys() \
```
ohadmo, line 82: You could use instead :
`cuda_root = os.getenv('CUDA_TOOLKIT_ROOT_DIR', '/usr/local/cuda')`
which defaults to the second argument

ahehn-nv: cool, that looks much cleaner. Thanks!

ahehn-nv:
rerun tests

PR #370, opened 2020-04-15
[cudamapper] Grouping reads into indices moved in cudamapper

mimaric:
Grouping reads into indices was done in `FastaParser::get_read_chunks()`. The function is somewhat related to `FastaParser`, but it is cudamapper-specific and one could argue that is has more to with `Index` than `FastaParser`, so now it is a free function called `group_reads_into_indices()` in cudamapper.

Extracted `IndexDescriptor` into a separate file (together with `group_reads_into_indices()`). `IndexDescriptor` is only vaguely related to `IndexCache`.

Improved types in `FastaParser`.

Created tests for `group_reads_into_indices()`

Part of issue #318 

@tijyojwad FYI

PR #373, opened 2020-04-16
[cudaampper] main.cu uses namespace claragenomics::cudamapper

mimaric:
As discussed in PR #370 there is no reason not to have the content of `cudamapper/src/main.cu` in namespace `claragenomics::cudamapper::`

### DIFF
--- a/cudamapper/src/main.cu
+++ b/cudamapper/src/main.cu
@@ -32,6 +32,11 @@
 #include <claragenomics/cudaaligner/aligner.hpp>
 #include <claragenomics/cudaaligner/alignment.hpp>
 
+namespace claragenomics
+{
+namespace cudamapper
+{
+
 namespace
 {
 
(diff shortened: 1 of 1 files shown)


cudamapper/src/main.cu
```diff
@@ -775,3 +775,13 @@ int main(int argc, char* argv[])
 
     return 0;
 }
+
+} // namespace cudamapper
+} // namespace claragenomics
+
+/// \brief main function
+/// main function cannot be in a namespace so using this function to call actual main function
+int main(int argc, char* argv[])
```
ahehn-nv, line 784: Why not just leave main in the global namespace and do `using namespace claragenomcs::cudamapper;` inside?

ahehn-nv: discussed offline.

mimaric:
rerun tests

mimaric:
rerun tests

PR #374, opened 2020-04-16
[cudamapper] Index batching

mimaric:
Function groups indices into batches that are to be loaded into `IndexCacheHost` and `IndexCacheDevice`

Part of issue #318


mimaric:
rerun tests

cudamapper/src/index_batcher.cu
```diff
@@ -0,0 +1,134 @@
...
+* distribution of this software and related documentation without an express
+* license agreement from NVIDIA CORPORATION is strictly prohibited.
+*/
+
+#include "index_batcher.cuh"
+
+#include <algorithm>
+#include <exception>
+
+#include <claragenomics/utils/signed_integer_utils.hpp>
```
ahehn-nv, line 16: I think this include is not needed anymore.

cudamapper/tests/Test_CudamapperIndexBatcher.cu
```diff
@@ -0,0 +1,916 @@
...
+    number_of_indices_t target_indices_per_host_batch                   = 5;
+    number_of_indices_t target_indices_per_device_batch                 = 2;
+    std::shared_ptr<const claragenomics::io::FastaParser> target_parser = query_parser;
+    number_of_basepairs_t target_basepairs_per_index                    = 10;
+
+    const bool same_query_and_target = true;
+
+    // indices_per_host_batch different
+    target_indices_per_host_batch = 100;
+    ASSERT_THROW(const auto& generated_batches = generate_batches_of_indices(query_indices_per_host_batch,
```
ahehn-nv, line 860: This seems rather dangerous - binding a temporary to a `const&`. I would either not assign the return value at all or assign it to `auto generated_batches`.

ahehn-nv: These `ASSERT_THROW` checks are almost redundant at the moment.
If the last check works correctly, all other checks will pass too.
You need to start with identical query and target variables and then reassign only one for check test.

cudamapper/tests/Test_CudamapperIndexBatcher.cu
```diff
@@ -0,0 +1,916 @@
...
+        IndexBatch host_batch{std::move(host_batch_query_indices),
+                              std::move(host_batch_target_indices)};
+        std::vector<IndexBatch> device_batches;
+        {
+            // device query 0, device target 0
+            std::vector<IndexDescriptor> device_batch_query_indices{{0, 3}};
+            std::vector<IndexDescriptor> device_batch_target_indices{{0, 2}, {2, 1}};
+            device_batches.push_back({std::move(device_batch_query_indices),
+                                      std::move(device_batch_target_indices)});
+        }
```
ahehn-nv, line 186: Optional: This file is rather long, so I think it would increase readability if you just wrote
```
device_batches.push_back(
    {{0, 3}}, // query indicies
    {{0, 2}, {2, 1}} // target indices
);
```
instead of the recurring code blocks.
But since it is really just a minimal improvement and lots of typing to change. I probably wouldn't spend time on it. Just wanted to mention it.

cudamapper/src/index_batcher.cu
```diff
@@ -0,0 +1,134 @@
...
+                                                                                      query_indices_per_host_batch,
+                                                                                      target_indices_per_host_batch,
+                                                                                      same_query_and_target);
+
+    // create device batches for every host batch
+    std::vector<BatchOfIndices> all_batches;
+    for (IndexBatch& batch : host_batches)
+    {
+        // Device batches are only symmetric is their query and target indices are the same
+        const bool same_query_and_target_in_batch = same_query_and_target && std::equal(begin(batch.query_indices),
```
ahehn-nv, line 71: :+1:

cudamapper/tests/Test_CudamapperIndexBatcher.cu
```diff
@@ -0,0 +1,916 @@
...
+* NVIDIA CORPORATION and its licensors retain all intellectual property
+* and proprietary rights in and to this software, related documentation
+* and any modifications thereto.  Any use, reproduction, disclosure or
+* distribution of this software and related documentation without an express
+* license agreement from NVIDIA CORPORATION is strictly prohibited.
+*/
+
+#include "gtest/gtest.h"
+
+#include <exception>
```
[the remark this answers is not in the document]
mimaric: It's needed to cleanly use `std::invalid_argument`

cudamapper/tests/Test_CudamapperIndexBatcher.cu
```diff
@@ -0,0 +1,916 @@
...
+        const BatchOfIndices& expected_batch                                   = expected_batches[i];
+        const BatchOfIndices& generated_batch                                  = generated_batches[i];
+        const std::vector<IndexDescriptor>& expected_batch_host_query_indices  = expected_batch.host_batch.query_indices;
+        const std::vector<IndexDescriptor>& generated_batch_host_query_indices = generated_batch.host_batch.query_indices;
+        ASSERT_EQ(expected_batch_host_query_indices.size(), generated_batch_host_query_indices.size()) << "i: " << i;
+        for (std::size_t j = 0; j < expected_batch_host_query_indices.size(); ++j)
+        {
+            const IndexDescriptor& expected_host_query_index  = expected_batch_host_query_indices[j];
+            const IndexDescriptor& generated_host_query_index = generated_batch_host_query_indices[j];
+            ASSERT_EQ(expected_host_query_index.first_read(), generated_host_query_index.first_read()) << "i: " << i << ", j: " << j;
```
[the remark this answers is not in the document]
mimaric: I'm not sure what's better in this case. If it was a single call I'd say it's a better option, but as this is in a loop one might end up with millions of failed `EXPECT_EQ` which do not bring any new information.
On the other hand, I agree that this way one cannot see if next `ASSERT_EQ` will fail as well.
I'm ok with both options. @tijyojwad do you have an opinion on this?

cudamapper/tests/Test_CudamapperIndexBatcher.cu
mimaric: I agree about it not being necessary to assign the result at all. Done.

I only partially agree with the second comment. That assumes certain order of checks in the tests function.
Only could edit `generate_batches_of_indices()` so it first checks `indices_per_host_batch`  and then make a bug in check of `indices_per_device_batch` (or delete it completely) and the test would not catch that

cudamapper/src/index_batcher.cuh
```diff
@@ -0,0 +1,133 @@
...
+/// device_batches contains batches of indices that are parts of host_batch. These batches should be loaded into
+/// device memory one by one from host memory
+struct BatchOfIndices
+{
+    IndexBatch host_batch;
+    std::vector<IndexBatch> device_batches;
+};
+
+/// number_of_indices_t
+using number_of_indices_t = std::size_t;
```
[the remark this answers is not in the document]
mimaric: yes, let's start switching to it

cudamapper/tests/Test_CudamapperIndexBatcher.cu
```diff
@@ -0,0 +1,916 @@
...
+{
+namespace cudamapper
+{
+
+void test_generated_batches(const std::vector<BatchOfIndices>& generated_batches,
+                            const std::vector<BatchOfIndices>& expected_batches)
+{
+    ASSERT_EQ(expected_batches.size(), generated_batches.size());
+
+    for (std::size_t i = 0; i < expected_batches.size(); ++i)
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_batcher.cuh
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_batcher.cu
mimaric: done

cudamapper/tests/Test_CudamapperIndexBatcher.cu
mimaric: True, the reason I did it is to have something named, but those comments are as effective. Anyway, as you suggested, leaving it as it is

cudamapper/tests/Test_CudamapperIndexBatcher.cu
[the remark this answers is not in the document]
ahehn-nv: yes, right. good point...

cudamapper/tests/Test_CudamapperIndexBatcher.cu
ahehn-nv: Ouch... On my second comment: At first I thought we were talking about different things and wanted to explain again what I meant. Now, while trying that, I realized that you did exactly what I wanted to explain as fix. So please disregard my second comment - the way you do it is perfectly fine.
I got confused that all `target_*` variable got initialized separately and did not realize that they actually get initialized with the same value. Sorry.

Optional: Consider to initialize the `target_*` with their `query_*` counterpart, i.e.
```
number_of_indices_t target_indices_per_host_batch = query_indices_per_host_batch;
```
instead of initializing with the same literals.

mimaric: done

PR #375, opened 2020-04-16
[cmake] Combined mapper libraries to libcudamapper

ahehn-nv:


cudamapper/CMakeLists.txt
```diff
@@ -22,86 +22,51 @@ else()
...
-        src/overlapper.cpp)
-target_include_directories(overlapper_triggerred PUBLIC include ${CUB_DIR})
-target_link_libraries(overlapper_triggerred cgabase cgaio cgalogging cudaaligner)
-target_compile_options(overlapper_triggerred PRIVATE -Werror)
+target_include_directories(cudamapper
+    PUBLIC
+        $<INSTALL_INTERFACE:include>
+        $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
+)
+target_link_libraries(cudamapper cgabase cgaio cgalogging cub pthread)
```
mimaric, line 44: BTW, I realized we don't have to link `pthread` here or in `common/logging`. Can you please remove them as part of this PR?

ahehn-nv: I can see how `logging` does not need it, but `cudamapper`? We use `std::thread` in `cudamapper`, doesn't that require `pthread` anymore?

mimaric: I would assume now it links it automatically as it is part of standard...

ahehn-nv: I remember that it was required in the past, even though it was part of the standard library. Anyhow, that may have changed. removed.

PR #376, opened 2020-04-18
[cudamapper] Fix cmake failure.

rilango:
Fixes cmake error due to missing DESTINATION

### DIFF
--- a/cudamapper/CMakeLists.txt
+++ b/cudamapper/CMakeLists.txt
@@ -58,8 +58,9 @@ set_target_properties(cudamapper-bin PROPERTIES OUTPUT_NAME cudamapper)
 # Add tests folder
 add_subdirectory(tests)
 
-install(TARGETS cudamapper 
+install(TARGETS cudamapper
     EXPORT cudamapper
+    DESTINATION bin
 )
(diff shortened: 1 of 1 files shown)


mimaric:
Any idea why did this pass our CI?
On my machine CMake also fails without this change

rilango:
> Any idea why did this pass our CI?
> On my machine CMake also fails without this change

I have not yet investigated the root cause. I will investigate it.

mimaric:
Thank! For the record, this was broken in PR #375.

@ahehn-nv I see that you introduced `cudamapper-bin` in addition to `cudamapper` target. IIUC `cudamapper` was supposed to produce the library and `cudamapper-bin` the binary.
Knowing all this `DESTINATION bin` is probably not the best fix for this problem.

This is the error I was getting:
````
CMake Error at cudamapper/CMakeLists.txt:61 (install):
  install TARGETS given no ARCHIVE DESTINATION for static library target
  "cudamapper".

ahehn-nv:
Hm, I didn't not see this error. Anyhow, yes, it `cudamapper` is the library which should go to `lib`, while `cudamapper-bin` is the binary. #377 fixes this.

PR #379, opened 2020-04-20
[common] clear cuda error in cached allocator

tijyojwad:
This was causing an issue with cudaaligner benchmark where a subsequent run was failing because of the uncleared cuda error

### DIFF
--- a/common/base/include/claragenomics/utils/allocator.hpp
+++ b/common/base/include/claragenomics/utils/allocator.hpp
@@ -243,6 +243,8 @@ public:
         cudaError_t err = memory_resource_->DeviceAllocate(&ptr, n * sizeof(T), stream);
         if (err == cudaErrorMemoryAllocation)
         {
+            // Run cudaGetLastError() to clear the CUDA error before throwing the exception.
+            cudaGetLastError();
             throw device_memory_allocation_exception();
         }
         CGA_CU_CHECK_ERR(err);


mimaric:
Cached allocator actually does return CUDA API call's errors: https://github.com/NVlabs/cub/blob/c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/util_allocator.cuh#L519

@tijyojwad can you give some more details abut your use case? I don't really understand how do you manage to recover from such error anyway (trying to allocate a block which is as big as possible?)

ahehn-nv:
However, CUB also resets the cudaMalloc error internally, and only returns the error code:
https://github.com/NVlabs/cub/blob/c3cceac115c072fb63df1836ff46d8c60d9eb304/cub/util_allocator.cuh#L453

tijyojwad:
@mimaric that's exactly the use case. In the cudaaligner benchmark application we run with varying sizes. Anything that doesn't it in memory throws an exception which is caught, and the benchmarking moves on to the next set of inputs. Semantically to me it seemed that once the exception is caught I should be able to continue as before. However, what was happening was after the exception was caught, the next call to CGA_CU_CHECK_ERR(cudaPeekAtLastError()) was failing with 'out of memory' error code.

@ahehn-nv the cub code also runs several CUDA API calls after that error reset. Possible any of them persist? Also the error is being cleared only in the case when an exception is thrown, not in general

ahehn-nv:
Which allocator were you actually using? CUB's caching allocator or Milos' preallocating allocator?

tijyojwad:
@ahehn-nv I believe it's the CUB one, since I see `cga_enable_preallocating_caching_allocator:BOOL=OFF` and `cga_enable_caching_allocator:BOOL=ON` in my CMakeCache.txt

ahehn-nv:
Yes, that's CUB then. I'll have a closer look.

tijyojwad:
Closing this issue as CUB Cached Allocator will be removed from CGA.

PR #384, opened 2020-04-22
Close the loop on BatchSize in CUDAPoa libraries...

rilango:
A previous MR has some changes to BatchSize structure and associated
API. However, it did not contain change required to use this structure
in other APIs.

### DIFF
--- a/pyclaragenomics/claragenomics/bindings/cudapoa.pyx
+++ b/pyclaragenomics/claragenomics/bindings/cudapoa.pyx
@@ -65,7 +65,7 @@ cdef class CudaPoaBatch:
 
     def __cinit__(
             self,
-            max_seq_per_poa,
+            max_sequences_per_poa,
             max_gpu_mem,
             output_mask=cudapoa.consensus,
             device_id=0,
@@ -74,7 +74,7 @@ cdef class CudaPoaBatch:
             mismatch_score=-6,
             match_score=8,
             cuda_banded_alignment=False,
-            max_seq_size=1024,            
+            max_sequence_size=1024,
             max_concensus_size=None,
             max_nodes_per_window=None,
             max_nodes_per_window_banded=None,
@@ -84,15 +84,19 @@ cdef class CudaPoaBatch:
         partial order alignment across all windows in the batch.
 
         Args:
-            device_id : ID of GPU device to use
-            stream : CudaStream to use for GPU execution
+            max_sequences_per_poa : Maximum number of sequences per POA
             max_gpu_mem : Maximum GPU memory to use for this batch
             output_mask : Types of outputs to generate (consensus, msa)
-            batch_size : Structure encapsulating upper limits for POA batches
+            device_id : ID of GPU device to use
+            stream : CudaStream to use for GPU execution
             gap_score : Penalty for gaps
             mismatch_score : Penalty for mismatches
             match_score : Reward for match
             cuda_banded_alignment : Run POA using banded alignment
+            max_sequence_size : Maximum number of elements in a sequence
+            max_concensus_size : Maximum size of final consensus
+            max_nodes_per_window : Maximum number of nodes in a graph, 1 graph per window
+            max_nodes_per_window_banded : Maximum number of nodes in a graph, 1 graph per window in banded mode
         """
         cdef size_t st
         cdef _Stream temp_stream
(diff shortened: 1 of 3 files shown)


pyclaragenomics/claragenomics/bindings/cudapoa.pyx
```diff
@@ -65,11 +65,11 @@ cdef class CudaPoaBatch:
 
     def __cinit__(
             self,
-            max_sequences_per_poa,
-            gpu_mem,
+            max_mem,
+            output_mask,
+            batch_size,
```
tijyojwad, line 70: instead of exposing batch size as a struct, let's add the member variables as optional parameters to the API (similar to how gap_score and all the setup). so something like 

```
def __cinit__ (
self,
max_sequences_per_poa,
max_seq_size,
max_concensus_size,
max_mem,
output_mask,
....
max_nodes_per_poa,
)
```

and then send the `nodes_per_group` parameter to both banded and non banded paramter. I think in the C++ API as well we can keep just one and use the banded parameter to set the appropriate one internally - wdyt @r-mafi ?

rilango: Don't you think using BatchSize structure will make the binding immune to changes in C++ size (at-lease reduced ripple-effect)?

r-mafi: what I'm missing here is in C++, we have a constructor that assigns proper values to most of `BatchSize` members based on a minimal set of inputs (only `max_seq_sz` and `max_seq_per_poa`). How is that done here? Should user adjust `max_consensus_size` according to arbitrary `max_seq_size`? Are you suggesting to expose all `BatchSize` struct members or only a minimal set?

tijyojwad: I realized I had written that comment incorrectly. updated it.

@r-mafi - exposing only max seq ize and max seqs per poa as mandatory in python API. and accept the rest as option keyword args with default params. if default params are provided, then we use the 2 argument constructor for batch size, otherwise we use the 5 arg version. we expose all params, but make some optional and some mandatory. only thing is if some of the optional ones are provided, all the others need to be as well.

tijyojwad: @rilango - I personally don't like to expose structs as APIs since it's hard to track API changes as structs themselves change. in python the keyword args allows us to add new options without changing the mandatory args

r-mafi: I see, I agree with your comment regarding exposing structs in API.

rilango: Actually that was my point :) Avoid API method prototype if the underlying structure changes. But I get the point, python's keyword and default values this will also make batchsize completely optional. I will update a new version in a moment..

pyclaragenomics/claragenomics/bindings/cudapoa.pyx
```diff
@@ -99,23 +108,25 @@ cdef class CudaPoaBatch:
             st = stream.stream
             temp_stream = <_Stream>st
 
-        cdef int8_t output_mask
```
tijyojwad, line 102: I think this change needs to be reverted. the API exposed the output type to be a string - "consensus" or "msa". that's because we don't expose the cdef cudapoa.consensus type publicly

tijyojwad:
@rilango can you also check if the python samples run correctly? they're in `pyclaragenomics/samples/sample_cudapoa`

rilango:
I did not test it after the last change. let me check.

pyclaragenomics/samples/sample_cudapoa
```diff
@@ -38,11 +38,12 @@ def run_cudapoa(groups, msa, print_output):
...
     max_sequences_per_poa = 100
-    gpu_mem_per_batch = 0.9 * free
+    max_seq_sz = 1024
     batch = cudapoa.CudaPoaBatch(max_sequences_per_poa,
-                                 gpu_mem_per_batch,
+                                 max_seq_sz,
+                                 0.9 * free,
                                  stream=None,
-                                 output_type=("msa" if msa else "consensus"))
+                                 output_mask=msa)
```
tijyojwad, line 46: I think this line didn't need to be changed? It's still called `output_type` in the API and takes in a string, right? And either "msa" or "consensus" based on sample options?

pyclaragenomics/samples/sample_cudapoa
```diff
@@ -38,11 +38,12 @@ def run_cudapoa(groups, msa, print_output):
 
     # Create batch
     max_sequences_per_poa = 100
-    gpu_mem_per_batch = 0.9 * free
```
tijyojwad, line 41: let's keep the variable name around so it's easier for the user to understand what's being passed into the API

pyclaragenomics/test/test_cudapoa_bindings.py
```diff
@@ -20,7 +20,8 @@
 def test_cudapoa_simple_batch():
     device = cuda.cuda_get_device()
     free, total = cuda.cuda_get_mem_info(device)
-    batch = CudaPoaBatch(10, 0.9 * free, deivce_id=device)
+    batch = CudaPoaBatch(10, 1024, 0.9 * free, deivce_id=device,
```
tijyojwad, line 23: one last request @rilango  - let's add a version of the `test_cudapoa_simple_batch` that uses `banded=True`

tijyojwad: Thanks @rilango ! Looks great!

PR #395, opened 2020-04-23
[cudamapper] Improved IndexCacheHost and IndexCacheDevice

mimaric:
1) Both `IndexCacheHost` and `IndexCacheDevice` were only checking if the other cache (i.e. target for query and query for target) already had requested indices cached, but not if they themselves had it cached (which is more common). This is now implemented.

2) Excepted workflow is to instruct `IndexCacheHost` to generate some indices on device and copy them to host. Right after that `IndexCacheDevice` will request some of those indices from `IndexCacheHost` which will have to copy them back to device. In order to avoid this `IndexCacheHost` can be instructed to keep certain indices in device memory until they are requested for the first time (and also copy them to host). These indices should be the ones which will be requested by first `IndexCacheDevice`.

Part of issue #318


cudamapper/src/index_cache.cuh
```diff
@@ -67,10 +68,26 @@ public:
...
     /// \brief Discards previously cached query Indices, creates new Indices and copies them to host memory
-    void update_query_cache(const std::vector<IndexDescriptor>& descriptors_of_indices_to_cache);
+    ///
+    /// Expected usage pattern is to call update_query_cache() with certain index descriptors and then immediately retrieve some of them.
+    /// To avoid immediately copying back those indices from host to device it is possible to specify descriptors_of_indices_to_keep_on_device
+    //  which will be copied to host, but also kept on device until retrieved for the first time using get_index_from_query_cache
+    ///
+    /// \param descriptors_of_indices_to_cache descriptors on indices to keep in host memory
+    /// \param descriptors_of_indices_to_keep_on_device descriptors of indices to keep in device memory in addition to host memory until retrieved for the first time
+    void update_query_cache(const std::vector<IndexDescriptor>& descriptors_of_indices_to_cache,
```
ahehn-nv, line 78: I got a bit confused by the name `update_*_cache`. It took me a while to realize, that this does not just add elements to the cache, but actually generates the indices to be cached.
Would `generate_*_cache_content()` be a better name?

cudamapper/src/index_cache.cu
```diff
@@ -41,84 +41,164 @@ IndexCacheHost::IndexCacheHost(const bool reuse_data,
...
+                index_copy = existing_cache->second;
+                if (keep_on_device)
+                {
+                    index_on_device = index_copy->copy_index_to_device(allocator_, cuda_stream_);
+                }
+            }
+            else
+            {
+                // create index
+                index_on_device = claragenomics::cudamapper::Index::create_index(allocator_,
```
ahehn-nv, line 143: Optional: I would drop the `claragenomics::cudamapper` prefix.

cudamapper/src/index_cache.cu
```diff
@@ -41,84 +41,164 @@ IndexCacheHost::IndexCacheHost(const bool reuse_data,
...
+                                                                                 *parser,
+                                                                                 descriptor_of_index_to_cache.first_read(),
+                                                                                 descriptor_of_index_to_cache.first_read() + descriptor_of_index_to_cache.number_of_reads(),
+                                                                                 kmer_size_,
+                                                                                 window_size_,
+                                                                                 hash_representations_,
+                                                                                 filtering_parameter_,
+                                                                                 cuda_stream_);
+                // copy it to host memory
+                index_copy = claragenomics::cudamapper::IndexHostCopy::create_cache(*index_on_device,
```
ahehn-nv, line 153: Optional: I would drop the claragenomics::cudamapper prefix.

cudamapper/src/index_cache.cu
```diff
@@ -41,84 +41,164 @@ IndexCacheHost::IndexCacheHost(const bool reuse_data,
...
+                // index already cached
+                index_copy = existing_cache->second;
+                if (keep_on_device)
+                {
+                    index_on_device = index_copy->copy_index_to_device(allocator_, cuda_stream_);
+                }
+            }
+            else
+            {
+                // create index
```
ahehn-nv, line 142: Are duplicate entries in `descriptors_of_indices_to_cache` expected?
This function works for duplicate entries, but it will redo the index generation and then overwrite the previous cached index. If there are frequent duplicate entries, you might want to check the fresh cache, too.

cudamapper/src/index_cache.cuh
```diff
@@ -50,7 +51,7 @@ public:
     /// \param hash_representations // see Index
     /// \param filtering_parameter // see Index
     /// \param cuda_stream // device memory used for Index copy will only we freed up once all previously scheduled work on this stream has finished
-    IndexCacheHost(const bool reuse_data,
+    IndexCacheHost(const bool same_query_and_target,
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_cache.cuh
```diff
@@ -123,9 +161,9 @@ class IndexCacheDevice
 {
 public:
     /// \brief Constructor
-    /// \param reuse_data true means that both query and target are the same, meaning that if requested index exists in query cache it can also be used by target cache directly
+    /// \param same_query_and_target true means that both query and target are the same, meaning that if requested index exists in query cache it can also be used by target cache directly
     /// \param index_cache_host underlying host cache to get the indices from
-    IndexCacheDevice(const bool reuse_data,
+    IndexCacheDevice(const bool same_query_and_target,
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_cache.cuh
```diff
@@ -85,23 +102,44 @@ private:
...
+    /// If same_query_and_target_ is true function checks the other cache to see if that index is already in cache
     void update_cache(const std::vector<IndexDescriptor>& descriptors_of_indices_to_cache,
+                      const set_of_descriptors_t& descriptors_of_indices_to_keep_on_device,
                       const CacheToUpdate which_cache);
 
+    /// \brief Fetches requested index
+    /// Copies index from host to device memory, unless index is saved in temp device cache
+    /// If that is the case it returs that device copy are removes it from temp device cache
+    std::shared_ptr<Index> get_index_from_cache(const IndexDescriptor& descriptor_of_index_to_cache,
+                                                const CacheToUpdate which_cache);
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_cache.cuh
```diff
@@ -85,23 +102,44 @@ private:
...
                                             IndexDescriptorHash>;
 
+    using device_cache_type_t = std::unordered_map<IndexDescriptor,
+                                                   std::shared_ptr<Index>,
+                                                   IndexDescriptorHash>;
+
+    using set_of_descriptors_t = std::unordered_set<IndexDescriptor,
+                                                    IndexDescriptorHash>;
+
     enum class CacheToUpdate
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_cache.cu
mimaric: yes, that was a copy-paste error
done

cudamapper/src/index_cache.cu
mimaric: done

cudamapper/src/index_cache.cu
```diff
@@ -41,84 +41,164 @@ IndexCacheHost::IndexCacheHost(const bool reuse_data,
...
 {
-    // TODO: throw custom exception if index not found
-    return target_cache_.at(descriptor_of_index_to_cache)->copy_index_to_device(allocator_, cuda_stream_);
+    return get_index_from_cache(descriptor_of_index_to_cache,
+                                CacheToUpdate::target);
+}
+
+IndexCacheHost::set_of_descriptors_t IndexCacheHost::convert_vector_of_descriptors_into_set(const std::vector<IndexDescriptor>& descriptors)
+{
+    set_of_descriptors_t set_of_descriptors;
```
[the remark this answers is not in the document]
mimaric: good point, done and moved into `update_cache`

cudamapper/src/index_cache.cu
```diff
@@ -41,84 +41,164 @@ IndexCacheHost::IndexCacheHost(const bool reuse_data,
 {
 }
 
-void IndexCacheHost::update_query_cache(const std::vector<IndexDescriptor>& descriptors_of_indices_to_cache)
+void IndexCacheHost::update_query_cache(const std::vector<IndexDescriptor>& descriptors_of_indices_to_cache,
+                                        const std::vector<IndexDescriptor>& descriptors_of_indices_to_keep_on_device)
 {
-    update_cache(descriptors_of_indices_to_cache, CacheToUpdate::query);
+    update_cache(descriptors_of_indices_to_cache,
+                 convert_vector_of_descriptors_into_set(descriptors_of_indices_to_keep_on_device),
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/index_cache.cu
mimaric: No, every `IndexDescriptor` in `descriptors_of_indices_to_cache` is expected to be different.
I could add an assertion to check for that, but as the program would work correctly in that case as well I don't think it's necessary

cudamapper/src/index_cache.cuh
ahehn-nv: reads better now. thanks!

cudamapper/src/index_cache.cu
ahehn-nv: I agree, I don't think an assert is necessary here. Just wanted to check if this is a common case.

Issue #299, opened 2020-02-25
[pyclaragenomics] Increase the maximum sequence length limit


iiSeymour:
The 1,000 base limit with POA interface makes it quite restrictive for long-read applications and I've only been able to use it for proof of concept ideas. It would great to see this limit removed.

tijyojwad:
Hi @iiSeymour - agreed, and it's been on our plate for some time to make that configurable but sadly we haven't gotten around to it. We'll prioritize this effort and get it going.

tijyojwad:
@iiSeymour do you have a sense of how long your reads will be?

iiSeymour:
I'd say a practical increase would be up to 20Kbases but it's not uncommon for long reads to be much longer.

vellamike:
@tijyojwad we could consider doing this for the Python API, using windowing under the hood?

I don't know how SPOA handles these long reads, but I suspect the runtime is quadratic with read length. Is this right @rvaser ?

rvaser:
Yes, the time/space complexity inside SPOA is quadratic. But we could try and implemented the banded approach you guys have in the GPU version.

vellamike:
True, If we used banded would make it O(N) in space/time, but realistically we still want some form of stitching.

@rvaser happy to help with you guys on the banded alignment in SPOA. The method I used is not very complicated

rvaser:
I started fiddling with it this week, but it is a bit tiresome for the SIMD version. @vellamike, I will drop you an email tomorrow :)

iiSeymour:
Getting banded in SPOA would be great 👍

tijyojwad:
@iiSeymour just to keep you apprised of our developments, @r-mafi is adding changes in #327 to make sizes configurable so we can move to longer POAs.

iiSeymour:
Excellent, thanks @tijyojwad @r-mafi.

r-mafi:
@iiSeymour in case you have any long read data and can share them, it would help to verify the recent changes with the actual data. Thanks.

iiSeymour:
@r-mafi yes, I've followed up via email.

r-mafi:
> @r-mafi yes, I've followed up via email.

awesome, thank you!

tijyojwad:
[...]

PR #397, opened 2020-04-28
[cudapoa] changing variables related to dimension of POA arrays from uint16_t to template type

r-mafi:
many data structures defined in `cudapoa_kernels.cuh` associate with dimensions of score matrix, graph or reads sequence lengths. Dimension of cudapoa input was initially limited by `uint16_t` or `int16_t` types because of these data structures. Now those structures are templated and switch to 32-bit integer when needed.
Initial measurements indicated keeping data type to `int32_t` all the time has up to 20% degrading effect on the performance in `cudapoa_benchmarks`, therefore it has been changed to template type to switch to 32-bit only when needed.


tijyojwad:
This PR affects changes requested in #299

cudapoa/src/cudapoa_batch.hpp
```diff
@@ -156,8 +156,9 @@ class CudapoaBatch : public Batch
                                          num_nucleotides_copied_ * sizeof(uint8_t), cudaMemcpyHostToDevice, stream_));
         CGA_CU_CHECK_ERR(cudaMemcpyAsync(input_details_d_->window_details, input_details_h_->window_details,
                                          poa_count_ * sizeof(claragenomics::cudapoa::WindowDetails), cudaMemcpyHostToDevice, stream_));
+        /// ToDo may need to revise the following sizeof()
         CGA_CU_CHECK_ERR(cudaMemcpyAsync(input_details_d_->sequence_lengths, input_details_h_->sequence_lengths,
-                                         global_sequence_idx_ * sizeof(uint16_t), cudaMemcpyHostToDevice, stream_));
+                                         global_sequence_idx_ * sizeof(*input_details_h_->sequence_lengths), cudaMemcpyHostToDevice, stream_));
```
tijyojwad, line 161: I think this is the right way to do it! Using this for the rest of the `sizeof` calls in allocate_block.hpp would resolve the issue of types being out of sync right?

cudapoa/src/batch.cpp
```diff
@@ -30,31 +30,47 @@ std::unique_ptr<Batch> create_batch(int32_t device_id,
                                     bool cuda_banded_alignment)
 {
     // a decision flag to determine proper type definition for ScoreT
-    const bool use_32_bit_for_ScoreT = use32bitInt(batch_size, gap_score, mismatch_score, match_score);
+    const bool use_32_bit_for_ScoreT = use32bitScore(batch_size, gap_score, mismatch_score, match_score);
 
     if (use_32_bit_for_ScoreT)
```
tijyojwad, line 35: any benefit to calling a variable instead of the function directly (like being done for `use32bitSize`?

cudapoa/src/cudapoa_add_alignment.cu
```diff
@@ -273,91 +273,93 @@ __device__
...
-                                             base_weights,
-                                             sequence_begin_nodes_ids,
-                                             outgoing_edges_coverage,
-                                             outgoing_edges_coverage_count,
-                                             s,
-                                             max_sequences_per_poa,
-                                             max_limit_nodes_per_window);
-    *node_count        = new_node_count;
+    SizeT new_node_count;
+    /*uint8_t error_code =*/addAlignmentToGraph(new_node_count, nodes,
```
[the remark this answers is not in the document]
tijyojwad: and also let's remove the commented code

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -12,12 +12,11 @@
 
 #pragma once
 
-#include <claragenomics/cudapoa/cudapoa.hpp>
```
tijyojwad, line 15: doesn't this have the error types we need?

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -216,11 +216,11 @@ typedef struct GraphDetails
  */
 
 void generatePOA(claragenomics::cudapoa::OutputDetails* output_details_d,
```
tijyojwad, line 218: hmm I see that we're having to use `void` type for the structs since our implementation is in the `.cu` file, and if we make this templated then the whole thing would have to be included in the `.cuh`. do you want to give that a shot? Main reason I'm suggesting that is we won't have to cast types in `cudapoa_kernels.cu` if we make this function templated as well.

```
template<typename SizeT, typename ScoreT>
generatePOA(claragenomics::cudapoa::OutputDetails* output_details_d,
                      claragenomics::cudapoa::InputDetails<SizeT>* Input_details_d,
                      ...
)
```

what do you think?

tijyojwad: same for the other functions in this

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -12,12 +12,11 @@
 
 #pragma once
 
-#include <claragenomics/cudapoa/cudapoa.hpp>
-
 #include <stdint.h>
 #include <cuda_runtime_api.h>
 #include <stdio.h>
 #include <claragenomics/cudapoa/batch.hpp>
+#include "cudastructs.cuh"
```
tijyojwad, line 19: Is cudastructs.cuh needed in this file?

cudapoa/src/batch.cpp
r-mafi: no, initially I had added it with a short comment, but the comment is moved to function declaration. changed.

cudapoa/src/cudapoa_add_alignment.cu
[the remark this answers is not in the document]
r-mafi: done.

cudapoa/src/cudapoa_batch.hpp
r-mafi: yes, the only concern is for null pointers. but in `allocate_block.hpp` we know the pointer is not null.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: no `StatusType` in `cudapoa_kernels.cuh`, it's being used in the .cu file that includes `claragenomics/cudapoa/cudapoa.hpp`

cudapoa/src/cudapoa_kernels.cuh
r-mafi: no, not needed, removed.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: actually initially I had implemented it that way, but the code structurally got damaged, all those codes from `cudapoa_add_alignment.cu`, `cudapoa_nw.cu`,  `cudapoa_generate_consensus.cu`, `cudapoa_topsort` had to be moved under the same file in `cudapoa_kernels.cuh` leaving those files blank. I don't remember if I also faced circular dependency or not, but already did not like how the code looked like.  That's why I rewrote it back this way.

r-mafi:
@tijyojwad do you know what are the CI errors about? I could not relate them to the recent changes.

cudapoa/src/cudapoa_kernels.cuh
tijyojwad: Oh hmm... does it not work if each of those files are `#include <cudapoa_generate_consensus.cuh>` into the `cudapoa_kernels.cuh` file? so keeping the in different files but just including them into a single unit?

tijyojwad:
@r-mafi looks like a conda package install issue in the job, unrelated to your changes. Rerunning tests now to see if that fixes the issue

tijyojwad:
rerun tests

cudapoa/src/cudapoa_kernels.cuh
r-mafi: sorry, my comment was not clear enough. There are multiple functions in `cuda_kernels.cuh` with void pointers.  Namely: `generatePOA()`, `addAlignment()`, `runNW()`, `generateConsensusTestHost()` and `runTopSort()`. Other places where any of these functions are called, include `cuda_kernels.cuh`.
Now by making these functions template type, we need to either bring function definition (and all their dependent functions) into the same `cuda_kernels.cuh` file which as mentioned in my previous comment it won't have a nice structure. If we want to keep dependencies in their own file, e.g. we define  `addAlignment()` inside `cuda_kernels.cuh` but keep all the other functions defined in `cudapoa_add_alignments.cu` (i.e `addAlignmentKernel()` and `addAlignmentToGraph()` ) where they are, then there will be circular dependency, i.e. in `cudapoa_add_alignment.cu` we have to include 'cudapoa_kernels.cuh' and vice-versa.

or alternatively we could move each of these function declarations out of `cuda_kernels.cuh` and keep it where it's being defined. Then there will be 2 problems: 1- caller may require to decide what template type to use (something like what is done inside `generatePOA()` by calling `use32bitScore()`) 2- we don't have a single .cuh file to keep all POA kernels under the same umbrella.

PR #400, opened 2020-04-29
[cudamapper] Removed info about reads from Index

mimaric:
`Index` contained functions for converting `read_id` into read name and read length. This information if already available in `FastaParser` and is only used on host side when adding read names to `Overlap`s.

This also meant that a copy of `Index` had to be kept until `Overlap`s had been written to output, which is done on a separate host-only thread. This resulted in keeping two sets of indices: the one which is currently being processed and the one which has already been processed and is being written to output, which both increased memory consumption and lead to memory fragmentation.

The following functions have been removed, as well as underlying data structures:
`Index::read_id_to_read_name()`
`Index::read_id_to_read_length()`
`Index::read_ids_to_read_names()`
`Index::read_ids_to_read_lengths()`
`IndexHostCopyBase::read_id_to_read_names()`
`IndexHostCopyBase::read_id_to_read_lengths()`

In addition to this `FastaParser::get_sequence_by_id()` has been modified to return by reference, not by value.

All this allowed `Overlapper::update_read_names()` to accept `FastaParser` instead of `Index` and thus do not require keeping device data which it anyway did not use.

Note that `update_read_name()` probably belongs more to `Overlap` than to `Overlapper`, but this is something that will be refactored in another PR.

This PR is part of Issue #318

### DIFF
--- a/common/io/include/claragenomics/io/fasta_parser.hpp
+++ b/common/io/include/claragenomics/io/fasta_parser.hpp
@@ -43,8 +43,8 @@ public:
 
     /// \brief Fetch an entry from the FASTA file by index position in file.
     /// \param sequence_id Position of sequence in file. If sequence_id is invalid an error is thrown.
-    /// \return A FastaSequence object describing the entry.
-    virtual FastaSequence get_sequence_by_id(read_id_t sequence_id) const = 0;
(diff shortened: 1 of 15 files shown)


tijyojwad:
LGTM, merging!

PR #401, opened 2020-04-29
[pycga] update cudapoa sample to use long read data

tijyojwad:
Update python sample app to showcase long read POA using banded alignment. Also improve error handling and reporting in sample.


tijyojwad:
FYI @rilango - wanted to get you more involved with the python samples and how it's used as you start thinking about bindings more generally

pyclaragenomics/samples/sample_cudapoa
```diff
@@ -35,41 +35,67 @@ def run_cudapoa(groups, msa, print_output):
...
+        # If group was added and more space is left in batch, continue onto next group.
+        if (group_status == 0):
+            for seq_index, status in enumerate(seq_status):
+                if status != 0:
+                    print("Could not add sequence {} to POA {} - error {}".format(seq_index, poa_index, cudapoa.status_to_str(status)))
+            poa_index += 1
+
+        # Otherwise print error.
+        # Once batch is full or no groups are left, run POA processing.
+        if ((group_status == 1) or (poa_index == len(groups) - 1)):
```
r-mafi, line 77: do we need to consider a scenario that a single POA group is too large to be added? e.g. in bonito the first window consists of reads with length near 14K. Using full-alignment on a GPU with small memory it can run out of memory. Then will it get stuck in an infinite while loop? this can be tested by changing `gpu_mem_per_batch = 0.9 * free` to something smaller than 2GB.

pyclaragenomics/samples/sample_cudapoa
```diff
@@ -35,41 +35,67 @@ def run_cudapoa(groups, msa, print_output):
...
+    # Calculate max bounds for sequence size and sequences per POA.
+    max_sequences_per_poa = 0
+    max_seq_sz = 0
+    for group in groups:
+        longest_seq = len(max(group, key=len))
+        max_seq_sz = longest_seq if longest_seq > max_seq_sz else max_seq_sz
+        seq_in_poa = len(group)
+        max_sequences_per_poa = seq_in_poa if seq_in_poa > max_sequences_per_poa else max_sequences_per_poa
+
+    max_sequences_per_poa += 1
```
r-mafi, line 49: for `max_sequences_per_poa` I guess there is no need for +1. For example in bonito dataset, using 6 is good enough. But for `max_seq_sz` it is definitely needed.

tijyojwad: Yeah you're right. There's a semantic difference here in the way the two maxes are checked (one is inclusive and the other is not). Let me make the consistent in this PR.

pyclaragenomics/samples/sample_cudapoa
tijyojwad: Good catch! In this implementation it happened only when the last group was not addable to the batch, so it got stuck in an infinite loop.

Also found a deeper bug in the binding that was always enabling banded for cudapoa. Fixed that as well.

pyclaragenomics/samples/sample_cudapoa
tijyojwad: It looks like both the max sequence size and max sequences per poa are subject to the same `>=` condition, so I think the `+1` is needed for both cases

https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/src/cudapoa_batch.hpp#L488

r-mafi: A few lines after the comparison for `num_seq`, it gets updated after the comparison: https://github.com/clara-genomics/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/src/cudapoa_batch.hpp#L506
That's why for `max_sequences_per_poa` +1 is not needed and for the other one it is.
If you change them to be consistent, I think it's better to change them both not to need +1.

tijyojwad: Cool, thanks for pointing that one out! I've updated the code to make `max_seq_sz` and `max_seqs_per_poa` consistent in upper limit behavior

r-mafi: great, thanks.

Issue #386, opened 2020-04-22
[pyclaragenomics] upload pyclaragenomics API documentation


tijyojwad:
publish pyclaragenomics API documentation to public location

tijyojwad:
WIP Python API documentation uploaded to https://clara-genomics.github.io/ClaraGenomicsAnalysis/latest_dev/python/index.html

tijyojwad:
Done, and available on https://clara-parabricks.github.io/GenomeWorks/

PR #413, opened 2020-05-05
[doc] fix python documentation generation

tijyojwad:
Python documentation generation was semi-broken because of a missing sphinx option called "implicit-namespaces" after this change was introduced earlier.

Related to #386


tijyojwad:
@ohadmo remove doc generation from style check because without bindings bring generated doc generation was broken anyway (and the command still passed successfully!). I put the doc generation contents into a script and filing another task to properly enable python doc generation test.

PR #417, opened 2020-05-06
[CI] Fix PyPI upload command

ohadmo:


ci/release/pypi_uploader.sh
```diff
@@ -31,13 +31,14 @@ for f in ${WORKSPACE}/pyclaragenomics/pyclaragenomics_wheel/*.whl; do
         exit 1
     else
         conda install -c conda-forge twine
+        pip install 'readme-renderer>=21.0' # to support py3.5 images
```
tijyojwad, line 34: let's make this `python3 -m pip install ...` to be on the safer side and consistent with the upload command

PR #418, opened 2020-05-06
[cudapoa] int32_t bug fixes

r-mafi:
Through the conversion of some cudapoa data structure member types from `uint16_t` or `int16_t` to template type `SizeT`, some changes were missed. For example some loop counters which had to be changed along with the template types.
Also fixed overflow in computing index for score matrix in full-NW kernel. Banded NW is left unchanged as considering a bandwidth of 1000 (it is currently ~128), sequences should be larger than 2M to cause overflow.
Additionally removed `IndexT` template type in NW related kernels, as with presence of SizeT, it was redundant.


cudapoa/src/cudapoa_nw.cu
```diff
@@ -438,7 +445,9 @@ __device__
             j = prev_j;
 
         } // end of while
-        if (loop_count >= (read_count + graph_count + 2))
+
+        /// ToDo replace UINT16_MAX
+        if (loop_count >= (read_length + graph_count + 2))
         {
             aligned_nodes = UINT16_MAX;
```
r-mafi, line 452: I guess `-1` would be a better option to return for error code.

PR #420, opened 2020-05-08
[cuapoa] bug fixes related to changing type to int32_t

r-mafi:
- Through the conversion of some cudapoa data structure member types from uint16_t or int16_t to template type SizeT, some changes were missed. For example some loop counters which had to be changed along with the template types.

- Also fixed overflow in computing index for score matrix in full-NW kernel. Banded NW is left unchanged as considering a bandwidth of 1000 (it is currently ~128), sequences should be larger than 2M to cause overflow.

- Additionally removed IndexT template type in NW related kernels, as with presence of SizeT, it was redundant.

- Fixed a bug caused by initializing minimum score value to `SHRT_INT`, replaces by `std::numeric_limits(ScoreT)`
- Changed return value forfailed alignment from `UINT16_MAX` to `-1`


cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -21,34 +22,35 @@ namespace claragenomics
...
-__device__ uint16_t get_band_start_for_row(uint16_t row_idx, float gradient, uint16_t band_width, uint16_t max_column)
+template <typename SizeT>
+__device__ SizeT get_band_start_for_row(SizeT row_idx, float gradient, SizeT band_width, SizeT max_column)
 {
 
-    int16_t start_pos = uint16_t(row_idx * gradient) - band_width / 2;
+    SizeT start_pos = SizeT(row_idx * gradient) - band_width / 2;
 
-    start_pos = max(start_pos, 0);
+    start_pos = max(start_pos, (SizeT)0);
```
tijyojwad, line 31: `static_cast`

cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -21,34 +22,35 @@ namespace claragenomics
...
-    int16_t end_pos = start_pos + band_width;
+    SizeT end_pos = start_pos + band_width;
 
     if (end_pos > max_column)
     {
         start_pos = max_column - band_width + CELLS_PER_THREAD;
     };
 
-    start_pos = max(start_pos, 0);
+    start_pos = max(start_pos, (SizeT)0);
```
tijyojwad, line 40: `static_cast`

cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -80,49 +82,49 @@ __device__ void set_score(ScoreT* scores, uint16_t row, uint16_t column, ScoreT
...
 {
-    int32_t lane_idx    = threadIdx.x % WARP_SIZE;
-    uint16_t band_start = get_band_start_for_row(row, gradient, band_width, max_column);
-    uint16_t band_end   = band_start + band_width;
+    int16_t lane_idx = threadIdx.x % WARP_SIZE;
+    SizeT band_start = get_band_start_for_row(row, gradient, band_width, max_column);
+    SizeT band_end   = band_start + band_width;
 
-    uint16_t initialization_offset = (band_start == 0) ? 1 : band_start;
+    SizeT initialization_offset = (band_start == 0) ? (SizeT)1 : band_start;
```
tijyojwad, line 92: static_cast

cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -211,29 +215,29 @@ __device__
...
+                ScoreT penalty = score_type_min_limit;
                 for (uint16_t p = 0; p < pred_count; p++)
                 {
                     SizeT pred_node_id        = incoming_edges[node_id * CUDAPOA_MAX_NODE_EDGES + p];
                     SizeT pred_node_graph_pos = node_id_to_pos[pred_node_id] + 1;
-                    penalty                   = max(penalty, get_score(scores, pred_node_graph_pos, 0, gradient, band_width, read_length + 1, min_score_abs));
+                    penalty                   = max(penalty, get_score(scores, pred_node_graph_pos, (SizeT)0, gradient, band_width, static_cast<SizeT>(read_length + 1), min_score_value));
                 }
-                set_score(scores, i, 0, (ScoreT)(penalty + gap_score), gradient, band_width, max_column);
+                set_score(scores, i, (SizeT)0, static_cast<ScoreT>(penalty + gap_score), gradient, band_width, max_column);
```
tijyojwad, line 240: generally seeing lots of `(SizeT)` uses here (all preceding your changes). can you standardize those to static_cast :)?

r-mafi: done. I was under impression that it is to define the type by compiler and no casting is involved. Regardless all such cases are now changed to `static_cast<>`

cudapoa/src/cudapoa_nw_banded.cu
r-mafi: done

cudapoa/src/cudapoa_nw_banded.cu
r-mafi: done

cudapoa/src/cudapoa_nw_banded.cu
r-mafi: fixed

cudapoa/src/cudapoa_nw_banded.cu
tijyojwad: Thanks @r-mafi ! I realized my earlier statement might have been confusing. All of those old type casts were "pre-dating" your changes, not preceding hehe. So none of this was from your changes, but I'm glad you were able to fix those in this PR too :)!

raydouglass:
rerun tests

r-mafi:
rerun tests

PR #422, opened 2020-05-08
[cudamapper] Overlap end rescue via similarity estimation

edawson:
This adds a kmer-based similarity calculation function (and helper functions) to cudamapper_utils. This will eventually get integrated into a process for extending overlaps at the flanks of sequences.


cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,58 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+        }
+        else
+        {
+            ++b_index;
+        }
+    }
+    return shared_count;
+}
+
+float similarity(std::string a, std::string b, int kmer_length, int stride)
```
vellamike, line 113: pass by reference?

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,29 @@ namespace cudamapper
...
+/// \brief Given two sequences 'a' and 'b', calculate an estimate of their similarity
+/// Calculates the approximate nucleotide identity (or "similarity")
+/// estimated from the Jaccard index of the kmers of strings a and b.
+/// Note: This function assumes that a and b are on the same strand; you may need to
+/// reverse-complement one of the strings if testing similarity on strings from different
+/// strands.
+/// \param a A C++ string
+/// \param b A C++ string
+/// \param kmer_length The kmer length to use for estimating similarity.
+/// \param stride The number of bases to stride between kmers.
```
vellamike, line 75: needs a `\return` documented

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,29 @@ namespace cudamapper
 /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
 void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overlap>& unfused_overlaps);
 
+/// \brief Given a string s, produce its kmers (length <kmer-length>) and return them as a vector of strings.
+/// \param s A string sequence to kmerize.
+/// \param kmer_length A kmer length to use for producing kmers.
+/// \param stride The number of bases to skip when selecting kmers (most often, this should be equal to 1).
```
vellamike, line 57: needs `\return` documented

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,58 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+    std::vector<std::string> a_kmers = kmerize_string(a, kmer_length, stride);
+    std::vector<std::string> b_kmers = kmerize_string(b, kmer_length, stride);
+    std::sort(a_kmers.begin(), a_kmers.end());
+    std::sort(b_kmers.begin(), b_kmers.end());
+
+    std::size_t shared_kmers = count_shared_elements(a_kmers, b_kmers);
+    // Calculate "containment", i.e., the total number of shared elements divided by
+    // the number of elements in the smallest set. Min: 0, Max: 1.
+    std::size_t shortest_kmer_set_length = std::min(a_kmers.size(), b_kmers.size());
+    return static_cast<float>(shared_kmers) / static_cast<float>(shortest_kmer_set_length);
```
vellamike, line 124: I don't think this is calculating the Jaccard index (as per the documentation).

Say k=1, stride =1, string1 = 'ACGTA', string2 = 'CGTA'

`shared kmers` then = 4 (C,G,T,A) and `shortest_kmer_set_length` =  1 giving similarity = 1.

However, we want the intersection / union of the two sets which is = 4/5 = 0.8

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,58 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
     set_relative_strand(fused_overlap);
     fused_overlaps.push_back(fused_overlap);
 }
+
+std::vector<std::string> kmerize_string(const std::string& s, int kmer_length, int stride)
```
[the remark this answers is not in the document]
edawson: I agree - uint8_t would probably do, but should I go ahead and make it a uint32_t?

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,31 @@ namespace cudamapper
...
+/// Calculates the approximate nucleotide identity (or "similarity")
+/// estimated from the Jaccard index of the kmers of strings a and b.
+/// Note: This function assumes that a and b are on the same strand; you may need to
+/// reverse-complement one of the strings if testing similarity on strings from different
+/// strands.
+/// \param a A C++ string
+/// \param b A C++ string
+/// \param kmer_length The kmer length to use for estimating similarity.
+/// \param stride The number of bases to stride between kmers.
+float similarity(std::string a, std::string b, int kmer_length, int stride);
```
mimaric, line 76: Function name could be a bit more descriptive

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,63 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
     set_relative_strand(fused_overlap);
     fused_overlaps.push_back(fused_overlap);
 }
+
+std::string string_slice(const std::string& s, std::size_t start, std::size_t end)
```
mimaric, line 70: Ok, now I see that this is a "local" function, i.e. it is not exposed in header. In that case you can move it to an anonymous namespace so it does not get exposed to linker

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,63 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+{
+    return s.substr(start, end - start);
+}
+
+std::vector<std::string> kmerize_string(const std::string& s, int kmer_length, int stride)
```
mimaric, line 75: Again, you could use `string_view` in order to avoid dynamic memory allocations and copies for all these strings

cudamapper/src/cudamapper_utils.cpp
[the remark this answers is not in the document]
mimaric: @ahehn-nv and I discussed this a few time and decided to go for singed types. I don't think we changed it everywhere already, but new new codes should definitely use it. So in this case I'd use `std::int32_t`. I think there are still some places where we keep them 64-bit. if you get compiler warnings simply stick with 64-bit.
I'm planning to add kmer and window size types to `types.hpp` to avoid this type of confusion

cudamapper/src/cudamapper_utils.cpp
mimaric: Also use `const` for pass by value in function definitions, as discussed above

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,63 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+        }
+    }
+    return shared_count;
+}
+
+float similarity(std::string a, std::string b, int kmer_length, int stride)
+{
+    std::vector<std::string> a_kmers = kmerize_string(a, kmer_length, stride);
+    std::vector<std::string> b_kmers = kmerize_string(b, kmer_length, stride);
+    std::sort(a_kmers.begin(), a_kmers.end());
```
mimaric, line 117: We prefer using `begin(a_kmers)` over `a_kmers,begin()`

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,63 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+    std::sort(b_kmers.begin(), b_kmers.end());
+
+    std::size_t shared_kmers = count_shared_elements(a_kmers, b_kmers);
```
mimaric, line 120: Can `shared_kmers` be const?

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,29 @@ namespace cudamapper
 /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
 void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overlap>& unfused_overlaps);
 
+/// \brief Given a string s, produce its kmers (length <kmer-length>) and return them as a vector of strings.
+/// \param s A string sequence to kmerize.
+/// \param kmer_length A kmer length to use for producing kmers.
+/// \param stride The number of bases to skip when selecting kmers (most often, this should be equal to 1).
+std::vector<std::string> kmerize_string(const std::string& s, int kmer_length, int stride);
+
+/// \brief Given two sorted vectors of comparable types, return a size_t count of the number of shared elements.
```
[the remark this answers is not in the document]
mimaric: @vellamike It says `Given two sorted vectors`, but yes, I would also specify it in `\param` (see my comment in function definition for an idea how to avoid this constraint)

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,31 @@ namespace cudamapper
...
 
+std::string string_slice(const std::string& s, std::size_t start, std::size_t end);
+
+/// \brief Given a string s, produce its kmers (length <kmer-length>) and return them as a vector of strings.
+/// \param s A string sequence to kmerize.
+/// \param kmer_length A kmer length to use for producing kmers.
+/// \param stride The number of bases to skip when selecting kmers (most often, this should be equal to 1).
+std::vector<std::string> kmerize_string(const std::string& s, int kmer_length, int stride);
+
+/// \brief Given two sorted vectors of comparable types, return a size_t count of the number of shared elements.
```
mimaric, line 60: Let's say that vector `a` has `m` elements with value `x` and vector `b` `n` such elements. Should this results in `1` or `m*n` shared elements? I would assume former but your implementation would actually return `min(m, n)`

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -65,5 +66,63 @@ void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overl
...
+}
+
+template <typename T>
+std::size_t count_shared_elements(const std::vector<T>& a, const std::vector<T>& b)
+{
+    std::size_t a_index      = 0;
+    std::size_t b_index      = 0;
+    std::size_t shared_count = 0;
+
+    while (a_index < a.size() && b_index < b.size())
```
mimaric, line 93: This method is O(n), but requires sorted vectors which is O(n*log(n)).

To repeat my question from the header file:
```Let's say that vector a has m elements with value x and vector b n such elements. Should this results in 1 or m*n shared elements? I would assume former but your implementation would actually return min(m, n)```

Depending on the answer to that question you could use `std::unordered_set` or `std::unordered_map` with a counter as value to to this on unsorted vectors

File: cudamapper/src/overlapper.cpp [hunk 14]
```diff
@@ -203,5 +204,39 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps)
         overlaps.push_back(fused_overlap);
     }
 }
+
+void Overlapper::rescue_overlap_ends(std::vector<Overlap>& overlaps,
+                                     const io::FastaParser& query_parser,
+                                     const io::FastaParser& target_parser,
+                                     int max_extension,
```
mimaric, line 211: As discussed before: use types with specified number of bits, use `const` for pass by value parameters in function definitions where appropriate

cudamapper/src/overlapper.cpp
```diff
@@ -203,5 +204,39 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps)
...
+{
+
+    // Loop over all overlaps
+    // For each overlap, retrieve the read sequence and
+    // check the similarity of the overlapping head and tail sections (matched for length)
+    // If they are more than or equal to <required_similarity> similar, extend the overlap start/end fields by <max_extension> basepairs.
+
+    for (auto& overlap : overlaps)
+    {
+        std::string query_sequence  = query_parser.get_sequence_by_id(overlap.query_read_id_).seq;
```
mimaric, line 222: `const std::string&`?

cudamapper/src/overlapper.cpp
```diff
@@ -203,5 +204,39 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps)
...
+        std::string target_sequence = target_parser.get_sequence_by_id(overlap.target_read_id_).seq;
+        // TODO: reverse complement target if match is on '-' relative strand.
+
+        // overlap rescue at "head" (i.e., "left-side") of overlap
+        std::size_t query_rescue_head_start  = std::min(static_cast<int32_t>(0), static_cast<int32_t>(overlap.query_start_position_in_read_) - max_extension);
```
mimaric, line 227: @ahehn-nv Here's another example where safe `cast_to_signed()` could be useful.

It could be implemented similarly to https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/dba8fe0cedc30d181d8057cf2d7940830cae1f89/common/base/include/claragenomics/utils/signed_integer_utils.hpp#L22

ahehn-nv: First, concerning this function:
I think here is a bug. I haven't looked at the PR in detail, but I guess the idea was to make sure that `query_rescure_head_start` is greater-or-equal than zero. It actually makes sure that the value is either zero or negative (which will then become a huge unsigned value). So it should be `std::max`, right?

Concerning the `cast_to_signed`, I can see how this would help here, but I actually think that the problem is further down...
The example shows well, why I would like to avoid unsigned variables basically everywhere and use signed integers for everything that represents a number (including indices and sizes).
Fortunately, Eric was aware that `overlap.query_start_position_in_read_` was an unsigned variable and that he had to cast it to signed first in order for
```
max(0, position - max_extension);
```
to work. If he had been a little less careful, this would have been a pretty bad bug.

So, @mimaric @vellamike: Could we make `position_in_read_t` = `std::int32_t;` (i.e. `overlap.query_start_position_in_read_`) instead of `uint32_t`? Or do we expect reads of a length between 2x10^9 and 4x10^9 bps?

Then `query_rescue_head_start` could be of type `position_in_read_t` as well. Correct?

edawson: > the idea was to make sure that query_rescure_head_start is greater-or-equal than zero

That's correct, and this was indeed a bug (currently fixing).

> Fortunately, @edawson was aware that overlap.query_start_position_in_read_ was an unsigned variable

I have to credit the compiler here for knowing they were incompatible types. I don't think we'd see a position in the 2-4x10^9 (i.e., 2-4 gigabase) range unless we were doing genome-genome overlapping on a very large genome. I think it's fine to have unsigned positions but I'd like to add a note in the contributing docs reflecting that all position values are unsigned if so.

cudamapper/src/cudamapper_utils.cpp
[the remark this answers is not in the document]
edawson: resolved by 100bb80

cudamapper/src/cudamapper_utils.hpp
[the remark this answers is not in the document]
edawson: Fixed in 3b805e8e.

cudamapper/src/cudamapper_utils.cpp
edawson: Resolved in 100bb80

cudamapper/src/cudamapper_utils.hpp
edawson: Resolved in 100bb80.

cudamapper/src/cudamapper_utils.hpp
edawson: Resolved in 100bb80

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -73,6 +73,12 @@ class Overlapper
     /// \brief Identified overlaps which can be combined into a larger overlap and add them to the input vector
     /// \param overlaps reference to vector of Overlaps. New overlaps (result of fusing) are added to this vector
     static void post_process_overlaps(std::vector<Overlap>& overlaps);
+
+    static void rescue_overlap_ends(std::vector<Overlap>& overlaps,
```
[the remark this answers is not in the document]
edawson: Added in 8bb2aee

cudamapper/src/overlapper.cpp
mimaric: I'm ok with turning everything to signed and casting to unsigned if needed (e.g. getting container size). We'll just have to create a PR for that...

edawson:
Here's an example where the initial PAF (without end rescue) misses 24 overlapping bases at the end of a read
```
E_coli_test_100kb_95393_96393_1000	1000	1	636	+	E_coli_test_100kb_95053_96053_1000	1000	341	976	1170	635	255
```
[read1.fa.txt](https://github.com/clara-parabricks/ClaraGenomicsAnalysis/files/4617708/read1.fa.txt)
[read2.fa.txt](https://github.com/clara-parabricks/ClaraGenomicsAnalysis/files/4617709/read2.fa.txt)

[aln_sim_perfect_e_coli_reads.txt](https://github.com/clara-parabricks/ClaraGenomicsAnalysis/files/4617690/aln_sim_perfect_e_coli_reads.txt)

The end-rescued PAF then becomes (extension: 50bp; required_similarity: 0.5, which is likely a bit too low):
```
E_coli_test_100kb_95393_96393_1000	1000	1	660	+	E_coli_test_100kb_95053_96053_1000	1000	341	1000	1170	659	255
```

vellamike:
@edawson  I think the scope of the PR has been changed to include performing the overlap end rescue, can you detail this in the PR description?

vellamike:
Why is the start 1 not 0. Are PAF start and ends 1-based?

vellamike:
For the example with perfect reads, why is the required similarity 0.5? It should be >0.99 if the reads are identical?

edawson:
> Why is the start 1 not 0 (in your example of the 24-base extension). Are PAF start and ends 1-based?

PAF is 0-based, but here we don't attempt to extend the overlap because the 1-base portion is less than the kmer length (default of fifteen). We could add logic handle ends like this.

cudamapper/src/cudamapper_utils.cpp
edawson: I've now addressed this in 8bdeed851, but more broadly I think we should use the Jaccard containment, rather than the Jaccard similarity index, or change the bounds used for similarity estimation to always take at most the length of the shorter subsequence. Given that I already broke the scope of this PR I'm going to save that but it's worth a discussion.

CMakeLists.txt
```diff
@@ -88,6 +88,6 @@ endif()
 # Add auto formatting.
 cga_enable_formatting_targets()
 
-if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT  AND (${CMAKE_SOURCE_DIR} STREQUAL ${CMAKE_CURRENT_SOURCE_DIR}))
```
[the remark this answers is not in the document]
edawson: Removed in 8c1bab9. I think this was a merge artifact on my end.

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -9,6 +9,7 @@
 */
 
 #include <vector>
+#include <algorithm>
```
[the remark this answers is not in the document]
edawson: Fixed in 4c8c0e6.

cudamapper/src/cudamapper_utils.cpp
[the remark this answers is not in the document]
edawson: Fixed in 8e951f9.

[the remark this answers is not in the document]
edawson: Yes, string_view looks like the answer for this. I'll add it to my list for the GPU port. I didn't know the STL had string_view - this would have saved me a lot of pointer work in other projects.

cudamapper/tests/Test_CudamapperUtilsKmerFunctions.cpp
```diff
@@ -0,0 +1,76 @@
+/*
+* Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
edawson: Fixed in 3bafbeae.

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,31 @@ namespace cudamapper
 /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
 void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overlap>& unfused_overlaps);
 
+std::string string_slice(const std::string& s, std::size_t start, std::size_t end);
+
+/// \brief Given a string s, produce its kmers (length <kmer-length>) and return them as a vector of strings.
+/// \param s A string sequence to kmerize.
+/// \param kmer_length A kmer length to use for producing kmers.
+/// \param stride The number of bases to skip when selecting kmers (most often, this should be equal to 1).
+std::vector<std::string> kmerize_string(const std::string& s, int kmer_length, int stride);
```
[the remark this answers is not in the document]
edawson: Fixed in 38b864c8.

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,31 @@ namespace cudamapper
 /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
 void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overlap>& unfused_overlaps);
 
+std::string string_slice(const std::string& s, std::size_t start, std::size_t end);
+
+/// \brief Given a string s, produce its kmers (length <kmer-length>) and return them as a vector of strings.
+/// \param s A string sequence to kmerize.
+/// \param kmer_length A kmer length to use for producing kmers.
```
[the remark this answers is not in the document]
edawson: Fixed in b373c7f0.

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -49,5 +49,31 @@ namespace cudamapper
 /// \param unfused_overlaps vector of overlaps, sorted by (query_id, target_id) combination and query_start_position
 void fuse_overlaps(std::vector<Overlap>& fused_overlaps, const std::vector<Overlap>& unfused_overlaps);
 
+std::string string_slice(const std::string& s, std::size_t start, std::size_t end);
```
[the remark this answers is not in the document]
edawson: Addressed here: 3c135fca, however this is now a local function in overlapper.cpp (addressed in aac541d9a060)

cudamapper/src/cudamapper_utils.hpp
edawson: Addressed in edd521b

cudamapper/src/cudamapper_utils.cpp
edawson: Addressed in aac541d. Again, I learned something new here!

cudamapper/src/cudamapper_utils.cpp
edawson: Addressed in previous comment and should be prioritized in the GPU port.

cudamapper/src/cudamapper_utils.cpp
edawson: Neat - addressed in fcbe66735.

cudamapper/src/cudamapper_utils.cpp
edawson: Yes it can - addressed in 8e951f9a.

cudamapper/src/cudamapper_utils.hpp
edawson: I intended for this to return `min(m,n)`; I don't want to discount duplicates. We could add a version to return *unique* shared elements in a new PR.

cudamapper/src/cudamapper_utils.cpp
edawson: I'll have to talk with @vellamike about how this might behave in low complexity genomic regions, but I think we want the "counting" behavior that an `unordered_map` would offer. We may have to test both to see.

cudamapper/src/overlapper.cpp
edawson: Addressed by 8e951f9a.

cudamapper/src/overlapper.cpp
edawson: Yep - addressed in  8e951f9a

cudamapper/src/overlapper.cpp
edawson: Addressed in 2b78224. I switched to `position_in_read_t_` which I think is the proper type to use here. I also fixed the bug described by @ahehn-nv .

edawson:
> For the example with perfect reads, why is the required similarity 0.5? It should be >0.99 if the reads are identical?

This has to do with allowing substrings of different lengths while using the Jaccard similarity.

In the following lines, I take the substring at the (head) end of the overlap. If there are at least `extension` basepairs remaining in the string, that substring is of length `extension`; otherwise it extends to the end of the string. I don't enforce equal-length substrings for query and target, so they may be of differing lengths.
https://github.com/edawson/ClaraGenomicsAnalysis/blob/4c8c0e65578c1d38b478792b6802ce38fdbff412/cudamapper/src/overlapper.cpp#L220-L224

I can see now why this was confusing. If these strings are different lengths, the similarity coefficient is not a robust estimator of their identity. For example: if A is a substring ( length 1/2 * |B| ) of B, the Jaccard similarity of (A, B) is only 0.5. I initially used the Jaccard containment for this reason, as the Jaccard containment of (A, B) in this case is 1.

I think the solution is:
- Use substrings that are the same length (requires a little extra math is all) and
- use the Jaccard similarity coefficient on these equal-length strings, with a default required_similarity of something closer to 1 (e.g. 0.9 if we want to allow some error).

edawson:
@mimaric Do our test harnesses and systems support `std::string_view`? I added a test for `count_shared_elements` to see if it would work with `string_view`. This ran fine on my local system but failed to compile with `file not found: string_view` in CI.

mimaric:
> @mimaric Do our test harnesses and systems support `std::string_view`? I added a test for `count_shared_elements` to see if it would work with `string_view`. This ran fine on my local system but failed to compile with `file not found: string_view` in CI.

`std::string_view` was added in C++17 and if I undersand correclty first added in GCC 7.1: https://gcc.gnu.org/onlinedocs/libstdc++/manual/status.html.  The minimal GCC version we support is 5.4 (I believe our tests also use that version). Older versions of GCC support `std::experimental::string_view` instead.
We had the same issue with `std::optional`, I'd suggest doing the same with `std::string_view`: https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/c298a2bc70cf038700fb61e1cfecf24a5f2b6edf/common/base/include/claragenomics/types.hpp#L36

cudamapper/src/cudamapper_utils.hpp
mimaric: I'm fine with whatever version suits your algorithm, no need to add function we don't need at the moment.
I would just extend the comment so it's clear to the user without looking into implementation.

edawson: Ah, got it! This is now documented in 069cae0.

cudamapper/src/cudamapper_utils.cpp
edawson: Should I consider this blocking for this PR? I've created a new branch that will address some of the performance issues, including moving to string_view, refactoring the main loop portion, and implementing a counting `unordered_map` approach.

edawson:
Results from 83cbbda:

```
(base) eric@odin:~/sandbox$ time ./clara/build/cudamapper/cudamapper e_coli.substring_reads.n5000.l1000.fa.txt e_coli.substring_reads.n5000.l1000.fa.txt | sort > norec.paf
Query file: e_coli.substring_reads.n5000.l1000.fa.txt, number of reads: 5000
Target file: e_coli.substring_reads.n5000.l1000.fa.txt, number of reads: 5000
Programmatically looking for max cached memory
Using device memory cache of 7703862313 bytes
Processing query range: (0 - 4999)

real    0m2.197s
user    0m1.325s
sys     0m0.267s
(base) eric@odin:~/sandbox$ time ./clara/build/cudamapper/cudamapper -R e_coli.substring_reads.n5000.l1000.fa.txt e_coli.substring_reads.n5000.l1000.fa.txt | sort > rec.paf
Query file: e_coli.substring_reads.n5000.l1000.fa.txt, number of reads: 5000
Target file: e_coli.substring_reads.n5000.l1000.fa.txt, number of reads: 5000
Programmatically looking for max cached memory
Using device memory cache of 7703862313 bytes
Processing query range: (0 - 4999)

real    0m2.211s
user    0m1.499s
sys     0m0.257s
(base) eric@odin:~/sandbox$
```
So speed is now comparable on a small test set, whether end rescue is one or not. There is some fluctuation that I assume is due to caching/paging.

Accuracy-wise, we're now extending many overlaps. Note the third and last columns:

```
(base) eric@odin:~/sandbox$ head -n 5 norec.paf | cut -f 1,3,4,5,6,8,9
E_coli_test_100kb_10017_11017_1000      0       529     +       E_coli_test_100kb_9561_10561_1000       456     985
E_coli_test_100kb_10017_11017_1000      0       537     +       E_coli_test_100kb_9575_10575_1000       442     979
E_coli_test_100kb_10017_11017_1000      0       553     +       E_coli_test_100kb_9587_10587_1000       430     983
E_coli_test_100kb_10017_11017_1000      0       564     +       E_coli_test_100kb_9599_10599_1000       418     982
E_coli_test_100kb_10017_11017_1000      0       564     +       E_coli_test_100kb_9607_10607_1000       410     974
(base) eric@odin:~/sandbox$ head -n 5 rec.paf | cut -f 1,3,4,5,6,8,9
E_coli_test_100kb_10017_11017_1000      0       544     +       E_coli_test_100kb_9561_10561_1000       456     1000
E_coli_test_100kb_10017_11017_1000      0       558     +       E_coli_test_100kb_9575_10575_1000       442     1000
E_coli_test_100kb_10017_11017_1000      0       570     +       E_coli_test_100kb_9587_10587_1000       430     1000
E_coli_test_100kb_10017_11017_1000      0       582     +       E_coli_test_100kb_9599_10599_1000       418     1000
E_coli_test_100kb_10017_11017_1000      0       590     +       E_coli_test_100kb_9607_10607_1000       410     1000
(base) eric@odin:~/sandbox$
```

These are perfect reads, so if the sequences truly overlap their identity will be `1.0`. We should test this on real data to establish a good default for the required_similarity to extend and overlap.

edawson:
@vellamike when you're ready, I think this is ready to review (again) and merge, though I'd like to push it through a lot more testing in PEF before it's recommended practice.

cudamapper/src/overlapper.cpp
```diff
@@ -55,6 +59,34 @@ bool overlaps_mergable(const claraparabricks::genomeworks::cudamapper::Overlap o
...
+        s[i]           = static_cast<char>(complement_array[static_cast<int>(s[len - 1 - i] - 65)]);
+        s[len - 1 - i] = static_cast<char>(complement_array[static_cast<int>(tmp) - 65]);
+    }
+}
+std::string string_slice(const std::string& s, const std::size_t start, const std::size_t end)
+{
+    return s.substr(start, end - start);
+}
+
+claraparabricks::genomeworks::cga_string_view_t string_view_slice(const claraparabricks::genomeworks::cga_string_view_t& s, const std::size_t start, const std::size_t end)
```
mimaric, line 85: Does this actually has to be a separate function? It's probably a matter of taste...

cudamapper/src/overlapper.cpp
```diff
@@ -152,6 +184,134 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps)
...
+    cga_string_view_t target_tail_sequence = string_view_slice(target_sequence, overlap.target_end_position_in_read_, overlap.target_end_position_in_read_ + tail_rescue_size);
+
+    const float tail_similarity = sequence_jaccard_similarity(query_tail_sequence, target_tail_sequence, 15, 1);
+    if (tail_similarity >= required_similarity)
+    {
+        overlap.query_end_position_in_read_  = overlap.query_end_position_in_read_ + tail_rescue_size;
+        overlap.target_end_position_in_read_ = overlap.target_end_position_in_read_ + tail_rescue_size;
+    }
+
+    //     std::cerr <<
```
mimaric, line 228: Please remove if not needed anymore

cudamapper/src/main.cu
```diff
@@ -122,6 +125,7 @@ struct ApplicationParameteres
     std::int32_t min_bases_per_residue          = 100;  // b
     float min_overlap_fraction                  = 0.95; // z
     bool all_to_all                             = false;
+    bool perform_overlap_end_rescue             = false;
```
mimaric, line 128: So easier understanding add `// R` and move above `bool all_to_all`

cudamapper/src/main.cu
```diff
@@ -406,9 +414,11 @@ void writer_thread_function(std::mutex& overlaps_writer_mtx,
                             const io::FastaParser& query_parser,
                             const io::FastaParser& target_parser,
                             const std::vector<std::string> cigar,
-                            const int32_t kmer_size,
                             const int32_t device_id,
-                            const int32_t number_of_device)
+                            const int32_t number_of_device,
+                            const int32_t kmer_size,
+                            bool perform_overlap_end_rescue)
```
mimaric, line 420: Add params to function documentation

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -45,6 +45,56 @@ void print_paf(const std::vector<Overlap>& overlaps,
                std::mutex& write_output_mutex,
                int32_t number_of_devices);
 
+// ///@brief Given a string s, return the substring s[start, end].
+// ///
+// ///@param s A string
+// ///@param start The 0-based start index of the returned substring.
+// ///@param end The 0-based end index of the returned substring
+// ///@return A std::string from s[start] to s[end].
+// std::string string_slice(const std::string& s, std::size_t start, std::size_t end);
```
mimaric, line 54: Remove if not needed anymore

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -45,6 +45,56 @@ void print_paf(const std::vector<Overlap>& overlaps,
...
+/// implemented in similarity) but is more robust to strings of different lengths.
+/// Note: This function assumes that a and b are on the same strand; you may need to
+/// reverse-complement one of the strings if testing similarity on strings from different
+/// strands.
+/// \param a A C++ string
+/// \param b A C++ string
+/// \param kmer_size The kmer length to use for estimating similarity.
+/// \param stride The number of bases to stride between kmers.
+/// \return The estimated Jaccard containment as a float.
+float sequence_jaccard_containment(const std::string& a, const std::string& b, std::int32_t kmer_size, std::int32_t stride);
```
mimaric, line 96: Did we switch to `sequence_jaccard_similarity()` completely? If yes then this function can be removed.

cudamapper/src/overlapper.cpp
```diff
@@ -7,15 +7,19 @@
 * distribution of this software and related documentation without an express
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
-
-#include <claragenomics/cudamapper/overlapper.hpp>
-
 #include <algorithm>
+#include <cstddef>
+#include <future>
+#include <mutex>
```
mimaric, line 13: Are these headers needed?

cudamapper/src/overlapper.cpp
```diff
@@ -7,15 +7,19 @@
...
 #include <vector>
 
+#include <claragenomics/cudamapper/overlapper.hpp>
 #include <claragenomics/io/fasta_parser.hpp>
 #include <claragenomics/utils/cudautils.hpp>
 #include <claragenomics/utils/signed_integer_utils.hpp>
+#include <claragenomics/types.hpp>
```
mimaric, line 20: This header is already included in `overlapper.hpp`

cudamapper/src/overlapper.cpp
```diff
@@ -7,15 +7,19 @@
...
-#include <claragenomics/cudamapper/overlapper.hpp>
-
 #include <algorithm>
+#include <cstddef>
+#include <future>
+#include <mutex>
 #include <vector>
 
+#include <claragenomics/cudamapper/overlapper.hpp>
 #include <claragenomics/io/fasta_parser.hpp>
```
mimaric, line 17: Not your change, but this header is already included in `overlapper.hpp` so you can remove it as well

cudamapper/src/cudamapper_utils.cpp
edawson: I've addressed this in 45e5f54 - the behavior is now to take equal-size windows in the query/target and compare using the Jaccard similarity coefficient.

CMakeLists.txt
```diff
@@ -90,4 +90,4 @@ cga_enable_formatting_targets()
 
 if (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
     set (CMAKE_INSTALL_PREFIX ${CMAKE_BINARY_DIR}/install CACHE PATH "The default SDK install path is install" FORCE)
-endif (CMAKE_INSTALL_PREFIX_INITIALIZED_TO_DEFAULT)
+endif ()
```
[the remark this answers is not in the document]
edawson: Removed in 0f03b38

README.md
```diff
@@ -43,15 +43,15 @@ This will clone the repo to the `master` branch, which contains code for latest
 and hot-fixes.
 
 ```
-git clone --recursive -b master git@github.com:clara-genomics/ClaraGenomicsAnalysis.git
+git clone --recursive -b master https://github.com/clara-genomics/ClaraGenomicsAnalysis.git
```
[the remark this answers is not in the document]
edawson: Removed in dd260cb

common/base/include/claragenomics/types.hpp
```diff
@@ -43,11 +45,13 @@ template <typename T>
 using cga_optional_t                = std::optional<T>;
 using cga_nullopt_t                 = std::nullopt_t;
 constexpr cga_nullopt_t cga_nullopt = std::nullopt;
+using cga_string_view_t             = std::string_view;
```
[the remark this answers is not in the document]
edawson: Noted in 498ab04

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -11,6 +11,11 @@
 #pragma once
 
 #include <claragenomics/cudamapper/types.hpp>
+#include <claragenomics/utils/genomeutils.hpp>
```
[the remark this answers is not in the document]
edawson: Nope! Removed in ee0dc07.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -11,6 +11,11 @@
...
+#include <claragenomics/io/fasta_parser.hpp>
+#include <claragenomics/types.hpp>
+
+#include <thrust/execution_policy.h>
```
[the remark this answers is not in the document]
edawson: Nope - removed in 5ce3718.

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
```diff
@@ -59,6 +64,33 @@ class Overlapper
...
+
+    /// \brief Extends a single overlap at its ends if the similarity of the query and target sequences is above a specified threshold.
+    /// \param overlap An Overlap which is modified in place. Any of the query_start_position_in_read, query_end_position_in_read,
+    /// target_start_position_in_read, and target_end_position_in_read fields may be modified.
+    /// \param query_sequence A std::string_view of the query read sequence.
+    /// \param target_sequence A std::string_view of the target read sequence.
+    /// \param extension The number of bases to extend at the head and tail of the overlap. If the head or tail is shorter than extension,
+    /// the function only tries to extend to the end of the read.
+    /// \param required_similarity The minimum similarity to require to extend an overlap.
+    static void extend_overlap_by_sequence_similarity(Overlap& overlap,
```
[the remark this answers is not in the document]
edawson: `rescue_overlap_ends()` calls `extend_overlap_by_sequence_similarity()`. We don't call `extend_overlap...` externally as things stand but it seems like a useful function to have around. Tests operate on `extend_overlap_by_sequence_similarity()` as well and I assume this means it needs to be exposed.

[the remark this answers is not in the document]
mimaric: It that case I would suggest doing something similar to this: https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/e43a18c4288f57ee554f81391905a24e159dda44/cudamapper/src/index_gpu.cuh#L158

In other words creating
```
namespace details
{
namespace overlapper
{
    void extend_overlap_by_sequence_similarity(){ ... }
}
}
```

That way the function can still be tested independently (which is a problem with protected/private), but it is clear that the function is not meant for external use (this can be changed if such need arises)

cudamapper/src/cudamapper_utils.hpp
edawson: Removed in 7291c2a

cudamapper/src/cudamapper_utils.hpp
edawson: Yes - removed in  23f1e8c.

cudamapper/src/main.cu
edawson: Fixed in 119007d.

cudamapper/src/overlapper.cpp
edawson: Apparently not - fixed in 8a8c81f.

cudamapper/src/overlapper.cpp
edawson: Yes, and some more STL headers weren't needed: fixed in 7fbc965.

cudamapper/src/overlapper.cpp
```diff
@@ -55,6 +59,34 @@ bool overlaps_mergable(const claraparabricks::genomeworks::cudamapper::Overlap o
...
+void reverse_complement(std::string& s, const std::size_t len)
+{
+    for (std::size_t i = 0; i < len / 2; ++i)
+    {
+        char tmp       = s[i];
+        s[i]           = static_cast<char>(complement_array[static_cast<int>(s[len - 1 - i] - 65)]);
+        s[len - 1 - i] = static_cast<char>(complement_array[static_cast<int>(tmp) - 65]);
+    }
+}
+std::string string_slice(const std::string& s, const std::size_t start, const std::size_t end)
```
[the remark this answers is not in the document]
edawson: Yes, we still need in-place reverse-complementing for handling overlaps on the negative relative strand.

cudamapper/src/overlapper.cpp
edawson: I agree this is a matter of taste, but I like having it as a function like so. Eventually we may want to refactor this to handle overlaps on the "-" strand without reverse complementing the whole string, and having the slice function modularized facilitates that possible change.

cudamapper/src/overlapper.cpp
edawson: Fixed in cf60534

pyclaragenomics/README.md
```diff
@@ -20,8 +20,8 @@ pip install pyclaragenomics-cuda10-1
 ```
 
 Details of the packages are available here -
-- https://pypi.org/project/pyclaragenomics-cuda10-0/#description
-- https://pypi.org/project/pyclaragenomics-cuda10-1/#description
+- https://pypi.org/project/pyclaragenomics-cuda-10-0
```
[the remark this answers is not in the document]
edawson: This is not mine - I'm going to leave it.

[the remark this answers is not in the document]
mimaric: @tijyojwad @vellamike @ohadmo Anyone know where does this change comes from?

cudamapper/src/overlapper.cpp
mimaric: Ok, I agree we can leave it as it is.

cudamapper/src/cudamapper_utils.cpp
mimaric: I would say we don't have to address this right now

cudamapper/include/claragenomics/cudamapper/overlapper.hpp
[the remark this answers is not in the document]
edawson: Awesome - fixed in c4d7caa.

cudamapper/src/main.cu
edawson: Fixed in 0c393c8be.

cudamapper/src/overlapper.cpp
edawson: Done

cudamapper/src/cudamapper_utils.cpp
edawson: I assume it would be invalidated by moving the function to device as well, correct?

pyclaragenomics/README.md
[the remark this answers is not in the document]
tijyojwad: This change was merged into `master` branch and not `dev-v0.5.0`. Maybe this branch merged in master at some point?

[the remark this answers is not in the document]
edawson: Oops, that probably was me then - I'll revert this.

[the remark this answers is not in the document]
edawson: Reverted in 0562f1c.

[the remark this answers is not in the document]
mimaric: Thanks @tijyojwad! Any reason why this changed wasn't merged in `dev-v0.5.0` as well? I guess the idea was that `dev-v0.5.0` will get merged into `master` soon and `dev-v0.6.0` will be branched off `master`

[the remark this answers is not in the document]
tijyojwad: @mimaric it was absolutely supposed to be merged into master but slipped my mind. Needed manual merging because of the version update in master (which overwrites version in dev-v0.5.0). And since we merge master into dev before releasing, I figured I'd just do it before the release :)

PR #423, opened 2020-05-09
[cudapoa] fix for Error code 8

r-mafi:
Fixed a bug in banded NW where in some cases would throw `loop_count_exceeded_upper_bound` error code and failed to complete the alignment.
It would happen for cases where trace-back path would hit left boundary of the band.

### DIFF
--- a/cudapoa/src/cudapoa_nw_banded.cu
+++ b/cudapoa/src/cudapoa_nw_banded.cu
@@ -54,7 +54,7 @@ __device__ ScoreT* get_score_ptr(ScoreT* scores, SizeT row, SizeT column, float 
 
     if (column == 0)
     {
-        col_idx = band_start;
+        col_idx = 0;
     }
     else
     {


cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -54,7 +54,7 @@ __device__ ScoreT* get_score_ptr(ScoreT* scores, SizeT row, SizeT column, float
 
     if (column == 0)
     {
-        col_idx = band_start;
+        col_idx = 0;
```
tijyojwad, line 57: looking at the else condition, `col_idx` is set to `column - band_start`. Is the bug that when `column == 0`, `column` is the same as `band_start`? if so,c an we simplify this to have `col_idx = column - band_start` always?

r-mafi: That is correct. But there are cases where the caller function calls with `0` as as column index. such as 
https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/src/cudapoa_nw_banded.cu#L238
https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/src/cudapoa_nw_banded.cu#L259
So I thought to have the fix with minimal changes to the original code.
The point is we have mixed uses of local-index (local relative the the band), or global-index (global-index = local-index + band_start).
We can make them consistent and remove the if condition here and similarly in some other kernels such as set_score.
I think this was designed with reducing calls to `get_band_start_for_row()`, but this function needs to be called anyways at some point to convert local to global index.

tijyojwad: I see. We should make a clear distinction between the local and global index. But it doesn't have to be in this PR. Can you file an issue for this? Approving this PR. Thanks for the explanation!

PR #427, opened 2020-05-12
[cudapoa] optimal multi batch for sample_cudapoa

r-mafi:
Added a new function in `cudapoa/utils.hpp` where groups input windows based on their sizes in different batch-size bins. Windows with similar range of size are grouped together and launched with the same batch-size.
This multi-batch strategy can have a significant impact on scenarios where variance of window size is pretty high. If all windows have a similar size, multi-batch strategy will transform into a single batch launch.

- added using optimal multi-batch in `sample_cudapoa`, to find a small set of batches to cover all windows while maintaining high GPU occupancy
- changed default band-width from 128 to 256


cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -203,102 +348,117 @@ int main(int argc, char** argv)
...
-        {
-            // Check if all sequences in POA group wre added successfully.
-            for (const auto& s : seq_status)
+            if (status != StatusType::exceeded_maximum_poas && status != StatusType::success)
             {
-                if (s == StatusType::exceeded_maximum_sequence_size)
-                {
-                    std::cerr << "Dropping sequence because sequence exceeded maximum size" << std::endl;
-                }
+                std::cout << "Could not add POA group " << batch_window_ids[i] << "to batch " << b << ". Error code " << status << std::endl;
```
r-mafi, line 456: Considering that some of the windows with full alignment won't run on most of GPU cards, we can add more details to this error message, like what is the memory requirement and what is the available GPU memory.

r-mafi:
we can upload the 1000 window and 4000 window bonito dataset somewhere accessible for future benchmarking.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -124,22 +125,148 @@ void process_batch(Batch* batch, bool msa, bool print)
...
 
-void generate_window_data(const std::string& input_file, const int number_of_windows, const int max_sequences_per_poa,
-                          std::vector<std::vector<std::string>>& windows, BatchSize& batch_size)
+size_t estimate_max_poas(const BatchSize& batch_size, const bool banded_alignment, const bool msa_flag)
 {
-    parse_window_data_file(windows, input_file, number_of_windows); // Generate windows.
-    assert(get_size(windows) > 0);
+    int32_t matrix_sequence_dimension = banded_alignment ? CUDAPOA_BANDED_MAX_MATRIX_SEQUENCE_DIMENSION : batch_size.max_matrix_sequence_dimension;
+    int32_t matrix_graph_dimension    = banded_alignment ? batch_size.max_matrix_graph_dimension_banded : batch_size.max_matrix_graph_dimension;
+    int32_t max_nodes_per_window      = banded_alignment ? batch_size.max_nodes_per_window_banded : batch_size.max_nodes_per_window;
```
atadkase, line 132: Hey Ramin,
Is there a reason why we're using signed instead of unsigned?

r-mafi: Good point, although for size types it makes more sense to use unsigned int, but to be consistent with [`BatchSize`](https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/include/claragenomics/cudapoa/batch.hpp#L50), they're defined as `int32_t`
Now the question is why `BatchSize` uses signed integer? The main reason is there were a mixed set of signed and unsigned integers used in cudaPOA which all changed into a template type. Since we could cover unsigned cases with signed type, but not the other way around, all such data in cudaPOA were changed to signed integers.
so to keep it consistent with the rest of the code, let's leave it as it is.

ahehn-nv: I just saw this discussion and would like to comment.
Ideally, we should use signed integers for everything that should represent an integer number - even if it cannot be less than zero. Exceptions are if we really need the extra bit of unsigned to represent the large values (and cannot just use a larger signed integer), use the variable for bit-operations, or explicitly want modulo 2^type_bits arithmetic.
In short, the main reasons are that arithmetic with signed integers is less error prone and that signed integers can be checked for overflow or negative value errors (sometimes they also give the compiler a tiny bit more freedom to optimize).
(See, e.g. http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-mix and following guidelines.)
Today we _almost_ had one of these signed-unsigned arithmetic bugs: https://github.com/clara-parabricks/ClaraGenomicsAnalysis/pull/422#discussion_r423660027

atadkase: Thanks. That makes sense. I was just curious about this since the conventions for signed/unsigned seemed to be a bit mixed (size_t, intX_t, uintX_t).

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -124,22 +125,148 @@ void process_batch(Batch* batch, bool msa, bool print)
...
+
+    int64_t sizeof_ScoreT = use32bitScore(batch_size, gap_score, mismatch_score, match_score) ? 4 : 2;
+    int64_t sizeof_SizeT  = use32bitSize(batch_size, banded_alignment) ? 4 : 2;
+
+    // Calculate memory requirements for POA arrays
+    int64_t device_size_per_poa = 0;
+    int64_t input_size_per_poa  = batch_size.max_sequences_per_poa * batch_size.max_sequence_size;
+    int64_t output_size_per_poa = batch_size.max_concensus_size;
+    device_size_per_poa += output_size_per_poa * sizeof(uint8_t);                                                                                // output_details_d_->consensus
+    device_size_per_poa += (!msa_flag) ? output_size_per_poa * sizeof(uint16_t) : 0;                                                             // output_details_d_->coverage
```
atadkase, line 148: The size calculation logic overlaps with logic in calculate_space_per_poa in cudapoa/src/allocate_block.hpp
Would it help to split the space calculation into individual functions per detail so that you can reuse those there and here?

atadkase: (if the code cannot be reused)

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -124,22 +125,148 @@ void process_batch(Batch* batch, bool msa, bool print)
...
+    // is smaller than N, they can all fit in batch N and no need to create extra batches.
+    // For example. consider the following:
+    //
+    // bins_ranges      1        2       4       8       16      32      64      128     256     512
+    // bins frequency   0        0       0       0       0       0       10      51      0       0
+    // bins width       0        0       0       0       0       0       5120    3604    0       0
+    //
+    // note that bin_ranges represent max POAs. This means larger bin ranges follow with smaller corresponding max lengths (windows with shorter reads)
+    // In the example above, to process 10 windows that fall within bin range 64, we need to create one batch. This batch can process up to 64 windows of
+    // max length 5120 or smaller. This means all the windows in bin range 128 can also be processed with the same batch and no need to launch an extra batch
```
atadkase, line 249: Does the resource allocation for the batch change if we're using windows from bin_range 64 and add bin_range 128?

cudapoa/samples/sample_cudapoa.cpp
r-mafi: that would definitely follow DRY rule, but I was debating whether to keep this component for the arrays or just estimate based on score matrix size. I possibly keep the estimation as is and rewrite it without repeating the code.

cudapoa/samples/sample_cudapoa.cpp
r-mafi: no, once a batch is created with a given `batch_size`, it remains the same for all the windows that are added to it, whether they could be launched with a different `batch_size` or not.

cudapoa/samples/sample_cudapoa.cpp
ahehn-nv: Yes, unfortunately, we are not really consistent here (yet?), which is also dangerous. Nevertheless, I would prefer signed integers wherever possible (and avoid size_t).

cudapoa/samples/sample_cudapoa.cpp

cudapoa/samples/sample_cudapoa.cpp
r-mafi: changed to minimize repeated code.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -124,22 +126,127 @@ void process_batch(Batch* batch, bool msa, bool print)
...
+
+    size_t remaining_windows = num_windows;
+    for (size_t j = 0; j < num_bins; j++)
+    {
+        if (bins_frequency[j] > 0)
+        {
+            list_of_batch_sizes.emplace_back(bins_max_length[j], bins_num_reads[j]);
+            list_of_windows_per_batch.push_back(bins_window_list[j]);
+            remaining_windows -= bins_frequency[j];
+            if (bins_ranges[j] > remaining_windows)
```
atadkase, line 239: This will possibly result in working only at the tail end of the histogram. Is that intended?
```
 // bins_ranges       1        2       4       8       16      32      64      128     256     512
 // bins frequency    0        0       2       2       0       0       10      51      0       0
 // bins width            0        0       X       Y(<X)       0       0       5120    3604    0       0
```

r-mafi: That's correct. It was intentional, since in practice, or at least in the bonito tests cases, there were fewer windows that needed the larger bins (those with smaller max POA), and majority of the windows would be toward the tail. So I did not put any effort on collapsing intermediate bins into each other. Toward the end, each bin has a higher capacity (larger max POA) meaning corresponding batch can fit larger number of windows launched in parallel, also as mentioned earlier, most of windows fall within these ranges, therefore it was written to simply collapse the last bins into one.

cudapoa/samples/sample_cudapoa.cpp
atadkase: This looks good!

r-mafi:
@tijyojwad at some point you mentioned we better keep the sample_cudapoa simple and move binning algorithm and optimal multi-batch elsewhere. Please share your thoughts in more details, as this PR has been sitting here for a while. Thanks.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -124,22 +126,127 @@ void process_batch(Batch* batch, bool msa, bool print)
...
-        for (auto& seq : window)
+        sizeof_ScoreT = 4;
+        if (use32bitSize(batch_size, banded_alignment))
         {
-            max_read_length = std::max(max_read_length, get_size<int>(seq));
+            device_size_per_poa = BatchBlock<int32_t, int32_t>::compute_device_memory_per_poa(batch_size, banded_alignment, msa_flag);
         }
+        else
+        {
+            device_size_per_poa = BatchBlock<int32_t, int16_t>::compute_device_memory_per_poa(batch_size, banded_alignment, msa_flag);
```
tijyojwad, line 148: Since this is an implementation class, this breaks the abstraction barrier of only using public headers (i.e. those in the `include` folder)

tijyojwad:
@r-mafi The main concern with this PR is that it breaks public/private abstraction for our APIs in the sample, since the sample program uses headers files from the `src` folder instead of only from the `include` folder which exposes the public API.

Now since binning is quite useful for optimal cudapoa usage, we can expose the binning strategy in a `cudapoa/utils.hpp` header that takes in a vector of window sizes and returns the optimal binning strategy. The benefit of this is we can expose this functionality through python bindings as well, and then users across both C++ and python APIs can make use of it. What do you think?

r-mafi:
sounds good, will modify this PR and move binning algorithm that Ashutosh reviewed to `cudapoa/utils.hpp` in a separate PR.

r-mafi:
moved part of this PR to https://github.com/clara-parabricks/ClaraGenomicsAnalysis/pull/434, will modify this one after.

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -9,6 +9,8 @@
 */
 
 #include "../benchmarks/common/utils.hpp"
+#include "../src/cudapoa_kernels.cuh" // for estimate_max_poas()
```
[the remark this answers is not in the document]
r-mafi: refactored and moved these changes mainly to  `utils.hpp` and `allocate_block.hpp`

cudapoa/samples/sample_cudapoa.cpp
r-mafi: fixed

r-mafi:
rerun tests

cudapoa/samples/sample_cudapoa.cpp
r-mafi: @atadkase trying with some other datasets, I noticed it is needed to revise the strategy as you pointed, to merge intermediate bins whenever possible as well. I will revise it to be more general.

r-mafi: changed.

cudapoa/benchmarks/common/utils.hpp
```diff
@@ -93,6 +95,139 @@ inline std::string parse_golden_value_file(const std::string& filename)
...
+/// \param list_of_windows_per_batch [out] corresponding windows per batch-size
+/// \param windows [in] vector of input windows
+/// \param banded_alignment [in] flag indicating whether banded-alignment or full-alignment is going to be used, default is banded-alignment
+/// \param msa_flag [in] flag indicating whether MSA or consensus is going to be computed, default is consensus
+/// \param bins_capacity [in] pointer to vector of bins used to create separate different-sized windows, if null as input, a set of default bins will be used
+/// \param gpu_memory_usage_quota [in] portion of GPU available memory that will be used for compute each cudaPOA batch, default 0.9
+/// \param mismatch_score [in] mismatch score, default -6
+/// \param gap_score [in] gap score, default -8
+/// \param match_score [in] match core, default 8
+void get_multi_batch_sizes(std::vector<BatchSize>& list_of_batch_sizes,
```
tijyojwad, line 115: let's move this function into the cudapoa library API. that way it can be accessible through the public API.

location - `cudapoa/include/claragenomics/cudapoa/utils.hpp` for the function signature

couple of updates needed to the API - 
1. the `windows` argument should be updated to `poa_groups` and should accept `const std::vector<Group>&` with Group as defined in https://github.com/clara-parabricks/ClaraGenomicsAnalysis/blob/dev-v0.5.0/cudapoa/include/claragenomics/cudapoa/batch.hpp#L50
2. `list_of_windows_per_batch` should be `list_of_groups_per_batch`, and the documentation should reflect that accordingly

the implementation can be in `cudapoa/src/utils.cpp`

cudapoa/include/claragenomics/cudapoa/utils.hpp
```diff
@@ -0,0 +1,161 @@
...
+* NVIDIA CORPORATION and its licensors retain all intellectual property
+* and proprietary rights in and to this software, related documentation
+* and any modifications thereto.  Any use, reproduction, disclosure or
+* distribution of this software and related documentation without an express
+* license agreement from NVIDIA CORPORATION is strictly prohibited.
+*/
+
+#pragma once
+
+#include "../../../src/allocate_block.hpp" // for estimate_max_poas()
```
tijyojwad, line 13: @r-mafi I think we would need to move the implementation into a cpp file to hide this include, otherwise this is leaking into the public header. Once the implementation is in cpp, this include will move to the cpp file as well

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -8,7 +8,8 @@
 * license agreement from NVIDIA CORPORATION is strictly prohibited.
 */
 
-#include "../benchmarks/common/utils.hpp"
+#include "../include/claragenomics/cudapoa/utils.hpp" // for get_multi_batch_sizes()
```
tijyojwad, line 11: And this can just become `#include <claraparabricks/genomeworks/cudapoa/utils.hpp>`, much like thebutils included below

r-mafi:
moved the changes related to this PR to #464

Issue #424, opened 2020-05-10
[SDK] rename SDK to GenomeWorks


tijyojwad:
Description to be updated once final name of repo is decided

PR #430, opened 2020-05-13
[sdk] update namespace across whole repo

tijyojwad:
as per our discussions, added `claraparabricks::genomeworks` namespace to all modules

Part of #424

### DIFF
--- a/common/base/include/claragenomics/logging/logging.hpp
+++ b/common/base/include/claragenomics/logging/logging.hpp
@@ -93,6 +93,9 @@
 #include <spdlog/spdlog.h>
 #endif
 
+namespace claraparabricks
+{
+
 namespace genomeworks
 {
 namespace logging
@@ -177,5 +180,7 @@ LoggingStatus SetHeader(bool logTime, bool logLocation);
 
 } // namespace logging
 } // namespace genomeworks
+
+} // namespace claraparabricks
 
 /// \}
--- a/common/base/include/claragenomics/logging/logging.hpp
+++ b/common/base/include/claragenomics/logging/logging.hpp
@@ -98,6 +98,7 @@ namespace claraparabricks
 
 namespace genomeworks
 {
+
 namespace logging
 {
 /// \ingroup logging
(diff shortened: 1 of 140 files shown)


tijyojwad:
@vellamike @ahehn-nv @mimaric @r-mafi - whoever gets a chance to review it first, if you approve can you merge as well? Thanks!

common/base/include/claragenomics/logging/logging.hpp
```diff
@@ -93,7 +93,10 @@
...
 #endif
 
-namespace claragenomics
+namespace claraparabricks
+{
+
+namespace genomeworks
 {
 namespace logging
 {
```
mimaric, line 102: Do we want to have a new line between namespaces? We did not do it previously.
I'm fine with both, we should just pick one and stick to it

tijyojwad: It's not consistent right now. Some files have a new line and some don't. I want to make it consistent with a new line. I'll push a change to fix the older ones too

tijyojwad: Updated

common/base/include/claragenomics/utils/cudautils.hpp
```diff
@@ -147,7 +150,7 @@ std::size_t find_largest_contiguous_device_memory_section();
 /// \brief starts an NVTX range for profiling which stops automatically at the end of the scope
 /// \param varname an arbitrary variable name for the nvtx_range object, which doesn't conflict with other variables in the scope
 /// \param label the label/name of the NVTX range
-#define CGA_NVTX_RANGE(varname, label) ::claragenomics::cudautils::nvtx_range varname(label)
+#define CGA_NVTX_RANGE(varname, label) ::genomeworks::cudautils::nvtx_range varname(label)
```
ahehn-nv, line 153: This should be `::claraparabricks::genomeworks::cudautils::nvtx_range`

tijyojwad: good catch! fixed, and added profiling flag to release mode testing

PR #434, opened 2020-05-18
[cudapoa] changing sample file for bonito example

r-mafi:
- Current sample file based on bonito dataset is too large. Changed sample-bonito.txt to a new one with 55 windows and size < 5MB

- changed default mode of alignment in sample_cudapoa to banded

- modified the ad-hoc memory check in `BatchBlock` constructor

- modified error message where a large POA can't be added with more details about memory requirements

the above changes are partly taken from PR #427


cudapoa/src/cudapoa_batch.hpp
```diff
@@ -547,6 +548,12 @@ class CudapoaBatch : public Batch
 
         if (scores_size > avail_scorebuf_mem_)
         {
+            if (get_total_poas() == 0)
+            {
+                std::cout << "Memory available " << std::fixed << std::setprecision(2) << ((double)avail_scorebuf_mem_) / 1024. / 1024. / 1024.;
```
tijyojwad, line 553: `static_cast` for casting `avail_scorebuf_mem`

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -203,7 +201,7 @@ int main(int argc, char** argv)
     if (long_read)
     {
         const std::string input_file = std::string(CUDAPOA_BENCHMARK_DATA_DIR) + "/sample-bonito.txt";
-        generate_window_data(input_file, 8, 6, windows, batch_size);
+        generate_window_data(input_file, 55, 6, windows, batch_size);
```
tijyojwad, line 204: Do we need 55 specifically, or is that all the windows in the file? if you send `number_of_windows` as `-1` to the `parse_windows_file` function it returns all the windows in that file.

cudapoa/src/cudapoa_batch.hpp
r-mafi: good catch, it was an oversight due to copy/paste from an older code I had. fixed.

cudapoa/samples/sample_cudapoa.cpp
r-mafi: That's total number of windows. I wrote it consistent with the short-reads that explicitly says 1000, will change both to -1 then.

r-mafi: on a second thought, since it's sample, to demonstrate an example of both use cases, I changed back the short reads to 1000.

tijyojwad:
@r-mafi approved, and merged dev-v0.5.0 and kicked off tests

r-mafi:
> @r-mafi approved, and merged dev-v0.5.0 and kicked off tests

cool, thanks!

PR #437, opened 2020-05-19
[cudamapper] Using new index cache

### LINKED ISSUE #304: [cudamapper] segfault when running with multi GPU
Running with multi-GPU on any sizeable dataset (e.g 20k human ONT reads) is resulting in a segfault:

```
time /home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamapper /data/20k_reads_shuffled-chrX.fasta /data/20k_reads_shuffled-chrX.fasta -i 30 -w 5 -k 15 -c 30 -F 1e-5 -d 2 > out 

Query /data/20k_reads_shuffled-chrX.fasta index 20000
Target /data/20k_reads_shuffled-chrX.fasta index 20000
Processing query range: (0 - 1116)
Processing query range: (1117 - 2230)
Processing query range: (2231 - 3257)
Processing query range: (3258 - 4370)
Processing query range: (4371 - 5444)
Processing query range: (5445 - 6539)
Processing query range: (6540 - 7597)
Processing query range: (7598 - 8664)
Processing query range: (8665 - 9710)
[2]    5992 segmentation fault (core dumped)  /home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamappe
/home/mike/dev/ClaraGenomicsAnalysis/cmake-build-release/cudamapper/cudamappe  36.29s user 4.13s system 531% cpu 7.611 total
```

mimaric:
`main.cu` uses new index caching strategy with splitting target-query pairs into batches and uses `IndexCacheHost` and `IndexCacheDevice`.

Part of issue #318. Still not overlapping fetching indices from host with calculating overlaps.

It also makes multi-GPU run correctly again. Closes #304 

Sorry for not filing `ApplicationParameters` as a separate PR, it kind of got out of hand.

@edawson I rewrote `main.cu` almost entirely so the place where post-processing is done has changed, as well as where command line options are added. I'll adapt it to your changes from PR #422 as it will likely get merged before this PR, this is just FYI


cudamapper/src/main.cu
```diff
@@ -389,403 +182,331 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
-    // to indices which have device arrays associated with streams
-    for (cudaStream_t cuda_stream : cuda_streams)
+    // wait for all work to be done
+    for (int32_t device_id = 0; device_id < parameters.num_devices; ++device_id)
     {
-        CGA_CU_CHECK_ERR(cudaStreamSynchronize(cuda_stream));
-        CGA_CU_CHECK_ERR(cudaStreamDestroy(cuda_stream));
+        CGA_CU_CHECK_ERR(cudaSetDevice(device_id));
+        worker_threads[device_id].join();
+        CGA_CU_CHECK_ERR(cudaStreamDestroy(cuda_streams[device_id])); // no need to sync, it should be done at the end of worker_threads
```
vellamike, line 509: Should we put a device sync here anyway?

mimaric:
@edawson Can you check if I merged in your changes from PR #422 into `cudamapper/src/main.cu`, `cudamapper/src/application_parameters.hpp` and `cudamapper/src/application_parameters.cpp` correctly?

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -45,6 +46,11 @@ void print_paf(const std::vector<Overlap>& overlaps,
                std::mutex& write_output_mutex,
                int32_t number_of_devices);
 
+/// \brief crated a device allocator
```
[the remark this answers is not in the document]
mimaric: yes, done

cudamapper/src/main.cu
ahehn-nv: I'm not quite sure I'm following @mimaric.
I think it is a good idea that a worker thread only finishes when all the work that was assigned to it is done. So this implies a `cudaStreamSynchronize()` for each thread on its assigned stream. So, I don't see a reason for a device sync here.

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
 
-                OverlapperTriggered overlapper(allocator, cuda_stream);
-                CGA_NVTX_RANGE(profiler, "generate_overlaps");
+    // keep processing batches of indices until there are none left
+    cga_optional_t<BatchOfIndices> batch_of_indices = batches_of_indices.get_next_element();
+    while (batch_of_indices) // if optional is empty that means that there are no more batches to process and the thread can finish
+    {
+        const int64_t batch_number         = number_of_processed_batches.fetch_add(1); // as this is not called atomically with get_next_element() the value does not have to be completely accurate, but this is ok as the value is only use for displaying progress
+        const std::string progress_message = "Device " + std::to_string(device_id) + " took batch " + std::to_string(batch_number + 1) + " out of " + std::to_string(number_of_total_batches) + " batches in total\n";
+        std::cerr << progress_message; // TODO: possible race condition, switch to logging library
```
ahehn-nv, line 435: Will this be fixed in a separate PR? I'm wondering if we should remove it for the time being. What do you think?

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
+}
 
-    //First generate all the ranges independently, then loop over them.
-    std::vector<QueryTargetsRange> query_target_ranges;
+/// \brief does post-processing and writes data to output
+/// \param device_id
+/// \param application_parameters
+/// \param overlaps_and_cigars_to_write new data is added as it gets available
+/// \param output_mutex controls access to output to prevent race conditions
+void writer_thread_function(const int32_t device_id,
```
ahehn-nv, line 319: I find the name a bit misleading, because this function also does post-processing. How about renaming it to `postprocess_and_write_thread_function` or similar.

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
+{
+    CGA_NVTX_RANGE(profiler, ("main::writer_thread_for_device_" + std::to_string(device_id)).c_str());
+    // This function is expected to run in a separate thread so set current device in order to avoid problems
+    CGA_CU_CHECK_ERR(cudaSetDevice(device_id));
 
-    int target_idx = 0;
-    for (const IndexDescriptor& query_index_descriptor : query_index_descriptors)
+    // keep processing data as it arrives
+    cga_optional_t<OverlapsAndCigars> data_to_write = overlaps_and_cigars_to_write.get_next_element();
+    while (data_to_write) // if optional is empty that means that there will be no more overlaps to process and the thread can finish
```
ahehn-nv, line 330: Optional: I guess it's a matter of taste, but I would find `while(data_to_write = overlaps_and_cigars_to_write.get_next_element())` more intuitive than the assignment before and in the loop. It tells me while this assignment makes sense, do the loop.

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
+                                    std::ref(application_parameters),
+                                    std::ref(overlaps_and_cigars_to_write),
+                                    std::ref(output_mutex));
+    }
 
-                OverlapperTriggered overlapper(allocator, cuda_stream);
-                CGA_NVTX_RANGE(profiler, "generate_overlaps");
+    // keep processing batches of indices until there are none left
+    cga_optional_t<BatchOfIndices> batch_of_indices = batches_of_indices.get_next_element();
+    while (batch_of_indices) // if optional is empty that means that there are no more batches to process and the thread can finish
```
ahehn-nv, line 431: Optional: see other while loop above.

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
+/// \param host_cache data will be loaded into cache within the function
+/// \param device_cache data will be loaded into cache within the function
+/// \param overlaps_and_cigars_to_write overlaps and cigars are output here
+/// \param cuda_stream
+void process_one_batch(const BatchOfIndices& batch,
+                       const ApplicationParameters& application_parameters,
+                       DefaultDeviceAllocator device_allocator,
+                       IndexCacheHost& host_cache,
+                       IndexCacheDevice& device_cache,
+                       ThreadsafeProducerConsumer<OverlapsAndCigars>& overlaps_and_cigars_to_write,
```
ahehn-nv, line 277: Optional: How about `overlaps_and_cigars_to_postprocess` instead?

cudamapper/src/application_parameters.cpp
```diff
@@ -0,0 +1,307 @@
...
+            break;
+        case 't':
+            target_index_size = atoi(optarg);
+            break;
+        case 'F':
+            filtering_parameter = atof(optarg);
+            break;
+        case 'a':
+            alignment_engines = atoi(optarg);
+            throw_on_negative(alignment_engines, "Number of alignment engines should be non-negative");
```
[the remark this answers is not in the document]
ahehn-nv: Actually there are more values which should check that the values are non-negative. I guess this should be a separate PR...

cudamapper/src/main.cu
mimaric: Yes, that's exactly what I meant.
@vellamike Are you ok with leaving it as it is?

cudamapper/src/application_parameters.hpp
```diff
@@ -0,0 +1,86 @@
...
+    ///
+    /// If max_cached_memory is set that value is used, finds almost complete amount of available memory otherwise
+    /// Returns 0 if CGA_ENABLE_CACHING_ALLOCATOR is not set
+    ///
+    /// \return max_cached_memory_bytes
+    int64_t get_max_cached_memory_bytes();
+
+    /// @brief prints help message
+    /// @param exit_code
+    void help(int32_t exit_code = 0);
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/application_parameters.cpp
```diff
@@ -0,0 +1,307 @@
...
+        target_index_size = index_size;
+        std::cerr << "NOTE - Since query and target files are same, activating all_to_all mode. Query index size used for both files." << std::endl;
+    }
+
+    get_input_parsers(query_parser, target_parser);
+
+    max_cached_memory_bytes = get_max_cached_memory_bytes();
+}
+
+void ApplicationParameters::get_input_parsers(std::shared_ptr<io::FastaParser>& query_parser,
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/main.cu
mimaric: I would leave it as it is for now. In the case of a race condition progress message would still be visible, it would just be interleaved.
I find it very useful to have some form of progress indicator, especially for larger genomes. Initially I was only writing "new batch started" but then realized having some number is really convenient, even if it is not completely accurate

cudamapper/src/main.cu
mimaric: done

cudamapper/src/main.cu
mimaric: I went for `overlaps_and_cigars_to_process()` as at least for me that means both postprocess and write. What do you think?

cudamapper/src/main.cu
mimaric: You're right, that does look more readable.
done

cudamapper/src/main.cu
mimaric: done

cudamapper/src/main.cu
```diff
@@ -397,415 +182,341 @@ void align_overlaps(DefaultDeviceAllocator allocator,
...
+    // create host_cache, data is not loaded at this point but later as each batch gets processed
+    IndexCacheDevice device_cache(application_parameters.all_to_all,
+                                  host_cache);
+
+    // data structure used to exchnage data with writer_thread
+    ThreadsafeProducerConsumer<OverlapsAndCigars> overlaps_and_cigars_to_write;
+
+    // There should be at least one writer_thread per worker_thread. If more threads are available one thread should be reserved for
+    // worker_thread and all other threads should be writer_threads
+    const int32_t threads_per_device        = (std::thread::hardware_concurrency() + application_parameters.num_devices - 1) / application_parameters.num_devices;
```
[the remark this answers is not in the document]
mimaric: Yes, I forgot we have it.
done

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -36,14 +37,17 @@ namespace cudamapper
...
                const io::FastaParser& target_parser,
                int32_t kmer_size,
-               std::mutex& write_output_mutex,
-               int32_t number_of_devices);
+               std::mutex& write_output_mutex);
+
+/// \brief create a device allocator
+/// \param max_cached_memory_bytes
+/// \return device allocator
+DefaultDeviceAllocator get_device_allocator(const size_t max_cached_memory_bytes);
```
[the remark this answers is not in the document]
mimaric: True!
done

cudamapper/src/application_parameters.cpp
```diff
@@ -0,0 +1,307 @@
...
+
+    bool target_indices_in_host_memory_set   = false;
+    bool target_indices_in_device_memory_set = false;
+    int32_t argument                         = 0;
+    while ((argument = getopt_long(argc, argv, optstring.c_str(), options, nullptr)) != -1)
+    {
+        switch (argument)
+        {
+        case 'k':
+            kmer_size = atoi(optarg);
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/main.cu
ahehn-nv: yes, makes sense.

cudamapper/src/application_parameters.cpp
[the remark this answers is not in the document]
mimaric: Yes, I would leave this for a separate PR. We should generally standardize checks in this function.

cudamapper/src/main.cu
ahehn-nv: sounds good.

cudamapper/src/application_parameters.cpp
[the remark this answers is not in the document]
ahehn-nv: Opened #451 for this.

cudamapper/src/main.cu
ahehn-nv: :) Glad you agree.

cudamapper/src/main.cu
vellamike: Yes

mimaric:
rerun tests

mimaric:
rerun tests

PR #444, opened 2020-05-21
[cudapoa] variable bandwidth

r-mafi:
changed banded-alignment to accept different lengths of bandwidth


cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -79,16 +81,19 @@ struct BatchSize
         , max_matrix_graph_dimension(cudautils::align<int32_t, 4>(max_nodes_per_window))
         , max_matrix_graph_dimension_banded(cudautils::align<int32_t, 4>(max_nodes_per_window_banded))
         , max_matrix_sequence_dimension(cudautils::align<int32_t, 4>(max_sequence_size))
+        /// ensure 32-byte alignment for bandwidth
+        , alignment_bandwidth(cudautils::align<int32_t, 32>(bandwidth))
```
r-mafi, line 85: I guess this should change to be aligned with 128 instead of 32.

cudapoa/src/cudapoa_nw_banded.cu
```diff
@@ -197,16 +200,16 @@ __device__
...
+    int64_t score_index;
 
     //Calculate gradient for the scores matrix
     float gradient = float(read_length + 1) / float(graph_count + 1);
 
-    SizeT band_width = WARP_SIZE * CELLS_PER_THREAD;
-
-    SizeT max_column = read_length + 1;
+    SizeT max_column                    = read_length + 1;
+    SizeT max_matrix_sequence_dimension = band_width + CUDAPOA_BANDED_MATRIX_RIGHT_PADDING;
```
tijyojwad, line 209: I don't recall now what the `CUDAPOA_BANDED_MATRIX_RIGHT_PADDING` was for. It looks like it has to be added to every indexing parameter. can you remind me what it was for? It could become a source of bugs if it's not added carefully to each index variable, so I'm wondering if there's a way to get rid of it...

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -56,21 +56,23 @@ struct BatchSize
...
-    /// Maximum vertical dimension of scoring matrix, which stores graph
+    /// Maximum vertical dimension of scoring matrix, which stores POA graph
     int32_t max_matrix_graph_dimension = max_nodes_per_window;
-    /// Maximum vertical dimension of scoring matrix, which stores graph
+    /// Maximum vertical dimension of scoring matrix in banded alignment, which stores POA graph
     int32_t max_matrix_graph_dimension_banded = max_nodes_per_window_banded;
     /// Maximum horizontal dimension of scoring matrix, which stores sequences
     int32_t max_matrix_sequence_dimension = max_sequence_size;
+    /// Bandwidth used in banded alignment
+    int32_t alignment_bandwidth;
```
tijyojwad, line 70: can we rename this to `alignment_band_width`? it's a cosmetic change, but since they differ sementically from the signal processing use of `bandwidth` I just wanted to make that distinction :)

pyclaragenomics/claragenomics/bindings/cudapoa.pyx
```diff
@@ -73,6 +73,7 @@ cdef class CudaPoaBatch:
             mismatch_score=-6,
             match_score=8,
             cuda_banded_alignment=False,
+            bandwidth_size=128,
```
tijyojwad, line 76: `bandwidth_size` -> `alignment_band_width`

cudapoa/include/claragenomics/cudapoa/batch.hpp
```diff
@@ -79,16 +81,19 @@ struct BatchSize
         , max_matrix_graph_dimension(cudautils::align<int32_t, 4>(max_nodes_per_window))
         , max_matrix_graph_dimension_banded(cudautils::align<int32_t, 4>(max_nodes_per_window_banded))
         , max_matrix_sequence_dimension(cudautils::align<int32_t, 4>(max_sequence_size))
+        /// ensure 128-alignment for bandwidth size
+        , alignment_bandwidth(cudautils::align<int32_t, 128>(bandwidth))
```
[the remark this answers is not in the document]
r-mafi: that's right. done.

cudapoa/src/cudapoa_nw_banded.cu
r-mafi: from what I understand, bands per row are centered around diagonal. Depending on the shape of the NW matrix (how tall or fat it is), the location of the band in each row can shift up to a couple of cells to the right relative to the band location of the previous row.
    __________________________                            __________________________----
        ___________________________           =>       __________________________----
            ____________________________                 __________________________----          

 In the banded score matrix, for processing elements toward the end of a row, we need to access to its upper adjacent cells, which due to shifting, if not padded with extra length, it will be invalid memory access. So either we can use if statement for those end cells which results in branch divergence, or pad the score matrix with some extra length.

pyclaragenomics/claragenomics/bindings/cudapoa.pyx
r-mafi: done

cudapoa/include/claragenomics/cudapoa/batch.hpp
r-mafi: sure, done.

r-mafi:
rerun tests

r-mafi:
rerun tests

cudapoa/src/cudapoa_nw_banded.cu
tijyojwad: Thanks for the explanation, @r-mafi !

PR #455, opened 2020-06-11
[cudamapper] Not storing nullptr to host index cache when skip_copy_to_host is set

mimaric:
When `skip_copy_to_host` is set pointer to host copy is null and should not be saved to cache.

Part of #318


cudamapper/src/index_cache.cu
```diff
@@ -159,19 +159,22 @@ void IndexCacheHost::generate_cache_content(const std::vector<IndexDescriptor>&
...
-        assert(nullptr != index_copy);
-
         // save pointer to cached index
-        new_cache[descriptor_of_index_to_cache] = index_copy;
+        if (!skip_copy_to_host)
+        {
+            assert(nullptr != index_on_host);
+            new_cache[descriptor_of_index_to_cache] = index_on_host;
+        }
+
```
ahehn-nv, line 177: Just to be sure: If `skip_copy_to_host` is set it means you process the data on the device immediately and you will not need the data after that operation anymore. Is that correct?
It's essentially the special case for the first index, which you use immediately, right?

mimaric: I believe you're right, but let me explain it in another way.

There are `keep_on_device` and `skip_copy_to_host`.
`keep_on_device` is determined for each index separately, depending on whether the index in present in `descriptors_of_indices_to_keep_on_device`. This is the list of indices that are going to be requested immediately after the cache is built, but are also likely going to be requested again, so they are both kept on device and copied to host.
`skip_copy_to_host` is used used when it is known that the whole index is only going to be requested once, e.g. when dealing with small genomes or when the ratio between host and device index sizes is such that in the end only a few indices fall into a batch.

ahehn-nv: Ok, makes sense. Thanks!

PR #463, opened 2020-06-15
[cudamapper] Support from multiple associated streams

### LINKED ISSUE #414: [cga] Support multiple streams in device_buffer and DevicePreallocatedAllocator
`DevicePreallocatedAllocator::DeviceAllocate()` currently accept `cudaStream_t`. `DevicePreallocatedAllocator::DeviceFree()` waits on that stream before actually deallocating memory in order to prevent the memory from being deallocated before all work that uses it is done.
In cases when the same buffer is used by two or more streams it is not possible to wait on all of them and the user has to make sure that all other streams have finished.

The goal it to support associating one allocation with multiple streams.

This should also be supported by `device_buffer`/`buffer`.

Consider using variadic template to pass one or more streams, store them in a `vector` internally and loop over them in `DevicePreallocatedAllocator::DeviceFree()`

This issue is part of issue #318

mimaric:
`device_buffer` and `DevicePreallocatedAllocator` now support multiple associated streams.

Closes #414 
Part of #318


common/base/include/claragenomics/utils/allocator.hpp
```diff
@@ -230,19 +230,21 @@ class CachingDeviceAllocator
 
     /// \brief asynchronously allocates a device array with enough space for n elements of value_type
     /// \param n number of elements to allocate the array for
-    /// \param stream CUDA stream to be associated with this method
+    /// \param streams CUDA streams to be associated with this memory allocation, when memory allocation is deallocated cudaStreamSynchronize() is called on all of them
     /// \return pointer to allocated memory
     /// \throw device_memory_allocation_exception if allocation was not successful
-    pointer allocate(std::size_t n, cudaStream_t stream = 0)
+    pointer allocate(std::size_t n, const std::vector<cudaStream_t>& streams = {{0}})
```
ahehn-nv, line 236: I think we don't need to provide a default argument for the stream parameter here. This should be done further up. Furthermore, the CudaMallocAllocator and this class should have the same interface and the CudaMallocAllocator does not provide a default argument either.

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
```diff
@@ -78,14 +82,16 @@ public:
     /// Memory allocation is aligned by 256 bytes
     /// \param ptr on return pointer to allocated memory, nullptr if allocation was not successful
     /// \param bytes_needed
-    /// \param associated_stream on deallocation this block will be free only once all work in this stream has finished
+    /// \param associated_streams on deallocation this block will be freed only once all previosly scheduled work in these streams has finished, using default stream as default
     /// \return cudaSuccess if allocation was successful, cudaErrorMemoryAllocation otherwise
     cudaError_t DeviceAllocate(void** ptr,
                                size_t bytes_needed,
-                               cudaStream_t associated_stream = 0)
+                               const std::vector<cudaStream_t>& associated_streams = {{0}})
```
ahehn-nv, line 89: Do we really need to provide the default argument for any interface compatibility here? Otherwise I suggest to drop it and let the caller handle this explicitly.

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
```diff
@@ -78,14 +82,16 @@ public:
...
     {
+        assert(!associated_streams.empty());
+
         std::lock_guard<std::mutex> mutex_lock_guard(memory_operation_mutex_);
-        return get_free_block(ptr, bytes_needed, associated_stream);
+        return get_free_block(ptr, bytes_needed, associated_streams);
```
ahehn-nv, line 94: Optional/off-topic: Now reading the code, I stumbled over `get_free_block`. How about renaming the function to `allocate_free_block`?

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
```diff
@@ -221,8 +227,11 @@ private:
                                                    });
         assert(block_to_be_freed_iter != std::end(used_blocks_));
 
-        // ** wait for all work on associated_stream to finish before freeing up this memory block
-        CGA_CU_ABORT_ON_ERR(cudaStreamSynchronize(block_to_be_freed_iter->associated_stream));
+        // ** wait for all work on associated_streams to finish before freeing up this memory block
+        for (const cudaStream_t& associated_stream : block_to_be_freed_iter->associated_streams)
```
ahehn-nv, line 231: Optional: Just use `cudaStream_t`, since it is only a tiny handle. Furthermore, doing a Synchronize on a `const` argument is a bit confusing imo.

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
```diff
@@ -221,8 +227,11 @@ private:
                                                    });
         assert(block_to_be_freed_iter != std::end(used_blocks_));
 
-        // ** wait for all work on associated_stream to finish before freeing up this memory block
-        CGA_CU_ABORT_ON_ERR(cudaStreamSynchronize(block_to_be_freed_iter->associated_stream));
+        // ** wait for all work on associated_streams to finish before freeing up this memory block
```
ahehn-nv, line 230: How about adding a comment like "Warning: Please do not rely on this cudaStreamSynchronize in the caller. The allocator implementation may guarantee the lifetime of the memory block without this implicit synchronization in the future."?

common/base/include/claragenomics/utils/device_buffer.hpp
```diff
@@ -140,9 +151,9 @@ class buffer
     /// \return begin() + size().
     const_iterator end() const { return data_ + size_; }
 
-    /// \brief This method returns the associated stream.
+    /// \brief This method returns the associated streams.
     /// \return associated stream.
-    cudaStream_t get_stream() const { return stream_; }
+    const std::vector<cudaStream_t>& get_streams() const { return streams_; }
```
ahehn-nv, line 156: Is this function actually used?

common/base/include/claragenomics/utils/device_buffer.hpp
```diff
@@ -187,14 +198,14 @@ class buffer
...
-        swap(a.stream_, b.stream_);
+        swap(a.streams_, b.streams_);
         swap(a.allocator_, b.allocator_);
     }
 
 private:
     value_type* data_;
     size_type size_;
-    cudaStream_t stream_;
+    std::vector<cudaStream_t> streams_;
```
ahehn-nv, line 208: It is a bit redundant to have the associated streams here and in the allocator.
From what I understand this is only needed because of the `clear_and_resize()` and the `get_streams` methods, correct?
Wouldn't it be clearer if the `clear_and_resize()` method took the (potentially new) associated streams explicitly?

The only problematic case would be if `new_size == size_`. There you would need to add all new stream arguments to the memory block. Probably not better than the current solution, wanted to discuss it real quick nevertheless.

common/base/include/claragenomics/utils/device_buffer.hpp
```diff
@@ -52,32 +54,41 @@ class buffer
     /// \brief This constructor creates \p buffer with the given size.
     /// \param n The number of elements to create
     /// \param allocator The allocator to use by this buffer.
-    /// \param stream The CUDA stream to be associated with this allocation. Default is stream 0.
+    /// \param streams CUDA streams to be associated with this buffer. Freeing the memory (either through destructor, free() or clear_and_resize()) calls cudaStreamSynchronize() on all those streams. If no stream is specified default stream (0) is used.
```
ahehn-nv, line 57: Please don't document the "calls cudaStreamSynchronize" - we might want to change that at some point.
Instead you could write "the device memory provided by this buffer is guaranteed to live until all operations on the associated stream have completed."

common/base/include/claragenomics/utils/device_buffer.hpp
```diff
@@ -52,32 +54,41 @@ class buffer
...
         {
-            data_ = allocator_.allocate(size_, stream_);
+            data_ = allocator_.allocate(size_, streams_);
         }
     }
 
     /// \brief This constructor creates an empty \p buffer.
     /// \param allocator The allocator to use by this buffer.
-    /// \param stream The CUDA stream to be associated with this allocation. Default is stream 0.
+    /// \param streams CUDA streams to be associated with this buffer. Freeing the memory (either through destructor, free() or clear_and_resize()) calls cudaStreamSynchronize() on all those streams. If no stream is specified default stream (0) is used.
```
ahehn-nv, line 85: see above.

common/base/include/claragenomics/utils/allocator.hpp
mimaric: We do need it in order to be able to use it in thurst. Thrust calls custom allocator's `allocate(std::size n)`. As it does not provide any stream it should sync on the default stream.

This means actual bug actually me not including default value in `CudaMallocAllocator` as well

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
mimaric: What about `allocate_memory_block()` and `free_memory_block`.
For me it is more confusing that we have `get_free_block()` and `free_block()`, when one `free` is adjective and the verb

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
mimaric: Makes sense, I think we can be rather sure `cudaSream_t` won't explode in the future.
done

common/base/include/claragenomics/utils/device_buffer.hpp
mimaric: No, removed

common/base/include/claragenomics/utils/device_buffer.hpp
mimaric: I would not go for explicitly passing streams again. It could be that resizing is done in a subfunction, in that case the user would have to pass the stream explicitly

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
mimaric: done

common/base/include/claragenomics/utils/device_buffer.hpp
mimaric: done

common/base/include/claragenomics/utils/device_buffer.hpp
mimaric: done

mimaric:
rerun tests

mimaric:
rerun tests

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
ahehn-nv: Sounds good to me.

common/base/include/claragenomics/utils/device_buffer.hpp
ahehn-nv: Ok, let's leave it as it is then.

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
ahehn-nv: @mimaric

common/base/include/claraparabricks/genomeworks/utils/allocator.hpp
```diff
@@ -251,7 +255,7 @@ class CachingDeviceAllocator
         return static_cast<pointer>(ptr);
     }
 
-    /// \brief Asynchronously dealllocates allocated array
+    /// \brief Asynchronously dealllocates allocated array, calls cudaStreamSynchronize() on associated streams
```
ahehn-nv, line 258: How about changing this to a very vague "may call cudaStreamSynchronize()"?

mimaric:
rerun tests

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
mimaric: Hmm, yes, here it makes sense to drop the default argument. I thought it's the same as the comment above, but unlike `CachingDeviceAllocator` `DevicePreallocatedAllocator` does not get called by anyone else.

common/base/include/claraparabricks/genomeworks/utils/allocator.hpp
mimaric: yes, I missed this one

mimaric: done

common/base/include/claragenomics/utils/device_preallocated_allocator.cuh
mimaric: done

tijyojwad:
@mimaric following recommendation from @raydouglass , reran tests. but also updated branch with changes from master.

PR #464, opened 2020-06-15
[cudapoa] Multi-batch binning

r-mafi:
Added a new function in `cudapoa/utils.hpp` where groups input windows based on their sizes in different batch-size bins. Windows with similar range of size are grouped together and launched with the same batch-size.
This multi-batch strategy can have a significant impact on scenarios where variance of window size is pretty high. If all windows have a similar size, multi-batch strategy will transform into a single batch launch.

Additionally:
- added an example of using multi-batch binning API in `sample_cudapoa`, to find a small set of batches to cover all windows while maintaining high GPU occupancy
- changed default band-width from 128 to 256
- slightly changed `generate_random_sequence()` in `genmoeutils.hpp` to make it more robust


r-mafi:
@tijyojwad the previous PR #427 was way out of date and merging new changes caused problems, moved the changes to this new one.

r-mafi:
rerun tests

r-mafi:
rerun tests

cudapoa/include/claragenomics/cudapoa/utils.hpp
```diff
@@ -0,0 +1,57 @@
...
+* and proprietary rights in and to this software, related documentation
+* and any modifications thereto.  Any use, reproduction, disclosure or
+* distribution of this software and related documentation without an express
+* license agreement from NVIDIA CORPORATION is strictly prohibited.
+*/
+
+#pragma once
+
+#include <vector>
+#include "batch.hpp"
```
tijyojwad, line 14: let's make this `#include <claragenomics/cudapoa/batch.hpp>`

cudapoa/src/utils.cu
```diff
@@ -0,0 +1,159 @@
+/*
+* Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
tijyojwad, line 2: copyright 2020

cudapoa/src/utils.cu
```diff
@@ -0,0 +1,159 @@
+/*
```
tijyojwad, line 1: can this file be called `utils.cpp`? or does it need to be compiled with `nvcc`?

cudapoa/src/utils.cu
```diff
@@ -0,0 +1,159 @@
...
+namespace claraparabricks
+{
+
+namespace genomeworks
+{
+
+namespace cudapoa
+{
+
+/// \brief Create a small set of batch-sizes to increase GPU parallelization.
```
tijyojwad, line 23: the implementation file shouldn't need doxygen docs as long as the public header has it (which I believe it does in `utils.hpp`)

cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -190,113 +173,130 @@ int main(int argc, char** argv)
...
-        std::vector<StatusType> seq_status;
-        StatusType status = batch->add_poa_group(seq_status, poa_group);
+    // analyze the POA groups and create a minimal set of batches to process them all
+    std::vector<BatchSize> list_of_batch_sizes;
+    std::vector<std::vector<int32_t>> list_of_groups_per_batch;
 
-        // NOTE: If number of windows smaller than batch capacity, then run POA generation
-        // once last window is added to batch.
-        if (status == StatusType::exceeded_maximum_poas || (i == get_size(windows) - 1))
+    cudapoa::get_multi_batch_sizes(list_of_batch_sizes, list_of_groups_per_batch, poa_groups, banded, msa);
```
tijyojwad, line 212: since we're using the `cudapoa` namespace, we don't need to explicitly add `cudapoa::get_multi_batch_sizes`. It can just be `get_multi_batch_sizes`

cudapoa/src/utils.cu
r-mafi: good catch, I just copied and forgot to remove these comments.

cudapoa/src/utils.cu
r-mafi: no, initially it was .cpp that I was getting some unrelated compilation errors. Because it includes `allocate_block.hpp` needs `nvcc`

cudapoa/include/claragenomics/cudapoa/utils.hpp
r-mafi: done.

cudapoa/src/utils.cu
r-mafi: done.

cudapoa/samples/sample_cudapoa.cpp
r-mafi: that's right, fixed.

PR #467, opened 2020-06-16
[sdk] update include paths to use claraparabricks

tijyojwad:
Addresses part of #424


tijyojwad:
rerun tests

cudamapper/tests/Test_CudamapperOverlapper.cpp
```diff
@@ -11,7 +11,7 @@
 #include "gtest/gtest.h"
 #include <string>
 #include <vector>
-#include "../include/claragenomics/cudamapper/overlapper.hpp"
```
edawson, line 14: Is our convention that we take global includes over local ones? These two includes have distinctly different meanings if you've got both a global install (e.g. in /usr/local) and a local copy of the repo.

tijyojwad: the cmake setup adds the public include paths to the compiler, so we use the headers using this convention as they get picked up from the include paths.

for global vs local copy, I'd say in those situations the include lists should be setup to have only one of them, and not both. otherwise it gets pretty confusing which one is to use where and a potential recipe for nasty bugs...

r-mafi:
@tijyojwad searching for `claragenomics` I found few instances in python files such as `setup.py`. Do they need to change as well? otherwise LGTM.

tijyojwad:
> 
> 
> @tijyojwad searching for `claragenomics` I found few instances in python files such as `setup.py`. Do they need to change as well? otherwise LGTM.

Yep, the whole package name needs to be updated still. That will be in the next PR :).

PR #472, opened 2020-06-24
[cudapoa] CUDAPOA Binary

atadkase:
- Used cudamapper bin src as a template to create bin for cudapoa. Used cudapoa src from sample app.
- Moved benchmark utils to common utils folder in cudapoa as the parsing function is  required by multiple files.
- Added code to parse multiple fasta files.
- Updated respective cmake files


cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,168 @@
...
+
+    if (gpu_mem_allocation <= 0 || gpu_mem_allocation > 1.0)
+    {
+        std::cerr << "gpu-mem-alloc should be greater than 0 and less than 1.0" << std::endl;
+        exit(1);
+    }
+
+    if (consensus_mode < 0 || consensus_mode > 1)
+    {
+        std::cerr << "consensus_mode can only be 0 (consensus) or 1 (msa)" << std::endl;
```
r-mafi, line 89: the naming of `consensus_mode` may imply MSA is a mode of consensus, while MSA and Consensus are two possible outputs of cudaPOA analysis, it can be changed to analysis or output-type.

cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,168 @@
...
+            consensus(0)/msa(1) [0])"
+              << R"(
+        -f, --full-alignment
+            uses full alignment if this flag is passed [banded alignment])"
+              << R"(
+        -g, --graph-output
+            output path for printing graph in DOT format [disabled])"
+              << R"(
+        -n, --max-windows
+            maximum number of windows to use from file)"
```
r-mafi, line 152: not sure in documentation we use POA group or windows, they are closely related and within the code are used interchangeably, just wanted to ensure term window is defined in the docs.

cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,168 @@
...
+              << R"(
+        -r, --gpu-mem-alloc
+            fraction of GPU memory to be used for cudapoa [0.9])"
```
r-mafi, line 155: to be more precise, it is fraction of "available GPU memory", not to be confused with total mem.

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,277 @@
...
+    cudaMemGetInfo(&free, &total);
+
+    // Initialize internal logging framework.
+    Init();
+
+    // Initialize CUDAPOA batch object for batched processing of POAs on the GPU.
+    const int32_t device_id      = 0;
+    cudaStream_t stream          = 0;
+    size_t mem_per_batch         = gpu_mem_allocation * free; // Using 90% of GPU available memory for CUDAPOA batch.
+    const int32_t mismatch_score = -6, gap_score = -8, match_score = 8;
```
r-mafi, line 44: I forgot about the scores, they may also need to be defined in the input, with the current default values.

r-mafi: with checks about the sign. _gap_ and _mismatch_ should be negative, _match_ should be positive.

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,277 @@
...
+
+                    // After MSA/consensus is generated for batch, reset batch to make room for next set of POA groups.
+                    batch->reset();
+
+                    // In case that number of batch groups is more than the capacity available on GPU, the for loop breaks into smaller number of groups.
+                    // if adding group i in batch->add_poa_group is not successful, it wont be processed in this iteration, therefore we print i-1
+                    // to account for the fact that group i was excluded at this round.
+                    if (status == StatusType::success)
+                    {
+                        std::cout << "Processed groups " << group_count + group_count_offset << " - " << i + group_count_offset << " (batch " << b << ")" << std::endl;
```
r-mafi, line 223: we probably have to keep these log-type output as `cerr` not to be mixed with printing output data.

cudapoa/src/main.cpp
atadkase: Are 0s okay?

r-mafi: as long as mismatch and gap scores are less than match score, it should be fine. Additionally, in some places in the code, such as `use32bitScore()`, there is an underlying assumption that match score is non-negative and gap/mismatch score are non-positive. So 0 should be OK.

cudapoa/src/main.cpp
atadkase: Done.

cudapoa/src/main.cpp
atadkase: Done.

cudapoa/src/application_parameters.cpp
atadkase: Done.

cudapoa/src/application_parameters.cpp
atadkase: Done. Renamed to result

cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,225 @@
...
+        exit(1);
+    }
+}
+
+void ApplicationParameters::help(int32_t exit_code)
+{
+    std::cerr <<
+        R"(Usage: cudapoa [options ...]
+     options:
+        -i, --input <file>
```
atadkase, line 181: Not sure if I should leave this as input with `-i` or whether I should interpret any following strings as names for sequence files

cudapoa/src/application_parameters.cpp
atadkase: Changed to groups.

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -50,6 +57,136 @@ void get_multi_batch_sizes(std::vector<BatchSize>& list_of_batch_sizes,
...
+        num_reads_per_file[i] = fasta_parser_vec[i]->get_num_seqences();
+    }
+    const int32_t num_reads = num_reads_per_file[0];
+
+    for (int32_t i = 1; i < num_input_files; i++)
+    {
+        if (num_reads_per_file[i] != num_reads)
+        {
+            std::cerr << "Failed to read input files." << std::endl;
+            std::cerr << "Number of long-reads per input file do not match with each other." << std::endl;
```
tijyojwad, line 137: this is actually not strictly required. in the case of the bonito dataset we needed the groups to be of the same size, but in general each group could be different

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -50,6 +57,136 @@ void get_multi_batch_sizes(std::vector<BatchSize>& list_of_batch_sizes,
...
+    int32_t idx = 0;
+    for (auto& window : windows)
+    {
+        for (int32_t i = 0; i < num_input_files; i++)
+        {
+            window.push_back(fasta_parser_vec[i]->get_sequence_by_id(idx).seq);
+        }
+        idx++;
+    }
+    if (total_windows >= 0)
```
tijyojwad, line 153: I think this part of the code is shared between the `parse_window_data_file` and `parse_fasta_windows` - let's move into a common function?

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -50,6 +57,136 @@ void get_multi_batch_sizes(std::vector<BatchSize>& list_of_batch_sizes,
                            int32_t gap_score                   = -8,
                            int32_t match_score                 = 8);
 
+/// \brief Parses window data file
```
tijyojwad, line 60: I think it would be good to have an example of this data file. Can you add it to the description?
```
<num_sequences_in_group_0>
group0_seq0
group0_seq1
group0_seq2
...
...
<num_sequences_in_group_1>
group1_seq0
group1_seq1
group1_seq2
...
...
```

cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,225 @@
...
+{
+
+namespace cudapoa
+{
+
+ApplicationParameters::ApplicationParameters(int argc, char* argv[])
+{
+    struct option options[] = {
+        {"input", required_argument, 0, 'i'},
+        {"result", required_argument, 0, 'r'},
```
tijyojwad, line 34: Instead of result, let's make this an argument-less option to output "msa or not". by default we can output consensus

cudapoa/src/application_parameters.cpp
tijyojwad: If multiples files, okay to have this syntax `cudapoa -i file1.fa -i file2.fa -i file3.fa ...`

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+namespace claraparabricks
+{
+
+namespace genomeworks
+{
+
+namespace cudapoa
+{
+
+std::unique_ptr<Batch> initialize_batch(int32_t mismatch_score, int32_t gap_score, int32_t match_score, bool msa, bool banded_alignment, const double gpu_mem_allocation, const BatchSize& batch_size)
```
tijyojwad, line 26: let's make this multi line

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+int main(int argc, char* argv[])
+{
+    // Parse input parameters
+    const ApplicationParameters parameters(argc, argv);
+
+    // Load input data. Each window is represented as a vector of strings. The sample
+    // data has many such windows to process, hence the data is loaded into a vector
+    // of vector of strings.
+    std::vector<std::vector<std::string>> windows;
```
tijyojwad, line 131: as discussed earlier, let's rename `windows` to `groups`

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+            poa_entry.weights = nullptr;
+            group.push_back(poa_entry);
+        }
+    }
+
+    // analyze the POA groups and create a minimal set of batches to process them all
+    std::vector<BatchSize> list_of_batch_sizes;
+    std::vector<std::vector<int32_t>> list_of_groups_per_batch;
+
+    get_multi_batch_sizes(list_of_batch_sizes, list_of_groups_per_batch, poa_groups, parameters.banded, parameters.result == 1, parameters.band_width, nullptr, parameters.gpu_mem_allocation, parameters.mismatch_score, parameters.gap_score, parameters.match_score);
```
tijyojwad, line 172: let's make this multi line

cudapoa/src/main.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+    int32_t group_count_offset = 0;
+
+    for (int32_t b = 0; b < get_size(list_of_batch_sizes); b++)
+    {
+        auto& batch_size      = list_of_batch_sizes[b];
+        auto& batch_group_ids = list_of_groups_per_batch[b];
+
+        // Initialize batch.
+        std::unique_ptr<Batch> batch = initialize_batch(parameters.mismatch_score, parameters.gap_score, parameters.match_score, parameters.result == 1, parameters.banded, parameters.gpu_mem_allocation, batch_size);
```
tijyojwad, line 182: multi line

cudapoa/src/main.cpp
r-mafi: just to clarify a bit: from the changed files view it's not obvious, but in the `main.cpp`, file parser stores data in a vector of `strings` named `windows`, after that converts them into an array of `groups`. Regarding the interface, I see they are already changed to `groups`, e.g. [https://github.com/atadkase/ClaraGenomicsAnalysis/blob/atadkase/cudapoa-bin/cudapoa/src/application_parameters.cpp#L196](url)

cudapoa/src/main.cpp
atadkase: I was expecting the autoformatter to do this :sweat_smile: Let me go fix this.

cudapoa/src/main.cpp
atadkase: Yeah. I renamed the external interfaces to groups. Didn't percolate it through the rest of the code though.

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
atadkase: Changed it to adapt to whatever input is provided. Might result in end windows having lesser reads than initial windows.

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
atadkase: Moved

cudapoa/src/application_parameters.cpp
atadkase: This allows us to support mode 2 in the future, which can be to run both. Are we not going to have that mode later on?

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
atadkase: Added. Renamed to windows to keep external interface name consistent.

cudapoa/src/main.cpp
atadkase: Done.

cudapoa/src/application_parameters.cpp
tijyojwad: Indeed, and that's what the module API supports as well. But we haven't seen any use case yet where a user wants to run both, and I'm not sure if that's tested properly either ;). So let's make it more restricted for now, and open up later if we get a feature request.

cudapoa/src/application_parameters.cpp
```diff
@@ -0,0 +1,226 @@
...
+        case 'h':
+            help(0);
+        default:
+            exit(1);
+        }
+    }
+
+    if (gpu_mem_allocation <= 0 || gpu_mem_allocation > 1.0)
+    {
+        std::cerr << "gpu-mem-alloc should be greater than 0 and less than or equal to 1.0" << std::endl;
```
tijyojwad, line 95: looking through this again - let's throw a `runtime_error`exception in these cases?

cudapoa/src/main.cpp
tijyojwad: Okay perfect, thanks!

cudapoa/src/main.cpp
tijyojwad: I will look into why the formatter didn't take care of this...

cudapoa/src/application_parameters.cpp
atadkase: Sounds good. I've changed it to use the flag `-a` since the rest were already taken

cudapoa/src/application_parameters.cpp
atadkase: Changed for nearly everything except the input validation as it also prints the help message when 0 or invalid input is provided (useful to lookup usage).

tijyojwad: Perfect, thanks @atadkase !

PR #474, opened 2020-06-24
WIP - [pygw] fix intermittent test bug

tijyojwad:
issue appears to be with how we were handling the cuda aligner tests in python. the cleanup wasn't correct, which was leading to out of order destruction in pytest. using pytest fixture ensures the objects are deleted properly after each test.

temporarily resolves intermittent bug in tests

### DIFF
--- a/pygenomeworks/test/test_cudaaligner_bindings.py
+++ b/pygenomeworks/test/test_cudaaligner_bindings.py
@@ -16,6 +16,13 @@ from genomeworks.simulators.genomesim import PoissonGenomeSimulator
 from genomeworks.simulators.readsim import NoisyReadSimulator
 
 
+@pytest.fixture
+def get_stream():
+    stream = cuda.CudaStream()
+    yield stream
+    del stream
+
+
 @pytest.mark.gpu
 @pytest.mark.parametrize("query, target, cigar", [
     ("AAAAAAA", "TTTTTTT", "7M"),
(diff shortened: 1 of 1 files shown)


tijyojwad:
nvm @ahehn-nv @mimaric , bug not fixed yet. false alarm :)

ohadmo:
:clap: 
@tijyojwad Maybe we should move this fixture to a `conftest.py` file,  so it can be shared around different test files, we allocate a CudaStream() also in `test_cudapoa_bindings.py`

ahehn-nv:
Another solution is to make python somehow aware of CudaAlignerBatch's reference to the CudaStream, such that the garbage collection cannot delete the CudaStream before the C++ object behind CudaAlignerBatch.

ahehn-nv:
Thinking about it again, the approach with the CudaStream C++ object will probably not help (but would be good for consistency nevertheless), since python manages all the lifetime and calls to special functions. It would be surprising if it behaved differently just because it is a wrapped C++ object. So we should find a way to keep a reference to CudaStream until all destructors of CudaAlignerBatch have been called.

tijyojwad:
Yes, this is a temporary fix to the problem where we're explicitly keeping the CudaStream object longer than the CudaAlignerBatch. The solution is to keep the reference for CudaStream in the batch object around till destruction, and we do have some code for that, but it's not done right. The reference counting isn't happening correctly for the CudaStream object, and it its __dealloc__ gets called in arbitrary order in the test failure case. I have some leads for it, will create separate PR to address it. Merging this now.

And agreed #475 will help with consistency, but not his bug.

PR #475, opened 2020-06-25
[common] Introduced CudaStream an RAII class to wrap cuda streams

### LINKED ISSUE #393: [common] Wrapping CUDA stream in another class to catch stream destruction

ahehn-nv:
fixes #393

### DIFF
--- a/common/base/include/claraparabricks/genomeworks/utils/cudautils.hpp
+++ b/common/base/include/claraparabricks/genomeworks/utils/cudautils.hpp
@@ -24,6 +24,7 @@
 #include <cuda_runtime_api.h>
 #include <cassert>
 #include <string>
+#include <memory>
 
 #ifdef GW_PROFILING
 #include <nvToolsExt.h>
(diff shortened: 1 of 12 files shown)


pygenomeworks/genomeworks/cuda/cuda.pyx
```diff
@@ -32,40 +33,19 @@ cdef class CudaStream:
...
-            raise CudaRuntimeError(e)
-        self.stream = <size_t>s
+    cdef cuda_stream.CudaStream stream
 
     def __init__(self):
-        """Dummy implementation of __init__ function to allow
-        for Python subclassing.
+        """Constructs a CudaStream object to encapsulate a CUDA stream.
         """
+        self.stream = cuda_stream.make_cuda_stream()
```
tijyojwad, line 42: this needs to be in the `__cinit__` function. That's when all C objects need to be initialized

an empty `__init__ object also needs to be kept around to allow for python subclassing

pygenomeworks/genomeworks/cudaaligner/cudaaligner.pyx
```diff
@@ -122,7 +122,6 @@ cdef class CudaAlignerBatch:
     """Python API for CUDA-accelerated sequence to sequence alignment.
     """
     cdef unique_ptr[cudaaligner.Aligner] aligner
-    cdef public object stream
```
tijyojwad, line 125: why was this removed? I think there's a reference to this below?

ahehn-nv: Because I would like this class to use the python object, not the c++ object.

tijyojwad: Does the `self.stream = stream` line in `__cinit__` work if the `stream` variable is not `cdef`ed with some type?

pygenomeworks/genomeworks/cuda/cuda.pyx
ahehn-nv: I removed all the python changes from this PR.

pygenomeworks/genomeworks/cudaaligner/cudaaligner.pyx
ahehn-nv: I removed all the python changes from this PR.

mimaric (approved):

PR #477, opened 2020-06-28
[pygw] better fix for intermittent test failure

tijyojwad:
The problem is that in some cases the `CudaStream` object is destroyed before the `CudaAlignerBatch` object. Keeping a reference to the stream object should resolve this issue since python does reference counting, but through tracking reference counts it appears that the reference is getting lost somewhere. The issue is resolve by explicitly increasing and decreasing the ref count for stream within the cudaaligner and cudapoa batch objects.

It's a more permanent fix, but still not sure why the python ref counting mechanism appears to be broken. Tracking this in https://github.com/clara-parabricks/GenomeWorks/issues/476


raydouglass:
rerun tests

tijyojwad:
rerun tests

ahehn-nv (commented):
Looks good to me.
However, I'm not a (c)python expert. @ohadmo could you have a look too?

pygenomeworks/genomeworks/cudaaligner/cudaaligner.pyx
```diff
@@ -129,18 +130,18 @@ cdef class CudaAlignerBatch:
...
     cdef unique_ptr[cudaaligner.Aligner] aligner
-    cdef public object stream
+    cdef public CudaStream stream
 
     def __cinit__(
             self,
             max_query_length,
             max_target_length,
             max_alignments,
+            *args,
```
o8191, line 140: @tijyojwad Having `*args` as an argument before any key argument can lead to a bug. This way, the key-arguments after will only be able to be set by explicitly setting them when calling the function, since all the remaining positional arguments will be collected by `*args` parameter.
For instance:
```
def foo(a, b, *args, c=-1, **kargs):
    print(args)
    print(c)
    print(kargs)


if __name__ == '__main__':
    foo(1, 2, 3, d=5, e=8)
```
Output
```
(3,)
-1
{'d': 5, 'e': 8}
```
Compared to:
```
def foo(a, b, c=-1, *args, **kargs):
    print(args)
    print(c)
    print(kargs)


if __name__ == '__main__':
    foo(1, 2, 3, d=5, e=8)
```
Ouput:
```
()
3
{'d': 5, 'e': 8}
```

pygenomeworks/genomeworks/cudaaligner/cudaaligner.pyx
```diff
@@ -187,11 +194,11 @@ cdef class CudaAlignerBatch:
...
```
o8191, line 197: Same as the previous comment

pygenomeworks/genomeworks/cudaaligner/cudaaligner.pyx
tijyojwad: Good catch @o8191 ! Fixed. Can you review?

o8191 (approved):

Issue #471, opened 2020-06-24
[cudamapper] Default parameters should be hardened against many datasets / GPUs, match those of minimap2


edawson:
Cudamapper is supposed to be a drop-in replacement for minimap2. Our default parameters, however, differ from those of minimap2. In addition, they seem to be unstable on many GPUs, causing crashes.

This is a major barrier to entry for users - they want the program to run to completion, even at the expense of performance.

Minimap2 defaults:
- [x] Minimizer kmer size: 15
- [x]  Minimizer window size: 10
- [x] Minimum number of minimizers in a chain: 3
- [x] Max distance between minimizers before chain is terminated: 5000
- [x] Minimum overlap size: <500
- [x] Minibatch size: 500M

Most of these can be addressed by either a single cudamapper parameter or a combination of multiple parameters. 

In the case of minibatch size, it's not so much about matching the number as it is about providing a stable CLI. I find I'm often having to tweak the `-I, -i, -q, -Q, -c, -C, -m` parameters to balance memory usage (e.g. to prevent out-of-memory errors) and performance. I think we should establish safe defaults for long reads on 8GB, 16GB and 32GB memory GPUs. Even though we programatically check for max preallocated memory we often seem to OOM due to index size parameterizations. My vote would be to prioritize stability and provide a one-pager on tuning for max performance (acknowledging the limits of each GPU considering maximum read size).

edawson:
Here's an example where cudamapper fails with default parameters on a small (~65MB) FASTA file, likely because its longest read is >300 kilobases in length.
```
(base) eric@odin:~/sandbox/clara/build$ time ./cudamapper/cudamapper -k 15 -w 10 -m 6 ~/droso_split_1.fa ~/droso_split_1.fa > droso1.paf-C / --target-indices-in-host-memory not set, using -Q / --query-indices-in-host-memory value: 10
-c / --target-indices-in-device-memory not set, using -q / --query-indices-in-device-memory value: 5
NOTE - Since query and target files are same, activating all_to_all mode. Query index size used for both files.
Query file: /home/eric/droso_split_1.fa, number of reads: 9999
Target file: /home/eric/droso_split_1.fa, number of reads: 9999
Using device memory cache of 6442450944 bytes
Device 0 took batch 1 out of 1 batches in total
terminate called after throwing an instance of 'claraparabricks::genomeworks::device_memory_allocation_exception'
  what():  Could not allocate device memory!
Aborted (core dumped)

[...]

PR #478, opened 2020-06-30
[cudamapper] Change default parameters for -w, -F, -r, -b, and -z.

edawson:
Addresses changes requested in #471 . Changes the default values of the -F and -w parameters to be the same
as those of minimap2 for starters.


cudamapper/src/application_parameters.cpp
```diff
@@ -283,22 +283,22 @@ void ApplicationParameters::help(int32_t exit_code)
             length of batch sized used for target in MB [30])"
               << R"(
         -F, --filtering-parameter
-            filter all representations for which sketch_elements_with_that_representation/total_sketch_elements >= filtering_parameter), filtering disabled if filtering_parameter == 1.0 [1'000'000'001] (Min = 0.0, Max = 1.0))"
+            Remove representations with frequency (sketch_elements_with_that_representation/total_sketch_elements) >= filtering_parameter. Filtering is disabled if filtering_parameter == 1.0 [1'000'000'001] (Min = 0.0, Max = 1.0, default=1e-5))"
```
mimaric, line 286: I know this is my code, but even I don't remember what's `[query_indices_being_generated]`, so I would suggest removing it.
As in other lines `[]` is used for default value I would use it here as well

cudamapper/src/application_parameters.hpp
```diff
@@ -36,17 +36,17 @@ class ApplicationParameters
...
     int32_t index_size                      = 30;    // i
     int32_t target_index_size               = 30;    // t
-    double filtering_parameter              = 1.0;   // F
+    double filtering_parameter              = 1e-5;  // F
     int32_t alignment_engines               = 0;     // a
-    int32_t min_residues                    = 100;   // r
-    int32_t min_overlap_len                 = 2000;  // l
-    int32_t min_bases_per_residue           = 100;   // b
-    float min_overlap_fraction              = 0.95;  // z
+    int32_t min_residues                    = 3;     // r, recommended range: 1 - 10. Higher: more accurate. Lower: more sensitive
```
mimaric, line 46: Maybe put this in `help()` function?

PR #482, opened 2020-07-02
[cudapoa] Fixes bug with multiple fasta logic

atadkase:
Fixes a bug where each read is considered a separate group in fasta input.

### DIFF
--- a/cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
+++ b/cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
@@ -145,12 +145,10 @@ inline void parse_fasta_windows(std::vector<std::vector<std::string>>& windows, 
     const int32_t num_input_files     = input_paths.size();
     std::vector<std::shared_ptr<io::FastaParser>> fasta_parser_vec(num_input_files);
     std::vector<int32_t> num_reads_per_file(num_input_files);
-    int32_t max_num_reads = 0;
     for (int32_t i = 0; i < num_input_files; i++)
     {
         fasta_parser_vec[i]   = io::create_kseq_fasta_parser(input_paths[i], min_sequence_length, false);
         num_reads_per_file[i] = fasta_parser_vec[i]->get_num_seqences();
-        max_num_reads         = std::max(max_num_reads, num_reads_per_file[i]);
     }
 
     windows.resize(num_input_files);
--- a/cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
+++ b/cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
@@ -153,20 +153,15 @@ inline void parse_fasta_windows(std::vector<std::vector<std::string>>& windows, 
         max_num_reads         = std::max(max_num_reads, num_reads_per_file[i]);
     }
 
-    windows.resize(max_num_reads);
-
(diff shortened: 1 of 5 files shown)


cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -145,27 +145,19 @@ inline void parse_fasta_windows(std::vector<std::vector<std::string>>& windows,
...
-            if (idx < num_reads_per_file[i])
-            {
-                window.push_back(fasta_parser_vec[i]->get_sequence_by_id(idx).seq);
-            }
+            windows[i].push_back(fasta_parser_vec[i]->get_sequence_by_id(idx).seq);
         }
-        idx++;
     }
 
     resize_windows(windows, total_windows);
```
r-mafi, line 157: it is not the number of windows that is subject to change, but the size of windows (number of reads per window). For example if a fasta file has 2000 reads, we may be interested in processing only first 1000 reads, which can be defined with option `-M`. Therefore following changes are required:
1- changing `--max-groups` to `--max-group-size` 
2- changing help description to "maximum number of reads per POA group(s) read from input file(s) (-1 for all, >0 for limited) [-1]"
3- changing `resize_windows()` function to resize each window
4- changing `total_windows` in `utils.hpp' to `max_number_of_reads` or `window_size`

tijyojwad: @r-mafi the original implementation of `resize_windows` was to change the number of windows itself. it was meant to make a few windows look like a large number of windows by looping over them again, or to just select a subset of the windows.

what you're asking for is another feature that's worth having as a utility.

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -143,31 +143,16 @@ inline void parse_fasta_windows(std::vector<std::vector<std::string>>& windows,
 {
```
r-mafi, line 143: nit-picky: to keep the parser function names consistent, what do you think of renaming `parse_window_data_file()` and `parse_fasta_windows()` to something like `parse_cudapoa_file()` and `parse_fasta_files()`?

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
r-mafi: hmm, I see. What I had suggested initially for option `-M` was number of reads, not windows. Not sure if the latter has any practical usage. I see that it could be useful in sample-cudapoa, to create larger sample input data, but as an option for cudapoa binary, not sure if it has any practical usefulness.

atadkase: That can be done. Can we put that feature in a separate PR?

r-mafi: we can address these changes in a separate PR, I agree it is not related to this bug fix. We can later discuss how we want to change the input options in the binary, `BatchSize` constructor args are also something to decide in the future revisions.

r-mafi: created an issue #483 to follow up on this.

PR #492, opened 2020-07-09
[SDK] Copy license to pygenomeworks folder before pip executes setup()

ohadmo:
Related to https://github.com/clara-parabricks/GenomeWorks/pull/488#issuecomment-655772836
Copying the correct license before setup() is executed.

Since we use `pip wheel` and not `python setup.py bdist_wheel` to create a wheel package, the entire `pygenomeworks` folder is copied to a temporary folder. Therefore, we can not rely on static symlinks/relative path.


ohadmo:
rerun tests

Issue #490, opened 2020-07-08
[cudamapper] Process skipped pairs of indices separately


mimaric:
Keep a list of pairs of indices which were skipped due to an OOM error (Issue #489). Once all other pairs of indices have been processed go over skipped ones and process them one by one without keeping any other indices on device in order to maximize the amount of available device memory.

PR #493, opened 2020-07-09
[cudamapper] Skip index pair which cause OOM errors

### LINKED ISSUE #489: [cudamapper] Skip pairs of indices which cause out of memory errors
Depending on read characteristics some pairs of indices can cause OOM errors.
Skip such pairs of indices and print a message to the user.

mimaric:
Some pairs of indices generate so much anchors that the application runs our of memory.
Application now skips such pairs and prints the number of skipped pairs so the user can decide if that number is acceptable or the application should be rerun with different parameters.

Closes #489 

Application does not print which pairs have been skipped as a PR for #490 will significantly reduce the number of overall pairs which actually get skipped. That PR will include actually printing skipped pairs of indices

### DIFF
--- a/cudamapper/src/main.cu
+++ b/cudamapper/src/main.cu
@@ -228,42 +228,49 @@ void process_one_device_batch(const IndexBatch& device_batch,
             {
                 std::shared_ptr<Index> query_index  = device_cache.get_index_from_query_cache(query_index_descriptor);
                 std::shared_ptr<Index> target_index = device_cache.get_index_from_target_cache(target_index_descriptor);
-
-                // find anchors and overlaps
-                auto matcher = Matcher::create_matcher(device_allocator,
-                                                       *query_index,
-                                                       *target_index,
-                                                       cuda_stream);
-
-                std::vector<Overlap> overlaps;
-                OverlapperTriggered overlapper(device_allocator,
-                                               cuda_stream);
-                overlapper.get_overlaps(overlaps,
-                                        matcher->anchors(),
-                                        application_parameters.min_residues,
-                                        application_parameters.min_overlap_len,
-                                        application_parameters.min_bases_per_residue,
-                                        application_parameters.min_overlap_fraction);
-
-                // free up memory taken by matcher
-                matcher.reset(nullptr);
-
-                // Align overlaps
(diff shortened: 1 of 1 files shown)


mimaric:
rerun tests

cudamapper/src/main.cu
```diff
@@ -228,42 +229,49 @@ void process_one_device_batch(const IndexBatch& device_batch,
...
+                                       cigars);
+                    }
+
+                    // pass overlaps and cigars to writer thread
+                    overlaps_and_cigars_to_process.add_new_element({std::move(overlaps), std::move(cigars)});
+                }
+                catch (device_memory_allocation_exception& oom_exception)
+                {
+                    // if the application ran out of memory skip this pair of indices
+                    ++(number_of_skipped_pairs_of_indices);
```
edawson, line 273: Do we need this whole section in the `try` clause? I thought the oom_exception was only thrown by create_matcher. My only concern is that this could catch other oom_exceptions beyond the one we intend to catch with this PR.

cudamapper/src/main.cu
```diff
@@ -275,13 +283,15 @@ void process_one_device_batch(const IndexBatch& device_batch,
 /// \param host_cache data will be loaded into cache within the function
 /// \param device_cache data will be loaded into cache within the function
 /// \param overlaps_and_cigars_to_process overlaps and cigars are output to this structure and the then consumed by another thread
+/// \param number_of_skipped_pairs_of_indices
```
edawson, line 286: Can we note that indices are skipped due to oom_exceptions being thrown?

cudamapper/src/main.cu
mimaric: Actually that was my intention. Device OOM can happen in matcher, overlapper and aligner and we would like to catch them all. It's actually all the same case: running out of memory while processing a pair of indices.
Does that make sense?

cudamapper/src/main.cu
mimaric: done

cudamapper/src/main.cu
edawson: Yep, that makes sense and is cleaner than individual try-catch clauses for each call.

PR #495, opened 2020-07-10
[cudapoa] Adaptive Banding - first draft

r-mafi:
CUDA implementation of adaptive banding


cudapoa/samples/sample_cudapoa.cpp
```diff
@@ -42,7 +42,7 @@ std::unique_ptr<Batch> initialize_batch(bool msa, bool banded_alignment, const B
     cudaStream_t stream          = 0;
     size_t mem_per_batch         = 0.9 * free; // Using 90% of GPU available memory for CUDAPOA batch.
     const int32_t mismatch_score = -6, gap_score = -8, match_score = 8;
-
+    std::cout << adaptive_banded;
```
atadkase, line 45: I'll remove this cout.

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -593,7 +667,13 @@ void generatePOA(genomeworks::cudapoa::OutputDetails* output_details_d,
...
                                                                     batch_size.max_nodes_per_window,
                                                                     batch_size.max_matrix_graph_dimension,
-                                                                    batch_size.max_consensus_size);
+                                                                    batch_size.max_consensus_size,
+                                                                    cuda_adaptive_banding,
+                                                                    band_starts,
+                                                                    band_widths,
+                                                                    head_indices,
+                                                                    max_indices);
+            //std::cout<<"HEAR YE: "<<cuda_adaptive_banding<<std::endl;
```
atadkase, line 676: Need to remove this too! :joy:

cudapoa/src/allocate_block.hpp
```diff
@@ -251,6 +263,8 @@ class BatchBlock
         offset_d_ += cudautils::align<int64_t, 8>(sizeof(*graph_details_d->sorted_poa) * max_nodes_per_window_ * max_poas_);
         graph_details_d->sorted_poa_node_map = reinterpret_cast<decltype(graph_details_d->sorted_poa_node_map)>(&block_data_d_[offset_d_]);
         offset_d_ += cudautils::align<int64_t, 8>(sizeof(*graph_details_d->sorted_poa_node_map) * max_nodes_per_window_ * max_poas_);
+        graph_details_d->node_distance_to_head = reinterpret_cast<decltype(graph_details_d->node_distance_to_head)>(&block_data_d_[offset_d_]);
```
atadkase, line 266: Should this be fenced with adaptive flag?

cudapoa/src/batch.cu
```diff
@@ -52,33 +54,36 @@ std::unique_ptr<Batch> create_batch(int32_t device_id,
...
-                                                                    cuda_banded_alignment);
+                                                                    cuda_banded_alignment,
+                                                                    cuda_adaptive_alignment);
         }
     }
     else
     {
         // if ScoreT is 16-bit, then it's safe to assume SizeT is 16-bit
-        return std::make_unique<CudapoaBatch<int16_t, int16_t>>(device_id,
+        return std::make_unique<CudapoaBatch<int32_t, int32_t>>(device_id,
```
atadkase, line 77: I think this carried over from my debugging changes.

r-mafi: good catch. fixed.

cudapoa/src/allocate_block.hpp
r-mafi: yes, this was added before the other arrays (band_start ...), that's why it's left separate, but no reason to keep it that way. fixed.
As a next step, another change in `allocate_block` is to consider banded-alignment storage for adaptive-banding and add a memory check in adaptive NW kernel to ensure that assumption is not violated, otherwise throw an error code.

cudapoa/src/application_parameters.cpp
```diff
@@ -106,6 +110,17 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
...
 
+    if (!banded && adaptive)
+    {
+        throw std::runtime_error("adaptive alignment cannot run with full alignment");
+    }
+
+    if (adaptive && default_bandwidth)
+    {
+        // this is used to determine total score matrix size in adaptive-banded alignment
+        band_width = 2048;
```
atadkase, line 121: We can remove this code if 256 is alright for bandwidth.

cudapoa/src/cudapoa_batch.cuh
```diff
@@ -625,6 +630,7 @@ protected:
 
     // Use banded POA alignment
     bool banded_alignment_;
+    bool adaptive_banding_;
```
atadkase, line 633: [nit] Should we rename all the `[cuda_]banded_alignment` and `[cudap]adaptivie_banding/ed` to one convention? I can push that change in if needed.

r-mafi: I agree, will be good to have them consistent. Please make the changes if you have a chance.

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -419,279 +493,119 @@ void generatePOA(genomeworks::cudapoa::OutputDetails* output_details_d,
     uint16_t* outgoing_edges_coverage_count = graph_details_d->outgoing_edges_coverage_count;
     SizeT* node_id_to_msa_pos               = graph_details_d->node_id_to_msa_pos;
 
-    int32_t nwindows_per_block = CUDAPOA_THREADS_PER_BLOCK / WARP_SIZE;
-    int32_t nblocks            = (total_windows + nwindows_per_block - 1) / nwindows_per_block;
+    int32_t nwindows_per_block         = CUDAPOA_THREADS_PER_BLOCK / WARP_SIZE;
+    int32_t nblocks                    = (cuda_banded_alignment || cuda_adaptive_banding) ? total_windows : (total_windows + nwindows_per_block - 1) / nwindows_per_block;
```
atadkase, line 497: We can drop the `||` if banded flag is always on when adaptive is on.

cudapoa/src/application_parameters.cpp
r-mafi: yes, 256 is good for the current one for 2 reasons: 1- average band-widths in most cases is around 128, 2- score matrix allocation is based on `band_width*max_number_of_nodes_per_window`, in most cases graph length is smaller than `max_number_of_nodes_per_window` and even if average is adaptive band-width is more than 256, it still can run without problem.

r-mafi: although, if we change the heuristic rules in `set_adaptive_band_parameters()` we may need to adjust this default band-width again.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: In the earlier versions, these two flags were independent, but as you mentioned, currently the're dependent. But I think keeping `||` makes it less prone to possible bugs in future changes, since we do not rely on any implicit assumptions that are subject to change.

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -242,24 +272,52 @@ __global__ void generatePOAKernel(uint8_t* consensus_d,
         // Run Needleman-Wunsch alignment between graph and new sequence.
         SizeT alignment_length;
 
-        if (cuda_banded_alignment)
+        if (banded_alignment || cuda_adaptive_banded)
```
atadkase, line 275: Sorry I missed this permutation! Can you please add a commit dropping the cuda_ prefix here?
Thanks!

r-mafi: sure

PR #497, opened 2020-07-14
[cudaaligner] fixes for banded Myers

ahehn-nv:
Fixed two bugs in the banded Myers implementation for band widths > 1024 matrix elements and added testing.

* bug 1: carry of the diagonal band construction was not handled correctly.
* bug 2: backtrace could end up outside the band under certain conditions.


cudaaligner/src/myers_gpu.cu
```diff
@@ -586,27 +621,27 @@ __device__ void myers_compute_scores_diagonal_band_impl(
...
                     pv_local |= carry_down_bit;
                     mv_local &= ~carry_down_bit;
                 }
 
-                const int32_t carry_right = myers_advance_block(warp_mask, carry_right_bit, eq, pv_local, mv_local, carry_down);
-                carry_down                = ((pv_local & carry_down_bit) == WordType(0) ? 0 : 1) - ((mv_local & carry_down_bit) == WordType(0) ? 0 : 1);
-                score(idx, t)             = score(idx, t - 1) + carry_right + carry_down;
+                const int2 carry_right   = myers_advance_block2(warp_mask, carry_right_bit, eq, pv_local, mv_local, carry);
+                const int32_t carry_down = ((pv_local & carry_down_bit) == WordType(0) ? 0 : 1) - ((mv_local & carry_down_bit) == WordType(0) ? 0 : 1);
+                score(idx, t)            = score(idx, t - 1) + carry_right.x + carry_down;
```
tijyojwad, line 638: I'm not intimately familiar with the details here, but flagging this just in case & I see that carry down is paired with carry_rught.x (as opposed to .y). Just making sure that's the right member of the struct?

ahehn-nv: I need two "carries" along the horizontal: `carry_right.x` and `carry_right.y`, which correspond to the horizontal change in rows n-1 and n of the NW matrix, respectively, when going from column `t-1` to `t`.

The first "carry" I need to update `score(idx,t)`:
Since the score matrix is not the full NW matrix, but only the band along the diagonal, `idx` in `score(idx,t)` is relative to the first element of the band in the column of the NW matrix.
With respect to the full NW matrix, score(idx,t) moves down one row for each column. This means to get from `score(idx,t-1)` to `score(idx,t)`, I need to move one right and then one down, which results in score(idx,t) = score(idx, t-1)  + (change from column t-1 -> t in row n-1 in NW matrix) + (change in column t from row n-1 -> n in NW matrix).
Now, `carry_right.x` = (change from column t-1 -> t in row n-1 in NW matrix).

The second carry I need to actually "carry" from one warp iteration to the next:
The carry propagated from one warp iteration to the next, however, is the change from column `t-1` -> `t` in row n (not n-1). which is `carry_right.y`, because the next iteration will continue in row n+1.

Maybe `delta` would be a better variable name, since only one of the `carry_*` variables is an actual carry.

tijyojwad: Got it, thanks for the explanation @ahehn-nv ! Looks good to me, resolving :).

cudaaligner/tests/cudaaligner_test_cases.cpp
```diff
@@ -88,6 +88,10 @@ std::vector<TestCaseData> create_cudaaligner_test_cases()
...
...CGCTTTTTCACGCGAACGATGCGCTTCTTCATCGATGTTAATATCCAGTTTCTCATCCTGACGGCTAACCGCAGTCAGAGAATCGATTTTCAACTACAACTCTGCAATCAGCTGCAGTGCAAATCAAGGAAATTCCAGACTCACTAAAAATGTTCTGTATTGATAAACTCCGGATTCACCTGCTCCAGGGCTGGCTAAAAATTACCAGGATCGTTGTCGGCGATCGCTCGGAAGACACGCCACGTCGCGCCAAAACGCAACCGCGCGCTGGCATTAATCGCCTCTGATGATGAGATAACTTTGCTGATTGTCCAGAGCGGTGGGGTTTCCGGGGCAGCCAGTCAGGTTTACCAAAAGCTGGTCATCAATTTGCCTGTCGTCCGTGATTACATAAAAGCCCGCGAGCATATTTCAGGCTTCGTCAATATCCATCACTTCGCCGACAGTTATTCAGCCGCTATGATCATCGGCTGACCTGTCCACACGACGTCGGTATCGGTCTTTACCAGCAGTTTTCGAGTGGCCCAACATCAGGGCGTCAGTAAATAACATCTTCCAGCCACCGGAACGGTCATCAGCGTGCAGATACCGATACGATTGCCGCGGTGACCTGAACCGCGGCAACAACGAGCCCCACCGGCGTGACTTTACGTTCTGTCGTTGCGTATCGCGCCGACGCCGCGCCTCCGACATAACGCTCAATCTCTTTTAAATACTGGTTGATAAGTCGCCACGATCTGGGAATTTCAGGCTTCAACGTTCCGTTCCCCATCTGATACCGCCTGCTCTATGCACGGTTGGCAACTGGCGTGGGTCTTCCGTTGTTCAGCTCGCTGAAATGACACGCCGAACATATCGGCCTGACCGACGTGGCTTCCACCTTTCGCGTGTTGATCTGCCGCTTTTAACATCATAGCGAGTTCATCAGCGCGCTGCGCGATGTGTTGAGCAGATAAACACCCCGGACGCGATCAGTTTTTCCAGCACGCGACAGTTCTCGCTTTGCAGTGGTGTCCAGTTACCGGCGCAGAGAGATCGTCGGATCGCCAAGATGGCCATACCATTACCGGAGAAACTGATGATGGCCTCAATCAGGACCTTCACCGACCTTTTGATCAACACCAATACCATACCACGATTCGCCGTCATTGACGTGAAATATTAAGGCGGAGTTTATATCTGGTGGCAGGATTTTCAG";
```
mimaric, line 91: Optional: save this in a file like it's done in `cudamapper/data`?

cudaaligner/tests/Test_MyersAlgorithm.cu
```diff
@@ -0,0 +1,283 @@
+/*
```
mimaric, line 1: Not sure if git got confused here, but just in case you did not know in situation like this you can use `git mv cudaaligner/tests/Test_MyersAlgorithm.cpp cudaaligner/tests/Test_MyersAlgorithm.cu`.
Sometimes if the renamed file changed too much git still decides to treat them as deletion + addition instead of rename

cudaaligner/src/myers_gpu.cu
```diff
@@ -383,7 +418,8 @@ __global__ void myers_compute_score_matrix_kernel(
 __device__ int32_t myers_backtrace_banded(int8_t* path, device_matrix_view<WordType> const& pv, device_matrix_view<WordType> const& mv, device_matrix_view<int32_t> const& score, int32_t diagonal_begin, int32_t diagonal_end, int32_t band_width, int32_t target_size, int32_t query_size)
 {
     assert(threadIdx.x == 0);
-    using nw_score_t = int32_t;
+    using nw_score_t                    = int32_t;
+    GW_CONSTEXPR nw_score_t out_of_band = numeric_limits<nw_score_t>::max() - 1;
```
mimaric, line 422: Why `-1`?

cudaaligner/tests/Test_MyersAlgorithm.cu
ahehn-nv: I used `git mv`, but somehow git got confused. Probably because I added much code afterwards in the same commit.

cudaaligner/src/myers_gpu.cu
ahehn-nv: Because I do `x + 1 == ...` with the variable I assign it to and need to avoid the integer overflow.

cudaaligner/tests/cudaaligner_test_cases.cpp
ahehn-nv: I think this is definitely a good idea - especially the last test strings are too long to hard code them.
However, I'd like to postpone this change, since this is mostly a bug fix PR.

cudaaligner/src/myers_gpu.cu
ahehn-nv: see https://github.com/clara-parabricks/GenomeWorks/blob/416d1c7b2721d690f69866b66a4c6b011784b184/cudaaligner/src/myers_gpu.cu#L469

cudaaligner/tests/cudaaligner_test_cases.cpp
ahehn-nv: opened #505

mimaric: sounds good to me

cudaaligner/src/myers_gpu.cu
mimaric: Ok. Maybe a comment about it could help, but it's optional

ahehn-nv: true... I added a comment.

PR #506, opened 2020-07-16
[cudamapper] Fasta index ids

### LINKED ISSUE #503: [cudamapper] Overlaps at the same position in an index are dropped even when query and target files are not the same.
When using two different FASTA files, reads at index i in the query index and index j in the target index are not overlapped when i == j. This is because of a check that was introduced in https://github.com/clara-parabricks/GenomeWorks/blob/0702c2ffd672fa04887f3b4e82bfd18d1d213218/cudamapper/src/overlapper.cpp#L25, and is valid for all-to-all but not sequence-to-reference overlapping.

The filtering condition should be ignored when not running in all-to-all mode.

edawson:
This PR fixes #503 , which was caused by the fact that reads with different IDs can receive the same query_read_id_ / target_read_id_ in overlaps.


cudamapper/src/overlapper_triggered.cu
```diff
@@ -414,7 +420,7 @@ void OverlapperTriggered::get_overlaps(std::vector<Overlap>& fused_overlaps,
 
     device_buffer<Overlap> d_filtered_overlaps(n_fused_overlap, _allocator, _cuda_stream);
 
-    FilterOverlapOp filterOp(min_residues, min_overlap_len, min_bases_per_residue, min_overlap_fraction);
+    FilterOverlapOp filterOp(min_residues, min_overlap_len, min_bases_per_residue, min_overlap_fraction, query_index == target_index);
```
edawson, line 423: @mimaric @vellamike The reason this PR is failing tests is because this condition can be triggered if both pointer-to-indexes are the same int (e.g., 0). Do you have any recommendations for a safer why to check equivalence of two indexes?

edawson: This was fixed in the tests by passing `nullptr` in place of both indexes, but would still appreciate input on how to do this better.

cudamapper/include/claraparabricks/genomeworks/cudamapper/overlapper.hpp
```diff
@@ -72,12 +73,16 @@ class Overlapper
...
     /// \param d_anchors vector of anchors sorted by query_read_id -> target_read_id -> query_position_in_read -> target_position_in_read (meaning sorted by query_read_id, then within a group of anchors with the same value of query_read_id sorted by target_read_id and so on)
+    /// \param query_index A shared pointer to an index of sketch elements for the queries.
+    /// \param target_index A shared pointer to an index of sketch elements for the targets.
     /// \param min_residues smallest number of residues (anchors) for an overlap to be accepted
     /// \param min_overlap_len the smallest overlap distance which is accepted
     /// \param min_bases_per_residue the minimum number of nucleotides per residue (e.g minimizer) in an overlap
     /// \param min_overlap_fraction the minimum ratio between the shortest and longest of the target and query components of an overlap. e.g if Query range is (150,1000) and target range is (1000,2000) then overlap fraction is 0.85
     virtual void get_overlaps(std::vector<Overlap>& fused_overlaps,
                               const device_buffer<Anchor>& d_anchors,
+                              std::shared_ptr<Index> query_index,
```
mimaric, line 84: `ApplicationParameters` has `all_to_all` member (https://github.com/clara-parabricks/GenomeWorks/blob/dev-v0.5.0/cudamapper/src/application_parameters.hpp#L62). I think passing that bool is cleaner that passing both indices, plus that way we have a centralized place where we decide if we are doing map to reference or all to all.

cudamapper/src/overlapper_triggered.cu
mimaric: see my other comment, I think we can use `ApplicationParameters::all_to_all` (i.e. pass that bool to overlapper, no the whole `ApplicaitonParameters`)

cudamapper/include/claraparabricks/genomeworks/cudamapper/overlapper.hpp
edawson: @vellamike and I debated this a bit, since `all_to_all` gets passed around less frequently than indices do. I think you're right though, and this prevents the awkward pointer comparison I do in its place.

vellamike: OK. Thinking about it more and accounting for the failing test and I agree, let's go with passing in `all_to_all`

PR #507, opened 2020-07-16
[cudamapper] Do not apply minimizer frequency filtering by default when overlapping very few reads

edawson:
Change default behavior so that if the number of ingested reads in either the query file or target file is < 5, do not filter by default.

This behavior can be overriden by passing -F.


cudamapper/src/application_parameters.hpp
```diff
@@ -48,6 +48,7 @@ class ApplicationParameters
     int32_t index_size                      = 30;    // i
     int32_t target_index_size               = 30;    // t
     double filtering_parameter              = 1e-5;  // F
+    bool custom_filtering_parameter         = false; // Enables override of the adaptive default for -F (i.e., -F can be set even on small readsets)
```
vellamike, line 51: I don't understand the point of this variable, it's not settable by the user so what does it do?

cudamapper/src/application_parameters.cpp
```diff
@@ -219,6 +220,11 @@ void ApplicationParameters::create_input_parsers(std::shared_ptr<io::FastaParser
         target_parser = io::create_kseq_fasta_parser(target_filepath, kmer_size + windows_size - 1);
     }
 
+    if ((query_parser->get_num_seqences() < 5 || target_parser->get_num_seqences() < 5) && !custom_filtering_parameter)
```
vellamike, line 223: Shouldn't we check based on a number of bases rather than number of sequences?

edawson: Absolutely - can we extract that from the query/target parsers?

cudamapper/src/application_parameters.hpp
edawson: Ah, I should have just placed this one in the `.cpp`.

edawson: I realized that since this is needed by multiple functions in the class, it makes sense to have it be a class member. I've made it private to reflect that users should not modify it directly.

cudamapper/src/application_parameters.cpp
mimaric: Not directly, we would have to loop over sequences and check their lengths. This shouldn't be too costly as we can stop looping once we find enough bases

cudamapper/src/application_parameters.hpp
```diff
@@ -67,6 +67,7 @@ class ApplicationParameters
     int64_t max_cached_memory_bytes;
 
 private:
+    bool custom_filtering_parameter = false; // Enables override of the adaptive default for -F (i.e., -F can be set even on small readsets)
```
mimaric, line 70: Does this have to be a member variable? It could be declared in the constructor and passed to `create_input_parsers()`.

I actually think that changing filtering parameters should be done directly in constructor (or a dedicated function), not in `create_input_parsers()`. This is a consequence of properties of queries, it doesn't really have much to do with parser creation

cudamapper/src/application_parameters.hpp
mimaric: Ok, I now see this discussion as well. I think this value should be declared in constructor and passed around as needed, no need to keep it for the entire lifetime of the object

cudamapper/src/application_parameters.cpp
```diff
@@ -219,6 +222,27 @@ void ApplicationParameters::create_input_parsers(std::shared_ptr<io::FastaParser
         target_parser = io::create_kseq_fasta_parser(target_filepath, kmer_size + windows_size - 1);
     }
 
+    number_of_basepairs_t total_sequence_length                 = 0;
+    const number_of_basepairs_t minimum_for_automatic_filtering = 500000; // Require at least 0.5Mbp of sequence for filtering by default
+    number_of_reads_t query_index                               = 0;
+    number_of_reads_t target_index                              = 0;
```
mimaric, line 228: optional: this can be only one counter and just set back to zero before going for target sequences, they are never used simultaneously

cudamapper/src/application_parameters.cpp
```diff
@@ -219,6 +222,27 @@ void ApplicationParameters::create_input_parsers(std::shared_ptr<io::FastaParser
         target_parser = io::create_kseq_fasta_parser(target_filepath, kmer_size + windows_size - 1);
     }
 
+    number_of_basepairs_t total_sequence_length                 = 0;
+    const number_of_basepairs_t minimum_for_automatic_filtering = 500000; // Require at least 0.5Mbp of sequence for filtering by default
```
mimaric, line 226: Do we want to have this condition for query and target together or separate?
What happens if we have a very long query, but the target only has two reads?
(not sure about this, just asking)

cudamapper/src/application_parameters.cpp
edawson: Is that the preferred style? I had initially done this, but introduced the bug where I didn't reset the counter...

cudamapper/src/application_parameters.cpp
edawson: Agreed, though this isn't a case we target right now (like read-ref mapping).

mimaric: Ok, marking as resolved

cudamapper/src/application_parameters.cpp
mimaric: Makes sense, let's leave it as it is

cudamapper/src/application_parameters.hpp
```diff
@@ -73,6 +73,16 @@ class ApplicationParameters
     void create_input_parsers(std::shared_ptr<io::FastaParser>& query_parser,
                               std::shared_ptr<io::FastaParser>& target_parser);
 
+    ///
```
mimaric, line 76: It this line needed?

PR #508, opened 2020-07-19
[cudamapper] cudaMapper sample application

### LINKED ISSUE #340: [cudamapper] example API usage
Hi!

I really appreciated the example API usage for cudapoa and would like to ask for something similar for cudamapper.

Thanks!
Armin

nvvishanthi:
Add a cudaMappper sample application that shows a simple use of the APIs

closes #340


nvvishanthi:
@mimaric a couple of things:
- I couldn't figure out how to create an overlapper object with the public API. For instance the matcher has `create_matcher()` . Does it make sense to have something similar for the overlapper?
- the use of index_descriptor for grouping reads also isn't part of the public API. Should it be? It seems useful for the full cudamapper pipeline without having to reimplement a  the grouping/batching logic.

Still a few things that need to be resolved for this. What should the "default" fasta file be? I made some assumptions (in terms of read size as well as the all-to-all assumption) are those alright?

mimaric:
> I couldn't figure out how to create an overlapper object with the public API. For instance the matcher has `create_matcher()` . Does it make sense to have something similar for the overlapper?

Yes, this is a bug/missing feature. There should be a `make_overlapper()` function. I created #509 

> the use of index_descriptor for grouping reads also isn't part of the public API. Should it be? It seems useful for the full cudamapper pipeline without having to reimplement a  the grouping/batching logic.

This is an interesting question. Do you think it could be used exactly in the current form? When I was developing it I had our `IndexCache` strategy in mind, and as that should definitely be a private cudamapper thing I didn't consider putting anything else in the public interface.
Would you like to create an issue for it?

> Still a few things that need to be resolved for this. What should the "default" fasta file be? I made some assumptions (in terms of read size as well as the all-to-all assumption) are those alright?

I'd ask @edawson and @vellamike about this

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+                                                                 query_index_descriptor.first_read() + query_index_descriptor.number_of_reads(),
+                                                                 KMER_SIZE,
+                                                                 WINDOWS_SIZE,
+                                                                 true,                 // hash representations
+                                                                 FILTERING_PARAMETER); // filter parameter
+
+        for (const IndexDescriptor& target_index_descriptor : target_index_descriptors)
+        {
+            // skip pairs in which target batch has smaller id than query batch as it will be covered by symmetry
+            if (target_index_descriptor.first_read() >= query_index_descriptor.first_read())
```
mimaric, line 141: So we are only covering the all-to-all case? @vellamike?

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+    {
+        std::cout << "CUDA Mapper API sample program. Runs minimizer-based approximate mapping" << std::endl;
+        std::cout << "Usage:" << std::endl;
+        std::cout << "./sample_cudamapper [-p] [-h]" << std::endl;
+        std::cout << "-p : Print the overlaps to stdout" << std::endl;
+        std::cout << "-h : Print help message" << std::endl;
+        std::exit(0);
+    }
+
+    // Load FASTA/FASTQ file. Assume all-to-all
```
mimaric, line 214: Ok, Apparently we are assuming all-to-all. @vellamike?

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+
+    // All overlaps are saved to a single vector of chars and that vector is then printed to output.
+
+    // Allocate approximately 150 characters for each overlap which will be processed,
+    // if more characters are needed buffer will be reallocated.
+    std::vector<char> buffer(150 * number_of_overlaps_to_print);
+    // characters written buffer so far
+    int64_t chars_in_buffer = 0;
+
+    {
```
mimaric, line 65: This is a highly optimized version of PAF generating function. As this is supposed to be a simple sample without much caring for the performance I would suggesting going for a simpler solution (even as simple as printing each line individually). An alternative would be to expose cudamapper's function in the public interface, as discussed above

mimaric:
Regarding `IndexDesciptor`, I think it would make sense to expose it in public interface. Would you like to do it now or merge as it is and then follow up with a separate PR?

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: My reasoning with using this was so that the output would be in the same format as the main cudamapper binary, and I could sanity check the results easily. I agree that it would be nice to add some sort of formatted print function into the public interface. I think it was mentioned (or included as a bullet point) at the sync yesterday

nvvishanthi:
> Regarding `IndexDesciptor`, I think it would make sense to expose it in public interface. Would you like to do it now or merge as it is and then follow up with a separate PR?

I think merging it soon and then following up with another PR would be good.

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include <claraparabricks/genomeworks/cudamapper/index.hpp>
+#include <claraparabricks/genomeworks/cudamapper/matcher.hpp>
+#include <claraparabricks/genomeworks/cudamapper/overlapper.hpp>
+
+#include <cuda_runtime_api.h>
+#include <iostream>
+#include <string>
+
+using namespace claraparabricks::genomeworks;
+using namespace claraparabricks::genomeworks::cudamapper;
```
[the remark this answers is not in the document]
nvvishanthi: Ah ok, good to know. I saw the other sample_cudapoa and sample_cudaaligner use `using namespace` so I thought that was ok to do for the samples.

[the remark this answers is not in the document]
nvvishanthi: Or instead, is only having  `using claraparabricks::genomeworks` and getting rid of `using claraparabricks::genomeworks::cudamapper` ok?

cudamapper/samples/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
+#
+# Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
nvvishanthi: Fixed

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
+/*
+* Copyright (c) 2019, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
nvvishanthi: Fixed

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -0,0 +1,234 @@
...
+#include <cuda_runtime_api.h>
+#include <iostream>
+#include <string>
+
+using namespace claraparabricks::genomeworks;
+using namespace claraparabricks::genomeworks::cudamapper;
+
+// define constants. See cudamapper/src/application_parameters.hpp
+// constants used in multiple places
+#define INDEX_SIZE 30
```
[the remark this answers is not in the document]
nvvishanthi: Fixed

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: @mimaric I think iterating over the overlaps and doing this: ```
const int32_t added_chars = std::sprintf(buffer.data() + chars_in_buffer,
                                                     "%s\t%lu\t%i\t%i\t%c\t%s\t%lu\t%i\t%i\t%i\t%ld\t%i",
                                                     query_read_name.c_str(),
                                                     query_parser.get_sequence_by_id(overlaps[i].query_read_id_).seq.length(),
                                                     overlaps[i].query_start_position_in_read_,
                                                     overlaps[i].query_end_position_in_read_,
                                                     static_cast<unsigned char>(overlaps[i].relative_strand),
                                                     target_read_name.c_str(),
                                                     target_parser.get_sequence_by_id(overlaps[i].target_read_id_).seq.length(),
                                                     overlaps[i].target_start_position_in_read_,
                                                     overlaps[i].target_end_position_in_read_,
                                                     overlaps[i].num_residues_ * KMER_SIZE, // Print out the number of residue matches multiplied by kmer size to get approximate number of matching bases
                                                     std::max(std::abs(static_cast<int64_t>(overlaps[i].target_start_position_in_read_) - static_cast<int64_t>(overlaps[i].target_end_position_in_read_)),
                                                              std::abs(static_cast<int64_t>(overlaps[i].query_start_position_in_read_) - static_cast<int64_t>(overlaps[i].query_end_position_in_read_))), //Approximate alignment length
                                                     255);
...
printf("%s"..);
```
would be useful in the public API for basic formatted printing

**Comment by nvvishanthi** [comment]
Rebased + squashed commits

File: cudamapper/samples/sample_cudamapper.cpp (hunk 4 above)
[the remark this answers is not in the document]
Comment (mimaric) [reply cudamapper/samples/sample_cudamapper.cpp]: I would go for solution from `main.cu` for consistency, but I wouldn't mind it being different either

File: cudamapper/samples/sample_cudamapper.cpp (hunk 1 above)
Comment (mimaric) [reply cudamapper/samples/sample_cudamapper.cpp:137]: Clarified in private conversation, yes we are.

File: cudamapper/samples/sample_cudamapper.cpp (hunk 2 above)
Comment (mimaric) [reply cudamapper/samples/sample_cudamapper.cpp:211]: Clarified in private conversation, yes we are.

File: cudamapper/samples/sample_cudamapper.cpp (hunk 3 above)
Comment (mimaric) [reply cudamapper/samples/sample_cudamapper.cpp:61]: I think that the best solution would be to expose `print_paf()` in the public interface (creating a header file for it in `cudamapper/include/claraparabricks/genomeworks/cudamapper/`).
Until this is done I would just add `#include "../src/cudamapper_utils.hpp"` like we do for `IndexDescriptor`

Comment (mimaric) [reply cudamapper/samples/sample_cudamapper.cpp:61]: Created issue #516

**Comment by mimaric** [comment]
> > Regarding `IndexDesciptor`, I think it would make sense to expose it in public interface. Would you like to do it now or merge as it is and then follow up with a separate PR?
> 
> I think merging it as is and then following up with another PR would be good.

I agree. Created issue #517

**Comment by mimaric** [comment]
@nvvishanthi FYI after addressing a comment we just write "fixed", "done" or similar and let the comment originator mark it as resolved

**Comment by nvvishanthi** [comment]
> @nvvishanthi FYI after addressing a comment we just write "fixed", "done" or similar and let the comment originator mark it as resolved

Ah ok thanks for letting me know. Sorry about the self-resolve

File: cudamapper/samples/sample_cudamapper.cpp (hunk 3 above)
Comment (nvvishanthi) [reply cudamapper/samples/sample_cudamapper.cpp]: Fixed

File: cudamapper/samples/sample_cudamapper.cpp [hunk 8]
```diff
@@ -0,0 +1,234 @@
...
+#define MIN_RESIDUES 3
+#define MIN_OVERLAP_LEN 250
+#define MIN_BASES_PER_RESIDUE 1000
+#define MIN_OVERLAP_FRACTION 0.8
+
+// constant used in the indices
+#define FILTERING_PARAMETER 1e-5
+
+// prints overlaps to stdout
+void print_overlaps(const std::vector<Overlap>& overlaps,
```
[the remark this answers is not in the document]
nvvishanthi: Fixed

cudamapper/samples/sample_cudamapper.cpp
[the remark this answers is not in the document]
nvvishanthi: Fixed. Changed to be like `main.cu`

PR #510, opened 2020-07-20
[cudamapper] Add create_overlapper function to instantiate overlapper object

### LINKED ISSUE #509: [cudamapper] Create public interface for overlapper
Similarly to other classes in public interface `Overlapper` should also have a `create_overlapper()` function.

For example see https://github.com/clara-parabricks/GenomeWorks/blob/dbf5b61e0bf11124d8de366e6a4292a36f7b6038/cudamapper/include/claraparabricks/genomeworks/cudamapper/matcher.hpp#L51

nvvishanthi:
Expose instantiation of overlapper object through a `create_overlapper()` function in the public API and replace the Overlapper creation in main.cu with public API function call.

closes #509

### DIFF
--- a/cudamapper/include/claraparabricks/genomeworks/cudamapper/overlapper.hpp
+++ b/cudamapper/include/claraparabricks/genomeworks/cudamapper/overlapper.hpp
@@ -110,6 +110,13 @@ public:
                                     const io::FastaParser& target_parser,
                                     std::int32_t extension,
                                     float required_similarity);
+
+    /// \brief Creates a Overlapper object
+    /// \param allocator The device memory allocator to use for buffer allocations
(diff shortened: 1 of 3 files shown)


nvvishanthi:
@mimaric  this seemed easy enough to do, so I just added it real quick. Re-ran the main cudamapper binary on a couple of datasets, and behavior is the same with/without the change. Does this meet the criteria for #509?

mimaric:
Regarding failing `gpuCI/clara-genomics-analysis/check-style` test, you can locally run `make check-format` to check format or just `make format` in order to automatically correct formatting. If you decide to use `make format` directly I suggest staging your changes first in order to be able to see the formatting changes so you can learn the rules better

nvvishanthi:
> Looks good to me, just remove #include "overlapper_triggered.hpp" from line 40 in main.cu 
Ah good catch. Fixed.

`make format` fixed the formatting and `make check-format` is passing on my machine

PR #513, opened 2020-07-23
[cudaaligner] banded myers with fixed bandwidth and low memory requirements

ahehn-nv:
* Banded Myers takes an upper bound for the bandwidth now
* Banded Myers has a dynamic memory structure now - it doesn't allocate buffers of a fixed size for an alignment (based on the max_query_length and max_target_length given at construction) anymore.
* Added a new API to create a cudaaligner with a fixed bandwidth
* batched_device_matrices can accommodate matrices of different sizes now

### DIFF
--- a/cudaaligner/include/claraparabricks/genomeworks/cudaaligner/aligner.hpp
+++ b/cudaaligner/include/claraparabricks/genomeworks/cudaaligner/aligner.hpp
@@ -82,7 +82,7 @@ public:
     virtual void reset() = 0;
 };
 
-/// \brief Created Aligner object
+/// \brief Created Aligner object - DEPRECATED API
 ///
 /// \param max_query_length Maximum length of query string
 /// \param max_target_length Maximum length of target string
@@ -95,7 +95,7 @@ public:
 /// \return Unique pointer to Aligner object
 std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_target_length, int32_t max_alignments, AlignmentType type, DefaultDeviceAllocator allocator, cudaStream_t stream, int32_t device_id);
 
-/// \brief Created Aligner object
+/// \brief Created Aligner object - DEPRECATED API
 ///
 /// \param max_query_length Maximum length of query string
 /// \param max_target_length Maximum length of target string
@@ -108,7 +108,7 @@ std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_ta
 /// \return Unique pointer to Aligner object
 std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_target_length, int32_t max_alignments, AlignmentType type, cudaStream_t stream, int32_t device_id, int64_t max_device_memory_allocator_caching_size = -1);
 
-/// \brief Created Aligner object - NEW API
+/// \brief Created Aligner object
 ///
 /// \param type Type of aligner to construct
 /// \param max_bandwidth Maximum bandwidth for the Ukkonen band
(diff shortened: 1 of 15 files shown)


ahehn-nv:
Functionality to distinguish optimal/approximate alignments and add the number of edit of the alignments to the alignment object will follow in a separate PR once this is merged.

cudaaligner/src/aligner_global_myers_banded.cpp
```diff
@@ -29,43 +33,312 @@ namespace genomeworks
...
-        , scores(max_alignments, max_n_words * (max_target_length + 1), allocator, stream)
-        , query_patterns(max_alignments, max_n_words * 4, allocator, stream)
+
+constexpr int32_t n_alignments_initial_parameter = 1000; // for initial allocation - buffers will grow automatically if needed.
+
+using myers::WordType;
+constexpr int32_t word_size = sizeof(myers::WordType) * CHAR_BIT;
+
+template <typename T>
+using pinned_host_vector = std::vector<T, thrust::system::cuda::experimental::pinned_allocator<T>>;
```
mimaric, line 45: Should this be moved to `common/base/include/claraparabricks/genomeworks/types.hpp`?
Probably not necessary to do it right away

ahehn-nv: Good point. However, I think we should not move it to types.hpp, but a separate header file. It implies including header files for a pinned allocator from thrust and std::vector, which I would prefer not to force on every compilation unit. How about `claraparabricks/genomworks/utils/pinned_host_vector.hpp`?

cudaaligner/include/claraparabricks/genomeworks/cudaaligner/aligner.hpp
```diff
@@ -108,6 +108,28 @@ std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_ta
...
+/// \brief Created Aligner object - NEW API
+///
+/// \param type Type of aligner to construct
+/// \param max_bandwidth Maximum bandwidth for the Ukkonen band
+/// \param stream CUDA Stream used for GPU interaction of the object
+/// \param device_id GPU device ID to run all CUDA operations on
+/// \param max_device_memory Maximum amount of device memory used in bytes (-1 (default) for all available memory).
+///
+/// \return Unique pointer to Aligner object
+std::unique_ptr<Aligner> create_aligner(AlignmentType type, int32_t max_bandwidth, cudaStream_t stream, int32_t device_id, int64_t max_device_memory = -1);
```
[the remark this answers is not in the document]
ahehn-nv: Does this need to be mentioned in the documentation? In my opinion, it is an implementation detail the library user does not need to know about, what do you think?

cudaaligner/include/claraparabricks/genomeworks/cudaaligner/aligner.hpp
```diff
@@ -108,6 +108,28 @@ std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_ta
 /// \return Unique pointer to Aligner object
 std::unique_ptr<Aligner> create_aligner(int32_t max_query_length, int32_t max_target_length, int32_t max_alignments, AlignmentType type, cudaStream_t stream, int32_t device_id, int64_t max_device_memory_allocator_caching_size = -1);
 
+/// \brief Created Aligner object - NEW API
```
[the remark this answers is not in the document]
ahehn-nv: Fair point. I will change it.

cudaaligner/include/claraparabricks/genomeworks/cudaaligner/aligner.hpp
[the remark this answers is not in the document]
mimaric: It's just that the function above takes an allocator explicitly, while this one does not mention it, but yes, I agree this is an implementation detail

cudaaligner/src/aligner_global_myers_banded.cpp
mimaric: Yes, that sounds good. If it's too much work you can leave it as it is for now as well.

mimaric: Ok, now I see you already changed it

PR #519, opened 2020-07-29
[cudapoa] modified adaptive banded alignment

r-mafi:
current implementation is branched off abPOA in [PR #495 ](https://github.com/clara-parabricks/GenomeWorks/pull/495), followed by some changes that make it more similar to current static banded implementation. Forgetting about this evolutionary back and forth, the main differences of modified adaptive alignment with the present banded alignment is:
1- it extends band-width for cases where score matrix is not square (e.g. aligning a long read to a short POA graph, or the other way around)
2- in backtracking section, it checks if the trace is too close to band borders, if so, it reruns the alignment with a larger band-width and shifted from the main diagonal


cudapoa/src/cudapoa_batch.cuh
```diff
@@ -462,6 +464,10 @@ protected:
             GW_LOG_WARN("Kernel Error:: Loop count exceeded upper bound in nw algorithm in batch {}\n", bid_);
             output_status.emplace_back(error_type);
             break;
+        case genomeworks::cudapoa::StatusType::exceeded_adaptive_banded_matrix_size:
```
atadkase, line 467: I inserted this error message here as code 6 because it made sense for grouping, but it might affect error code tracking. What is the error code strategy? Should it be moved to the end of the list when adding new messages?

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
```diff
@@ -0,0 +1,572 @@
...
+            loop_count++;
+            ScoreT scores_ij = get_score_adaptive(scores, i, j, band_width, band_shift, gradient, max_column, min_score_value);
+            bool pred_found  = false;
+            // Check if move is diagonal.
+            if (i != 0 && j != 0)
+            {
+
+                if (rerun == 0)
+                {
+                    // check if backtrace gets too close or hits the band limits, if so stop and rerun with extended band-width
```
atadkase, line 460: Should this check be performed for vertical/horizontal moves too?

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
```diff
@@ -0,0 +1,572 @@
...
+                    SizeT threshold = max(1, max_column / (4 * static_band_width)); // ad-hoc rule 7
```
tijyojwad, line 461: what's intuition around 4 * band width here?

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
```diff
@@ -0,0 +1,572 @@
...
+                    if (j > threshold && j < max_column - threshold)
+                    {
+                        band_start = get_band_start_for_row_adaptive(i, gradient, band_width, band_shift, max_column);
+                        if (j <= band_start + threshold) // ad-hoc rule 8-a
+                        {
+                            aligned_nodes = -3;
```
tijyojwad, line 467: can you add a note somewhere describing the heuristics and each "backtrace failure" mode? e.g. `-3` is when backtrace is close to left edge, and -4` when close to right edge, and both have different heuristics for rerun

cudapoa/src/cudapoa_nw_banded.cuh
```diff
@@ -47,31 +47,18 @@ __device__ SizeT get_band_start_for_row(SizeT row_idx, float gradient, SizeT ban
...
+    start_pos = max(start_pos, 0);
 
     start_pos = start_pos - (start_pos % CELLS_PER_THREAD);
 
     return start_pos;
 }
 
 template <typename ScoreT, typename SizeT>
-__device__ ScoreT* get_score_ptr(ScoreT* scores, SizeT row, SizeT column, float gradient, SizeT band_width, SizeT max_column)
+__device__ ScoreT* get_score_ptr(ScoreT* scores, SizeT row, SizeT column, SizeT band_start, SizeT band_width)
```
tijyojwad, line 58: is this shared with the adaptive banding file? if so maybe we can move them into a shared `.cuh` file like `banded_nw_utils`?

pygenomeworks/genomeworks/cudapoa/cudapoa.pyx
```diff
@@ -162,6 +165,7 @@ cdef class CudaPoaBatch:
             mismatch_score=-6,
             match_score=8,
             cuda_banded_alignment=False,
+            cuda_corrective_alignment=False,
```
tijyojwad, line 168: this can also be updated to use an enum to mirror the C++ side of things

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
r-mafi: in the backtrace loop, at each iteration it considers whether the trace is moving along vertical, horizontal or diagonal path. Then after it is determined which one it is, we check if the found `(i, j)` happens to be too close to the assumed band bounds at row `i`, if so, we exit with error code -3 or -4.

cudapoa/src/application_parameters.cpp
```diff
@@ -182,6 +189,9 @@ void ApplicationParameters::help(int32_t exit_code)
         -b, --band-width <int>
             band-width for banded alignment (must be multiple of 128) [256])"
               << R"(
+        -c, --corrective-alignment
```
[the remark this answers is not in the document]
r-mafi: done.

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -231,6 +239,7 @@ __global__ void generatePOAKernel(uint8_t* consensus_d,
                 consensus[1] = static_cast<uint8_t>(StatusType::node_count_exceeded_maximum_graph_size);
                 warp_error   = true;
             }
+            /// ToDo clean all instances of node_distance
```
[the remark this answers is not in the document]
r-mafi: it's already done in this PR, removed ToDo item.

cudapoa/src/cudapoa_batch.cuh
r-mafi: I thought somewhere @tijyojwad had mentioned new error codes to be added to the bottom of the list. However, I think logical order makes sense, also we have previously removed some redundant error codes, so this won't be the first time the error list is changed.

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
atadkase: Sounds good.

cudapoa/src/cudapoa_generate_consensus.cuh
```diff
@@ -282,7 +282,7 @@ __device__ void generateConsensus(uint8_t* nodes,
     consensus[consensus_pos] = '\0';
 }
 
-template <bool cuda_banded_alignment = false, typename SizeT>
```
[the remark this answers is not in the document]
atadkase: I think we need to run this through PEF to verify this. I removed it because it was bloating the calls file with multiple permutations with all the different configurations.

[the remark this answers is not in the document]
tijyojwad: I think the cudapoa microbenchmarks would be sufficient for that. we don't need to run the end to end pipeline for it.

cudapoa/include/claraparabricks/genomeworks/cudapoa/batch.hpp
```diff
@@ -212,15 +212,16 @@ class Batch
...
-/// \param cuda_banded_alignment Whether to use banded alignment
+/// \param device_id                GPU device on which to run CUDA POA algorithm
+/// \param stream                   CUDA stream to use on GPU
+/// \param max_mem                  Maximum GPU memory to use for this batch.
+/// \param output_mask              which outputs to produce from POA (msa, consensus)
+/// \param batch_size               defines upper limits for size of a POA batch, i.e. sequence length and other related parameters
+/// \param gap_score                score to be assigned to a gap
+/// \param mismatch_score           score to be assigned to a mismatch
+/// \param match_score              score to be assigned for a match
+/// \param banded_alignment         whether to use banded alignment
```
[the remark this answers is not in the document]
r-mafi: added `enum BandMode` in `cudapoa.hpp`

cudapoa/include/claraparabricks/genomeworks/cudapoa/cudapoa.hpp
```diff
@@ -37,6 +37,7 @@ enum StatusType
     exceeded_maximum_sequences_per_poa,
     node_count_exceeded_maximum_graph_size,
     edge_count_exceeded_maximum_graph_size,
+    exceeded_adaptive_banded_matrix_size,
```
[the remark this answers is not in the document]
r-mafi: previously we have removed a redundant error code, so the order historically has been changed. How about printing an error code description, I guess currently we do it only in debug mode. I don't have problem with moving it to the end, but it will look a bit out of place.

cudapoa/include/claraparabricks/genomeworks/cudapoa/utils.hpp
```diff
@@ -55,6 +56,7 @@ void get_multi_batch_sizes(std::vector<BatchSize>& list_of_batch_sizes,
                            std::vector<std::vector<int32_t>>& list_of_groups_per_batch,
                            const std::vector<Group>& poa_groups,
                            bool banded_alignment               = true,
+                           bool adaptive_banded                = false,
```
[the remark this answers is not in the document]
r-mafi: done.

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
r-mafi: done

cudapoa/src/cudapoa_nw_banded.cuh
r-mafi: there are some slight differences between the two, but I agree there are multiple kernels similar in banded-NW and adaptive-NW which with some effort and be unified and combined together in a separate file. I'll create an issue to address this in a different PR.

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
r-mafi: good question, I found it's better to define the proximity tolerance relative to sequence length as opposed to an absolute value. Linear relation, max_column / 1024 seemed to work well. There is no inverse relation with band-width, replaced `4 * static_band_width` (which happened to be  = 1024) with `max_column / 1024`.

pygenomeworks/genomeworks/cudapoa/cudapoa.pyx
r-mafi: @atadkase will take a look at changes related to python binding.

cudapoa/include/claraparabricks/genomeworks/cudapoa/batch.hpp
```diff
@@ -56,42 +56,56 @@ struct Entry
...
+            max_nodes_per_graph    = cudautils::align<int32_t, 4>(4 * max_sequence_size);
+            matrix_graph_dimension = cudautils::align<int32_t, 4>(max_nodes_per_graph);
+            // 8 = CUDAPOA_BANDED_MATRIX_RIGHT_PADDING
+            matrix_sequence_dimension = cudautils::align<int32_t, 4>(alignment_band_width + 8);
+        }
+        else // BandMode::adaptive_band
+        {
+            max_nodes_per_graph    = cudautils::align<int32_t, 4>(4 * max_sequence_size);
+            matrix_graph_dimension = cudautils::align<int32_t, 4>(max_nodes_per_graph);
+            // 8 = CUDAPOA_BANDED_MATRIX_RIGHT_PADDING, *2 is to reserve extra memory for cases with extended band-width
```
tijyojwad, line 105: Having this 8 is not ideal here. I think the reason is that the macro is in one of the `.cu` files and is not accessible from here.

Now it looks like the `BatchConfig` isn't templated, so we should be able to move the implementation into a `batch.cpp` file. and perhaps there we can share a private header file that has the static macros for banding?

cudapoa/include/claraparabricks/genomeworks/cudapoa/cudapoa.hpp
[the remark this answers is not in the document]
tijyojwad: Fair enough, okay to leave as is then. And I agree, an error description function is longtime pending...

cudapoa/src/application_parameters.cpp
```diff
@@ -63,10 +62,10 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
         case 'a':
             msa = true;
             break;
-        case 'f':
-            banded = false;
-            break;
         case 'b':
+            band_mode = std::stoi(optarg);
```
tijyojwad, line 66: can this `band_mode` variable be made a `BandMode` type? then in this conditional we set the `BandMode` to the relevant type based on `stoi`, and all checks downstream can be based on `band_mode == BandMode::adaptive?` etc?

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -119,40 +126,43 @@ __global__ void generatePOAKernel(uint8_t* consensus_d,
...
+    SizeT* node_alignments                = &node_alignments_d[window_idx * max_nodes_per_graph * CUDAPOA_MAX_NODE_ALIGNMENTS];
+    uint16_t* node_alignment_count        = &node_alignment_count_d[window_idx * max_nodes_per_graph];
+    uint16_t* sorted_poa_local_edge_count = &sorted_poa_local_edge_count_d[window_idx * max_nodes_per_graph];
 
     int32_t scores_width = window_details_d[window_idx].scores_width;
 
     int64_t scores_offset;
-    if (cuda_banded_alignment)
+    int64_t banded_score_matrix_size;
+    if (banded_alignment || adaptive_banded)
```
tijyojwad, line 146: Optional - can `banded_alignment` variable be renamed to `static_banded` to match naming style of `adaptive_banded`?

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
tijyojwad: Cool, thanks!

tijyojwad: can you add this explanation to a comment?

cudapoa/src/cudapoa_nw_banded.cuh
tijyojwad: sounds good, thanks!

pygenomeworks/genomeworks/cudapoa/cudapoa.pyx
tijyojwad: awesome, thanks!

cudapoa/src/cudapoa_nw_adaptive_banded.cuh
r-mafi: done.

cudapoa/include/claraparabricks/genomeworks/cudapoa/batch.hpp
r-mafi: done, moved to existing `batch.cu` along some other similar changes replacing numbers with static macros.

cudapoa/src/application_parameters.cpp
r-mafi: done.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: sure, done.

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -69,7 +70,7 @@ namespace cudapoa
  * @param[in] mismatch_score                Score for finding a mismatch in alignment
  * @param[in] match_score                   Score for finding a match in alignment
  */
-template <int32_t TPB = 64, bool cuda_banded_alignment = false, bool msa = false, typename ScoreT, typename SizeT>
```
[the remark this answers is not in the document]
tijyojwad: to be investigated by @r-mafi later

PR #520, opened 2020-07-30
[cudamapper] Move print_paf() to public interface

### LINKED ISSUE #516: [cudamapper] Expose print_paf() in public interface
As discussed in #508 `print_paf()` should be moved from `cudamapper/src/cudamapper_utils.hpp` into a new file in `cudamapper/include/claraparabricks/genomeworks/cudamapper`

nvvishanthi:
Moves the `print_paf()` function to the public interface so standard formatted prints don't have to be implemented by user

closes #516


cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
```diff
@@ -0,0 +1,56 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#pragma once
+
+#include <mutex>
+#include <vector>
```
mimaric, line 20: Should we also include `#include <string>`?

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -24,6 +24,8 @@
 #include <claraparabricks/genomeworks/utils/cudautils.hpp>
 #include <claraparabricks/genomeworks/utils/signed_integer_utils.hpp>
 
+#include <claraparabricks/genomeworks/cudamapper/utils.hpp>
```
mimaric, line 27: I think it would be better to do this include this where needed (`main.cu`)

cudamapper/src/cudamapper_utils.hpp
```diff
@@ -28,28 +28,9 @@ namespace claraparabricks
 namespace genomeworks
 {
 
-namespace io
```
mimaric, line 31: I can't comment on lines which were not edited. I would also remove `#include <mutex>` above

mimaric: `#include <claraparabricks/genomeworks/utils/allocator.hpp>` also looks suspicious (probably not part of this PR, but we can fix it while we're here)

nvvishanthi: Fixed. Removed mutex and allocator

cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
nvvishanthi: Fixed

cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
```diff
@@ -0,0 +1,56 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#pragma once
+
+#include <mutex>
```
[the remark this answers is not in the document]
nvvishanthi: Is this in reference to the just the <mutex> header? IMO it makes sense to have it here, since `std::mutex` is in the function signature. Am I missing something as to why having the header here isn't necessary?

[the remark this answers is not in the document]
mimaric: Sorry, you're right, I missed the fact that `mutex` is a parameter

cudamapper/src/cudamapper_utils.cpp
nvvishanthi: Fixed, I think. Included only in source file (utils.cpp), main.cu, and sample_cudamapper.cpp

mimaric: Ok. As mentioned in the general comment, I would create a separate `.cpp` file for this header

nvvishanthi: I might be misunderstanding something, but I created a new `.cpp` file in `cudamapper/src/utils.cpp` that corresponds to the `utils.hpp` header. Is this what you meant?

mimaric: yes, looks good now!

PR #536, opened 2020-08-06
[cudamapper] Create public interface for IndexDescriptor class

### LINKED ISSUE #517: [cudamapper] Expose IndexDescriptor in public interface
`IndexDescriptor` is currently not part of public interface. During implementation of cudamapper sample (#508. #340) it turned out that `IndexDescriptor` would be useful in other implementations of cudamapper as well.

Expose `IndexDescriptor` in public interface, probably in the same way e.g. `Matcher` is exposed.

nvvishanthi:
- Creates a new class for the interface called IndexDescriptorBase.
- Refactor sample_cudamapper to use the public interface classes/functions

closes #517


nvvishanthi:
@mimaric I'm still working through refactoring the caching code to use the public interface as well as removing an instance of code duplication for `group_reads_into_indices()`, but I wanted to make sure that whatever I moved to the public interface is ok.

nvvishanthi:
>All it does is provide first_read and number_of_reads values. It also generates a hash from the two, but that doesn't seem to be used anywhere, or is it?
@ahehn-nv the hashing stuff is specifically in caching code (see `cudamapper/src/index_cache.cu`), so how we cache indices would need some modification.

I would support making this a simple struct (I think we could also stuff the hash-related parts into the struct, right?).

>Alternatively, we could get rid of first_read and number_of_reads in the public interface and make create_index() take an IndexDescriptor as parameter instead
If we can reconcile this with the index caching that requires the hash, then this is also fine. I'll leave it up to @tijyojwad and @mimaric, as to what the best solution would be

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
                                                            const cudaStream_t cuda_stream = 0);
 };
 
+/// IndexDescriptorBase - An "almost an Index" class from which an index can be created.
+/// Every Index is defined by its first read and the number of reads.
+class IndexDescriptorBase
+{
+public:
+    /// \brief destructor
+    ~IndexDescriptorBase() = default;
```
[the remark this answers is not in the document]
nvvishanthi: Ah got it, thanks :). Fixed

cudamapper/src/index_descriptor.hpp
```diff
@@ -48,13 +49,13 @@ class IndexDescriptor
     ~IndexDescriptor() = default;
```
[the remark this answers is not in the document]
nvvishanthi: Fixed

nvvishanthi:
rerun tests

ahehn-nv:
> @ahehn-nv the hashing stuff is specifically in caching code (see `cudamapper/src/index_cache.cu`), so how we cache indices would need some modification.

Ok, if the hash is not removed, then it should be a class imo, since the data in the hash is tightly coupled to the other two elements. So let's leave it as it is.

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
...
+
+    /// \brief returns hash value
+    virtual std::size_t get_hash() const = 0;
+
+    /// \brief generates a mapping of (k,w)-kmer-representation to all of its occurrences for one or more sequences
+    /// \param first_read The first read for the index descriptor
+    /// \param number_of_reads The number of reads until the end of the current index
+    /// \return instance of an IndexDescriptor
+    static std::unique_ptr<IndexDescriptorBase>
+    create_index_descriptor(read_id_t first_read, number_of_reads_t number_of_reads);
```
ahehn-nv, line 212: This function is only declared, but not defined.
Do we want users to create custom index descriptors?
So far they seem to taken from `group_reads_into_descriptors` only.

cudamapper/src/index.cu
```diff
@@ -64,6 +65,49 @@ std::unique_ptr<IndexHostCopyBase> IndexHostCopyBase::create_cache(const Index&
                                            cuda_stream);
 }
 
+std::unique_ptr<IndexDescriptorBase> create_index_descriptor(read_id_t first_read,
```
ahehn-nv, line 68: Why not just use the `std::make_unique<IndexDescriptor>(first_read, number_of_reads);`, directly in the `group_reads_into_descriptors()` function?

mimaric:
@nvvishanthi `IndexDescriptorHash` has to be a separate struct/class unless we want to have `IndexDescriptor::operator()` return hash value, which sounds confusing.

@ahehn-nv I like the idea of `Index` accepting `IndexDescriptor` instead of `first_read` and `number_of_reads explicitly`, but it should probably be done in a separate PR. I created #539 to track it.

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
                                                            const cudaStream_t cuda_stream = 0);
 };
 
+/// IndexDescriptorBase - An "almost an Index" class from which an index can be created.
```
mimaric, line 190: I would remove this sentence, `IndexDescriptor` only defines which reads go into `Index`, but a `FastaParser` is needed as well in order to create an `Index`

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
...
+    /// \brief getter
+    virtual read_id_t first_read() const = 0;
+
+    /// \brief getter
+    virtual number_of_reads_t number_of_reads() const = 0;
+
+    /// \brief returns hash value
+    virtual std::size_t get_hash() const = 0;
+
+    /// \brief generates a mapping of (k,w)-kmer-representation to all of its occurrences for one or more sequences
```
mimaric, line 207: copy-paste?

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
...
+    /// \param first_read The first read for the index descriptor
+    /// \param number_of_reads The number of reads until the end of the current index
+    /// \return instance of an IndexDescriptor
```
mimaric, line 210: You mean `IndexDescriptorBase`?

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
...
+    std::size_t operator()(const IndexDescriptorBase& index_descriptor) const;
+};
+
+/// \brief returns a vector of IndexDescriptors in which the sum of basepairs of all reads in one IndexDescriptor is at most max_basepairs_per_index
+/// If a single read exceeds max_chunk_size it will be placed in its own IndexDescriptor.
+///
+/// \param parser parser to get the reads from
+/// \param max_basepairs_per_index the maximum number of basepairs in an IndexDescriptor
+/// \return vector of pointers to IndexDescriptors
+std::vector<std::unique_ptr<IndexDescriptorBase>> group_reads_into_descriptors(const io::FastaParser& parser,
```
mimaric, line 228: Should this be placed in `cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp` or somewhere similar?

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
```diff
@@ -187,6 +187,46 @@ class IndexHostCopyBase
...
+    /// \brief generates a mapping of (k,w)-kmer-representation to all of its occurrences for one or more sequences
+    /// \param first_read The first read for the index descriptor
+    /// \param number_of_reads The number of reads until the end of the current index
+    /// \return instance of an IndexDescriptor
+    static std::unique_ptr<IndexDescriptorBase>
+    create_index_descriptor(read_id_t first_read, number_of_reads_t number_of_reads);
+};
+
+/// IndexDescriptorHash - operator() calculates hash of a given IndexDescriptor
+struct IndexDescriptorHash
```
mimaric, line 216: Do we want this to be exposed in public interface as well? If yes should this then be `IndexDescriptorHashBase`? I think we are hitting the limit of our design where public interface in header-only + `std::unique_ptr<FooBase> create_foo(....)`

mimaric (changes requested):
I think we are generally hitting the limit of our public interface design. I think that the idea Lars had when he was setting up the project was to have the public interface only expose large building blocks.

We should reconsider the whole header-only + `std::unique_ptr<FooBase> create_foo(....)` approach as we keep trying to find ways to avoid it, we usually indicated inadequate architecture

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: I figured that since this was index-related that it would be more appropriate to put into the index header. Though I don't have a strong reason why it shouldn't be in utils.hpp. The only thing that I have in there are the print_paf/print_bam functions

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: I could move this into the internal headers. I suppose a user might want to define the hashing scheme differently, if they needed one

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: removed

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: removed in latest version

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: removed

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: removed in favor of exposing the IndexDescriptor class directly, as discussed previously

cudamapper/include/claraparabricks/genomeworks/cudamapper/index.hpp
nvvishanthi: Moved to utils.hpp. I looked it over again, and it looks cudapoa exposes batching functions as a "utils" function. So like you said, that looks like the better thing to do

cudamapper/src/index.cu
nvvishanthi: removed in favor of using the class directly.

ahehn-nv (approved):
Looks good to me now.

mimaric (approved):

PR #538, opened 2020-08-18
[hotfix][sdk] update cmake structure

tijyojwad:
Update cmake structure for better integration with
other tools that use GW as cmake submodule.

Keep only one project for GW, and remove "project" settings
from all modules.

### DIFF
--- a/cmake/Format.cmake
+++ b/cmake/Format.cmake
@@ -56,7 +56,7 @@ function(gw_enable_formatting_targets)
         set(format_check_list "")
 
         foreach(source IN LISTS FOLDER_FORMAT_SOURCES)
-            file(RELATIVE_PATH RELATIVE_PATH ${CMAKE_SOURCE_DIR} ${source})
+            file(RELATIVE_PATH RELATIVE_PATH ${PROJECT_SOURCE_DIR} ${source})
 
             # Apply formatting
             get_filename_component(SYMBOL_BASE_DIR ${RELATIVE_PATH} DIRECTORY)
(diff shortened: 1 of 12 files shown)


ahehn-nv (approved):
Looks good. Just one question: why is it a hotfix?

cmake/Format.cmake
```diff
@@ -56,7 +56,7 @@ function(gw_enable_formatting_targets)
         set(format_check_list "")
 
         foreach(source IN LISTS FOLDER_FORMAT_SOURCES)
-            file(RELATIVE_PATH RELATIVE_PATH ${CMAKE_SOURCE_DIR} ${source})
+            file(RELATIVE_PATH RELATIVE_PATH ${PROJECT_SOURCE_DIR} ${source})
```
ohadmo, line 59: Since we have only one project now, maybe we should also change `CMAKE_CURRENT_SOURCE_DIR`  to `PROJECT_SOURCE_DIR`  in  https://github.com/clara-parabricks/GenomeWorks/blob/74e6424c156a7ee15b4137d4788a4257ee6482c4/CMakeLists.txt#L91

ohadmo (approved):

ohadmo:
> Looks good. Just one question: why is it a hotfix?

@tijyojwad to put the auto release job for the test? :smile:  
The release commit will also contain the updated `./VERSION` file

ahehn-nv:
:) that's a good reason.

cmake/Format.cmake
tijyojwad: `CMAKE_CURRENT_SOURCE_DIR` tracks the current directory in which the CMakeLists.txt file is being processed. `PROJECT_` doesn't have an equivalent to it since it's just tracking the current dir. `PROJECT_SOURCE_DIR` is the replacement for `CMAKE_SOURCE_DIR` though, since it'll track the root folder where `project` was called instead of where the top level `cmake` command was called from.

PR #540, opened 2020-08-21
API [cudaungappedextender] Add skeleton for API and API usage samples

atadkase:
WIP API Sample


cudaungappedextender/include/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+# Enable formatting
+gw_enable_auto_formatting("${CMAKE_CURRENT_SOURCE_DIR}")
```
ahehn-nv, line 2: There should be no need to have a CMakeLists.txt in the include folder. It should be enough to have this in the `/cudaungappedextender/CMakeLists.txt` or do you experience problems if you remove this file?

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+#pragma once
+
+namespace claraparabricks
+{
+namespace genomeworks
+{
+namespace cudaungappedextender
+{
+
+struct Anchor
```
ahehn-nv, line 25: I would name this `Seed` instead of `Anchor` to avoid confusion with the `Anchor` in cudamapper. It is almost the same thing, but in the end cudamapper's Anchor is different, because it also contains the read_id information. `Seed` also seems to be the vocabulary used by Sneha and Yatish.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+{
+    /// position of first sketch element in query_read_id_
+    position_in_read_t query_position_in_read_;
+    /// position of second sketch element in target_read_id_
+    position_in_read_t target_position_in_read_;
```
ahehn-nv, line 30: Only private data members should have a trailing `_`, for public members of a struct there is no need.

cudaungappedextender/samples/CMakeLists.txt
```diff
@@ -0,0 +1,3 @@
+# Enable formatting
+gw_enable_auto_formatting("${CMAKE_CURRENT_SOURCE_DIR}")
```
ahehn-nv, line 2: Again, the `gw_enable_auto_formatting("${CMAKE_CURRENT_SOURCE_DIR}")` in the `cudaungappedextender/` directory should suffice.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+                               const char* d_target, int32_t target_length,
+                               int32_t input_hspthresh, Anchor* d_seed_hits,
+                               int32_t num_hits, ScoredSegment* d_hsp_out,
+                               int32_t* d_num_hsps);
+
+    /// \brief Waits for CUDA accelerated extension to finish
+    ///
+    /// Blocking call that waits for all the extensions scheduled
+    /// on the GPU to come to completion.
+    StatusType sync_extensions();
```
ahehn-nv, line 80: I would just call this `sync();`. There is no ambiguity what to sync - so no need for the `_extensions` in my opinion. (I should also change this in cudaaligner.)

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+    /// \brief Return the computed segments
+    ///
+    /// \return Vector of Scored Segments
+    const std::vector<std::shared_ptr<ScoredSegment>>& get_scored_segments();
```
ahehn-nv, line 85: Since `ScoredSegment` is a small struct and is part of the public interface. There should be no need to make this a `shared_ptr`. Just `vector<ScoredSegement>` should be fine.

ahehn-nv: Also this method should be `const`.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /// score of the segment
+    int32_t score;
+};
+
+/// CUDA Ungapped Extension object
+class UngappedExtender
+{
+public:
+    /// \brief Destructor for UngappedExtender.
+    ~UngappedExtender() = default;
```
ahehn-nv, line 48: Since this is a concrete class and no other special functions are defined, there is no need to explicitly define the destructor as default. Let the compiler do it for you implicitly.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+    /// position of second sketch element in target_read_id_
+    position_in_read_t target_position_in_read_;
+};
+
+struct ScoredSegment
+{
+    /// Anchor for the segment
+    Anchor anchor;
+    /// length of the segment
+    int32_t len;
```
ahehn-nv, line 38: I would call this `length`.

ahehn-nv (changes requested):
Some comments - open for discussion. @atadkase @tijyojwad

cudaungappedextender/include/CMakeLists.txt
atadkase: Good point. I'll have to check. I was misled by the "CMAKE_CURRENT_SOURCE_DIR".

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: Can we have this as one common data structure re-used by both these elements? Since cudamapper also needs it we can make this a member of the Anchor struct. And yes, I just copied cudamapper's Anchor here for illustration.

ahehn-nv: Since they are so tiny, I would just leave them separate for now. If we happen to see any benefits later on, we may still combine them.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,92 @@
...
+    UngappedExtender(int* h_sub_mat, int input_xdrop, bool input_noentropy,
+                     int gpu_id = 0, cudaStream_t stream = 0);
+
+    /// \brief Host pointer prototype for ungapped extension
+    ///
+    /// Takes values from host data structures,
+    /// copies them over to device,
+    /// launches async extension kernels on specified stream. Filters
+    /// segments on device based on input_hspthresh.
+    StatusType ungapped_extend(const char* h_query, int32_t query_length,
```
atadkase, line 59: Should I rename this to just `extend_async` or `extend`?

ahehn-nv: Good point, I like `extend_async`.
Although it would create some inconsistency - we would need to adapt multiple modules to adhere to the `_async` convention.
I'd leave it as without `_async` for now, and discuss this with everybody on one of the sync calls.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+    /// \brief Host pointer prototype for ungapped extension
+    ///
+    /// Takes values from host data structures,
+    /// copies them over to device,
+    /// launches async extension kernels on specified stream. Filters
+    /// segments on device based on input_hspthresh.
+    StatusType extend_async(const char* h_query, int32_t query_length,
+                            const char* h_target, int32_t target_length,
+                            int32_t input_hspthresh,
```
atadkase, line 59: Thinking of renaming this to score_threshold

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+    /// Seed for the segment
+    Seed seed;
+    /// length of the segment
+    int32_t length;
+    /// score of the segment
+    int32_t score;
+};
+
+/// CUDA Ungapped Extension object
+class UngappedExtender
```
atadkase, line 44: Will be moved to a virtual class for implementation.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+struct Seed
+{
+    /// position of first sketch element in query_read_id_
+    position_in_read_t query_position_in_read;
+    /// position of second sketch element in target_read_id_
+    position_in_read_t target_position_in_read;
+};
+
+struct ScoredSegment
```
gsneha26, line 33: maybe name it SegmentPair

cudaungappedextender/samples/device_sample.cu
```diff
@@ -0,0 +1,142 @@
...
+#include <claraparabricks/genomeworks/utils/cudautils.hpp>
+
+using namespace claraparabricks::genomeworks;
+using namespace claraparabricks::genomeworks::cudaungappedextender;
+
+int main(int argc, char* argv[])
+{
+    const int32_t input_xdrop      = 10;
+    const int32_t input_no_entropy = 0;
+    const int32_t hsp_threshold    = 20000;
```
gsneha26, line 32: The values are 
`const int32_t input_xdrop      = 910;`
`const bool input_no_entropy = false;`
`const int32_t hsp_threshold    = 3000;`

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+#pragma once
+
+namespace claraparabricks
+{
+namespace genomeworks
+{
+namespace cudaungappedextender
+{
+
+struct Seed
```
gsneha26, line 25: `Seed` is the wrong name in the context of WGA. We can either name it `SeedHit` or revert back to the Genomeworks usage of `Anchor`.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+    ///
+    /// Memcopies to device memory are assumed to be done before this
+    /// function. Output array d_hsp_out must be pre-allocated on device.
+    /// Launches async extension kernel. Filters segments on device
+    /// based on input_hspthresh.
+    StatusType extend_async(const char* d_query, int32_t query_length,
+                            const char* d_target, int32_t target_length,
+                            int32_t input_hspthresh, Seed* d_seeds,
+                            int32_t num_seeds, ScoredSegment* d_hsp_out,
+                            int32_t* d_num_hsps);
```
gsneha26, line 72: `int32_t* d_num_hsps` is not required in `extend_async()`. Also, please remove it from `device_sample.cu` and `host_sample.cu`.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: `Anchor` in Genomeworks currently has references to the query and target sequences in that structure. `SeedHit` implies that there's a missing step in this offering which should generate the `Hits` from `Seeds`, which is indeed a step in segalign. However, do the seedHits there not form seed values for the isolated scope of the ungapped extender?

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: That's a good point.

cudaungappedextender/samples/device_sample.cu
atadkase: Changed.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
gsneha26: No, `Seeds` are just the subsequences in the query sequence. `SeedHit` is the target+query subsequence. Maybe we have `segmentPair` in place of `Seed` and `HSP` in place of `segmentPair`.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+    ///
+    /// Memcopies to device memory are assumed to be done before this
+    /// function. Output array d_hsp_out must be pre-allocated on device.
+    /// Launches async extension kernel. Filters segments on device
+    /// based on input_hspthresh.
+    StatusType extend_async(const char* d_query, int32_t query_length,
+                            const char* d_target, int32_t target_length,
+                            int32_t score_threshold, Seed* d_seeds,
+                            int32_t num_seeds, ScoredSegment* d_hsp_out,
+                            int32_t* d_num_hsps);
```
gsneha26, line 72: What exactly is `d_num_hsps`?

atadkase: It corresponds to the return value of UngappedExtend here: https://github.com/gsneha26/SegAlign/blob/GenomeWorks/src/seed_filter.cu#L1100
We would ideally like this function to return a completion status as it would be preferable to have all the ungapped extender work complete asynchronously. I'll have to check about the output compression to see if that decision can be made on the device without a CPU decision jump as it causes an unnecessary sync.

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: There's no length associated with the `SeedHit/SegmentPair/Seed` data structure. Is it valid to call it a subsequence/segment in this generalized context (putting aside the segalin usecase)? Also I'm trying to avoid adding the `HighScore/HighScoring` label to current `ScoredSegment` (which can be named as `ScoredSegmentPair` or `SegmentPair`) as it should be apparent from the `score_threshold` parameter that is passed in to keep it general.

gsneha26: Let's just make it `SeedPair` (seed length is the implied length) and `ScoredSegmentPair` then?

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
gsneha26: 1. The return value is of the type `uint32_t`, not an array.
2. So would you like to convert `int32_t* d_num_hsps` into the completion status array?
3. Could you elaborate on the part about the output compression?
3. Also, do you know what is going to be done about the `thrust` function calls in `UngappedExtend()`?

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: Yes. That works!

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
atadkase: 1,2: Oh no it's a single int that is returned by the wrapper to indicate how many ScoredSegments to copy back to host. 
3: The decision to perform output compression can be done on the device by launching the kernels and exiting early with a condition check to see if num_hsps > 0 or by using dynamic parallelism (I'll have to check the perf impact of this as it can use up useful GPU resources). Also thrust kernels can be replaced by our own implementations or by using device_execution_policy (https://thrust.github.io/doc/structthrust_1_1device__execution__policy.html) for masking/unmasking the kernel launch (have to explore that too).

atadkase: To clarify, I'm aiming to make the device pointer version completely asynchronous.

gsneha26: I think we should discuss this part about the post-processing in the device pointer and the corresponding private variables in the meeting/email thread.

cudaungappedextender/samples/segalign_seedfilter.cu
```diff
@@ -0,0 +1,414 @@
...
+#include "seed_filter.h"
+#include <iostream>
+#include <thrust/binary_search.h>
+#include <thrust/device_vector.h>
+#include <thrust/execution_policy.h>
+#include <thrust/iterator/constant_iterator.h>
+#include <thrust/scan.h>
+#include <thrust/unique.h>
+
+#define MAX_SEED_HITS_PER_GB 8388608
```
nvvishanthi, line 11: Can these be `constexpr`s instead of macros?

cudaungappedextender/samples/segalign_seedfilter.cu
```diff
@@ -0,0 +1,414 @@
...
+    d_ref_seq      = (char**)malloc(NUM_DEVICES * sizeof(char*));
+    d_query_seq    = (char**)malloc(BUFFER_DEPTH * NUM_DEVICES * sizeof(char*));
+    d_query_rc_seq = (char**)malloc(BUFFER_DEPTH * NUM_DEVICES * sizeof(char*));
+
+    d_index_table = (uint32_t**)malloc(NUM_DEVICES * sizeof(uint32_t*));
+    d_pos_table   = (uint32_t**)malloc(NUM_DEVICES * sizeof(uint32_t*));
+
+    d_seed_offsets = (uint64_t**)malloc(NUM_DEVICES * sizeof(uint64_t*));
+
+    d_hit_num_array = (uint32_t**)malloc(NUM_DEVICES * sizeof(int32_t*));
```
nvvishanthi, line 154: typo? `sizeof(uint32_t*)`, not that it really matters since they're the same size

cudaungappedextender/include/claraparabricks/genomeworks/cudaungappedextender/cudaungappedextender.hpp
```diff
@@ -0,0 +1,90 @@
...
+    /// score of the segment
+    int32_t score;
+};
+
+/// CUDA Ungapped Extension object
+class UngappedExtender
+{
+public:
+    /// \brief Constructor Prototype
+    UngappedExtender(int* h_sub_mat, int input_xdrop, bool input_noentropy,
```
atadkase, line 48: Should I move the xdrop parameter to the extend_aync call or is it a one-time set-and-forget?

gsneha26: It is one-time set-and-forget.

tijyojwad:
@atadkase should this be closed in favor of #564 ?

atadkase:
Closing as https://github.com/clara-parabricks/GenomeWorks/pull/564 replaces and implements this API.

PR #545, opened 2020-08-31
[cudamapper] Minimizer uses thrust::inclusive_scan

mimaric:
`Minimizer::generate_sketch_elements()` now uses `thrust::inclusive_scan()` to determine positions of reads' minimizers on device, instead of doing it on host. Total number of reads is small and the total number of minimizers still has to be copied to host so performance gains are not big, but it makes the code significantly cleaner.

This PR also removes `cudaStreamSynchronize()` from the end of `Minimizer::generate_sketch_elements()`, but note that freeing `device_buffer`s (for now) contains calls to `cudaStreamSynchronize()`

### DIFF
--- a/cudamapper/src/minimizer.cu
+++ b/cudamapper/src/minimizer.cu
@@ -1055,10 +1055,6 @@ Minimizer::GeneratedSketchElements Minimizer::generate_sketch_elements(DefaultDe
     GW_LOG_INFO("Deallocating {} bytes from read_id_to_windows_section_d", read_id_to_windows_section_d.size() * sizeof(decltype(read_id_to_windows_section_d)::value_type));
     read_id_to_windows_section_d.free();
 
-    // This is not completely necessary, but if removed one has to make sure that the next step
-    // uses the same stream or that sync is done in caller
-    GW_CU_CHECK_ERR(cudaStreamSynchronize(cuda_stream));
-
     return {std::move(representations_compressed_d),
             std::move(rest_compressed_d)};
 }
--- a/cudamapper/src/minimizer.cu
+++ b/cudamapper/src/minimizer.cu
@@ -17,6 +17,9 @@
 #include <limits>
 #include "minimizer.hpp"
 
+#include <thrust/scan.h>
+#include <thrust/execution_policy.h>
+
 namespace claraparabricks
 {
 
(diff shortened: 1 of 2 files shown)


cudamapper/src/minimizer.cu
```diff
@@ -836,32 +839,34 @@ __global__ void find_back_end_minimizers(const std::uint64_t minimizer_size,
...
                                     Minimizer::ReadidPositionDirection* const rest_compressed,
-                                    const ArrayBlock* const read_id_to_compressed_minimizers,
-                                    std::uint32_t offset)
+                                    const std::int64_t* const read_id_to_compressed_minimizers,
+                                    std::uint32_t read_id_of_first_read)
 {
-    const auto& first_input_minimizer  = read_id_to_windows_section[blockIdx.x].first_element_;
-    const auto& first_output_minimizer = read_id_to_compressed_minimizers[blockIdx.x].first_element_;
-    const auto& number_of_minimizers   = read_id_to_compressed_minimizers[blockIdx.x].block_size_;
+    const auto& first_input_minimizer = read_id_to_windows_section[blockIdx.x].first_element_;
```
ahehn-nv, line 861: Optional: while you are at it, could you remove the reference?

cudamapper/src/minimizer.cu
```diff
@@ -906,11 +911,11 @@ Minimizer::GeneratedSketchElements Minimizer::generate_sketch_elements(DefaultDe
     device_buffer<char> window_minimizers_direction_d(total_windows, allocator, cuda_stream);
     GW_LOG_INFO("Allocating {} bytes for window_minimizers_position_in_read_d", total_windows * sizeof(position_in_read_t));
     device_buffer<position_in_read_t> window_minimizers_position_in_read_d(total_windows, allocator, cuda_stream);
-    GW_LOG_INFO("Allocating {} bytes for read_id_to_minimizers_written_d", number_of_reads_to_add * sizeof(std::uint32_t));
-    device_buffer<std::uint32_t> read_id_to_minimizers_written_d(number_of_reads_to_add, allocator, cuda_stream);
+    GW_LOG_INFO("Allocating {} bytes for read_id_to_minimizers_written_d", number_of_reads_to_add * sizeof(std::int64_t));
+    device_buffer<std::int64_t> read_id_to_minimizers_written_d(number_of_reads_to_add, allocator, cuda_stream);
```
ahehn-nv, line 915: Wouldn't `int32_t` be large enough?

cudamapper/src/minimizer.cu
mimaric: done

cudamapper/src/minimizer.cu
mimaric: I was thinking about this as well. This array is also used as an output of `inclusive_sum()` for it should be able to keep the total number of minimizers in an index. I think 31 bits should still be enough for this (otherwise matcher would probably go OOM on every system in foreseeable future), but I'm not sure about mapping to long reference.
As this array is expected to have a few thousand elements I would keep it 64-bit just in case, if you agree

ahehn-nv: Makes sense. Thanks for the additional explanation.

ahehn-nv (approved):

PR #549, opened 2020-09-01
[cudamapper] Let create_index() accept an IndexDescriptor

nvvishanthi:
As discussed in [!536](https://github.com/clara-parabricks/GenomeWorks/pull/536), allow the create_index() function to accept an IndexDescriptor object

closes [!539](https://github.com/clara-parabricks/GenomeWorks/issues/539)


nvvishanthi:
Currently this MR needs !536 before it can be merged. Also, added a new create_index() function instead of replacing the old one, in case we needed to retain compatibility. Not sure if this is required, however

mimaric:
I think we should completely replace the old interface, passing `IndexDescriptor` instead of `read_id` and `number_of_reads` makes way more sense.
This does mean that a larger change is going to be needed, hopefully it will only consist of finding and replacing calls.

nvvishanthi:
@mimaric I removed the old interface and fixed usage. Anything else that needs to be done?

mimaric:
> @mimaric I removed the old interface and fixed usage. Anything else that needs to be done?

I think this change should be propagated to `IndexGPU` as well, i.e. `IndexGPU`'s constructor should take `IndexDescriptor`. That way it would be consistent with `create_index()` and as we already discussed `IndexDescriptor` makes the interface clearer

nvvishanthi:
> I think this change should be propagated to IndexGPU as well, i.e. IndexGPU's constructor should take IndexDescriptor.
Makes sense I will make the change. At first I didn't think to change that because IndexGPU is an internal type, and IndexDescriptor was an "interface" type, but I guess I was overthinking that

nvvishanthi:
Done (I think) and all tests passed :). Had to change some of the tests, so I wanted to check and make sure I didn't alter the behavior to something incorrect

mimaric:
Ok, so ready for re-review?

nvvishanthi:
yup I think so

mimaric (approved):

PR #556, opened 2020-09-11
[cudamapper]: Print in SAM format

nvvishanthi:
Add outputting overlapped sequences + alignments in SAM/BAM format

closes [!335](https://github.com/clara-parabricks/GenomeWorks/issues/335)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -41,6 +41,8 @@ option(gw_profiling "Compile a binary for profiling with NVTX markers." OFF)
 option(gw_enable_caching_allocator "Enable caching allocator." ON)
 option(gw_generate_docs "Generate Doxygen documentation" ON)
 option(gw_cuda_gen_all_arch "ON: Generate optimized CUDA code for all architectures | OFF: for detected architectures only" OFF)
+# Optionally build htslib for SAM/BAM support. Requires autoconf to be installed
+option(gw_build_htslib "Build 3rdparty htslib that allows output in SAM/BAM format" OFF)
 
 # Must be included before others for options value validation
 include(cmake/Utils.cmake)
--- a/README.md
+++ b/README.md
@@ -108,6 +108,10 @@ Minimum requirements -
 4. Python 3.6.7+
 5. CMake (>= 3.10.2)
 
+Optional requirements -
+
+1. autoconf (required to output SAM/BAM files)
+
 ## GenomeWorks Setup
 
 ### Build and Install
@@ -123,6 +127,8 @@ make -j install
 NOTE : The `gw_cuda_gen_all_arch=OFF` option pre-generates optimized code only for the GPU(s) on your system.
 For building a binary that pre-generates opimized code for all common GPU architectures, please remove the option
 or set it to `ON`.
+
+NOTE : (OPTIONAL) To enable outputting overlaps in SAM/BAM format, pass the `gw_build_htslib=ON` option.
 
 ### Package generation
 Package generation puts the libraries, headers and binaries built by the `make` command above
--- a/cmake/3rdparty.cmake
+++ b/cmake/3rdparty.cmake
@@ -50,3 +50,10 @@ set_property(TARGET cub APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CUB_DIR
 
 set(KSEQPP_DIR ${PROJECT_SOURCE_DIR}/3rdparty/kseqpp/src CACHE STRING
     "Path to kseqpp repo")
+
+if (gw_build_htslib)
+    include(cmake/BuildHTSLib.cmake)
+    build_htslib_source()
+else()
+    message(STATUS "Not building htslib, overlap output to SAM & BAM unavailable. Enable with -Dgw_build_htslib=ON")
+endif()
--- a/cudamapper/CMakeLists.txt
+++ b/cudamapper/CMakeLists.txt
@@ -67,7 +67,15 @@ target_include_directories(${MODULE_NAME}
         $<INSTALL_INTERFACE:include>
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
 )
-target_link_libraries(${MODULE_NAME} gwbase gwio cub)
+
+# optionally link against htslib
+if (TARGET htslib_project)
+    add_compile_definitions(GW_BUILD_HTSLIB)
+    add_dependencies(${MODULE_NAME} htslib_project)
+    target_link_libraries(${MODULE_NAME} htslib gwbase gwio cub)
+else()
+     target_link_libraries(${MODULE_NAME} gwbase gwio cub)
+endif()
 target_compile_options(${MODULE_NAME} PRIVATE -Werror)
 
 add_doxygen_source_dir(${CMAKE_CURRENT_SOURCE_DIR}/include)
--- a/cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
+++ b/cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
@@ -20,6 +20,10 @@
 #include <vector>
 #include <string>
 
+#ifdef GW_BUILD_HTSLIB
+#include "sam.h"
+#endif
+
 #include <claraparabricks/genomeworks/cudamapper/types.hpp>
 #include <claraparabricks/genomeworks/cudamapper/index.hpp>
 
@@ -37,6 +41,14 @@ class FastaParser;
 namespace cudamapper
 {
 
+// enum to determine output format
+enum class OutputFormat
+{
+    PAF,
+    SAM,
+    BAM
+};
+
 /// \brief prints overlaps to stdout in <a href="https://github.com/lh3/miniasm/blob/master/PAF.md">PAF format</a>
 /// \param overlaps vector of overlap objects
 /// \param cigars CIGAR strings. Empty vector if none exist
@@ -51,6 +63,25 @@ void print_paf(const std::vector<Overlap>& overlaps,
                int32_t kmer_size,
                std::mutex& write_output_mutex);
 
+#ifdef GW_BUILD_HTSLIB
+/// \brief prints overlaps to stdout in <a href="https://samtools.github.io/hts-specs/SAMv1.pdf">BAM format</a>
+/// \param overlaps vector of overlap objects
+/// \param cigars CIGAR strings. Empty vector if none exist
+/// \param query_parser needed for read names and lengths
+/// \param target_parser needed for read names and lengths
+/// \param format print in either BAM or SAM
+/// \param write_output_mutex mutex that enables exclusive access to output stream
+/// \param argc (optional) number of command line arguments used to generated the @PG CL sections
+/// \param argv (optional) command line arguments used to generated the @PG CL sections
+void print_sam(const std::vector<Overlap>& overlaps,
+               const std::vector<std::string>& cigars,
+               const io::FastaParser& query_parser,
+               const io::FastaParser& target_parser,
+               OutputFormat format,
+               std::mutex& write_output_mutex,
+               int argc     = -1,
+               char* argv[] = nullptr);
+#endif
 /// \brief returns a vector of IndexDescriptors in which the sum of basepairs of all reads in one IndexDescriptor is at most max_basepairs_per_index
 /// If a single read exceeds max_chunk_size it will be placed in its own IndexDescriptor.
 ///
@@ -59,6 +90,7 @@ void print_paf(const std::vector<Overlap>& overlaps,
 /// \return vector of IndexDescriptors
 std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& parser,
                                                       number_of_basepairs_t max_basepairs_per_index = 1000000);
+
 } // namespace cudamapper
 
 } // namespace genomeworks
--- a/cudamapper/samples/sample_cudamapper.cpp
+++ b/cudamapper/samples/sample_cudamapper.cpp
@@ -73,7 +73,8 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                    const std::shared_ptr<const io::FastaParser> query_parser,
                    const std::shared_ptr<const io::FastaParser> target_parser,
                    DefaultDeviceAllocator allocator,
-                   bool print)
+                   const bool print,
+                   const OutputFormat format)
 {
     // extra variables used in print_paf. Note "cigars" are typically found during alignment.
     const std::vector<std::string> cigars(0);
@@ -126,7 +127,19 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                 // print overlaps
                 if (print)
                 {
-                    print_paf(overlaps, cigars, *query_parser, *target_parser, KMER_SIZE, print_mutex);
+#ifdef GW_BUILD_HTSLIB
+                    if (format == OutputFormat::PAF)
+                    {
+#endif
+                        print_paf(overlaps, cigars, *query_parser, *target_parser, KMER_SIZE, print_mutex);
+#ifdef GW_BUILD_HTSLIB
+                    }
+                    // SAM or BAM, depends on type of format
+                    else
+                    {
+                        print_sam(overlaps, cigars, *query_parser, *target_parser, format, print_mutex);
+                    }
+#endif
                 }
             }
         }
@@ -138,11 +151,12 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
 int main(int argc, char** argv)
 {
     // parse command line options
-    int c      = 0;
-    bool help  = false;
-    bool print = false;
+    int c               = 0;
+    bool help           = false;
+    bool print          = false;
+    OutputFormat format = OutputFormat::PAF;
 
-    while ((c = getopt(argc, argv, "hp")) != -1)
+    while ((c = getopt(argc, argv, "hpSB")) != -1)
     {
         switch (c)
         {
@@ -152,6 +166,18 @@ int main(int argc, char** argv)
         case 'h':
             help = true;
             break;
+        case 'S':
+#ifndef GW_BUILD_HTSLIB
+            throw std::runtime_error("ERROR: Argument -S cannot be used without htslib");
+#endif
+            format = OutputFormat::SAM;
+            break;
+        case 'B':
+#ifndef GW_BUILD_HTSLIB
+            throw std::runtime_error("ERROR: Argument -B cannot be used without htslib");
+#endif
+            format = OutputFormat::BAM;
+            break;
         }
     }
 
@@ -160,9 +186,11 @@ int main(int argc, char** argv)
     {
         std::cout << "CUDA Mapper API sample program. Runs minimizer-based approximate mapping" << std::endl;
         std::cout << "Usage:" << std::endl;
-        std::cout << "./sample_cudamapper [-p] [-h]" << std::endl;
+        std::cout << "./sample_cudamapper [-p] [-h] [-S] [-B]" << std::endl;
         std::cout << "-p : Print the overlaps to stdout" << std::endl;
         std::cout << "-h : Print help message" << std::endl;
+        std::cout << "-S : Print in SAM format" << std::endl;
+        std::cout << "-B : Print in BAM format" << std::endl;
         std::exit(0);
     }
 
@@ -183,7 +211,7 @@ int main(int argc, char** argv)
     std::vector<IndexDescriptor> query_index_descriptors  = initialize_batch(query_parser, allocator);
     std::vector<IndexDescriptor> target_index_descriptors = initialize_batch(target_parser, allocator);
 
-    process_batch(query_index_descriptors, target_index_descriptors, query_parser, target_parser, allocator, print);
+    process_batch(query_index_descriptors, target_index_descriptors, query_parser, target_parser, allocator, print, format);
 
     return 0;
 }
--- a/cudamapper/src/application_parameters.cpp
+++ b/cudamapper/src/application_parameters.cpp
@@ -23,6 +23,7 @@
 #include <claraparabricks/genomeworks/cudamapper/index.hpp>
 #include <claraparabricks/genomeworks/io/fasta_parser.hpp>
 #include <claraparabricks/genomeworks/utils/signed_integer_utils.hpp>
+#include <claraparabricks/genomeworks/cudamapper/utils.hpp>
 #include <claraparabricks/genomeworks/version.hpp>
 
 namespace claraparabricks
@@ -59,12 +60,13 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
         {"help", no_argument, 0, 'h'},
     };
 
-    std::string optstring = "k:w:d:m:i:t:F:a:r:l:b:z:RDQ:q:C:c:vh";
+    std::string optstring = "k:w:d:m:i:t:F:a:r:l:b:z:RDQ:q:C:c:BSvh";
 
     bool target_indices_in_host_memory_set   = false;
     bool target_indices_in_device_memory_set = false;
     bool custom_filtering_parameter          = false;
     int32_t argument                         = 0;
+    format                                   = OutputFormat::PAF;
     while ((argument = getopt_long(argc, argv, optstring.c_str(), options, nullptr)) != -1)
     {
         switch (argument)
@@ -130,6 +132,18 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
         case 'c':
             target_indices_in_device_memory     = std::stoi(optarg);
             target_indices_in_device_memory_set = true;
+            break;
+        case 'S':
+#ifndef GW_BUILD_HTSLIB
+            throw std::runtime_error("ERROR: Argument -S cannot be used without htslib");
+#endif
+            format = OutputFormat::SAM;
+            break;
+        case 'B':
+#ifndef GW_BUILD_HTSLIB
+            throw std::runtime_error("ERROR: Argument -B cannot be used without htslib");
+#endif
+            format = OutputFormat::BAM;
             break;
         case 'v':
             print_version();
@@ -354,6 +368,12 @@ void ApplicationParameters::help(int32_t exit_code)
               << R"(
         -c, --target-indices-in-device-memory
             number of target indices to keep in device memory [5])"
+              << R"(
+        -S
+            print overlaps in SAM format"
+              << R"(
+        -B
+            print overlaps in BAM format"
               << R"(
         -v, --version
             Version information)"
--- a/cudamapper/src/main.cu
+++ b/cudamapper/src/main.cu
@@ -370,15 +370,32 @@ void postprocess_and_write_thread_function(const int32_t device_id,
                                                 0.5);
             }
 
-            // write to output
+            // write to output PAF, SAM, or BAM format
             {
-                GW_NVTX_RANGE(profiler, "main::postprocess_and_write_thread::print_paf");
-                print_paf(overlaps,
-                          cigars,
-                          *application_parameters.query_parser,
-                          *application_parameters.target_parser,
-                          application_parameters.kmer_size,
-                          output_mutex);
+                GW_NVTX_RANGE(profiler, "main::postprocess_and_write_thread::print_function");
+#ifdef GW_BUILD_HTSLIB
+                if (application_parameters.format == OutputFormat::PAF)
+                {
+#endif
+                    print_paf(overlaps,
+                              cigars,
+                              *application_parameters.query_parser,
+                              *application_parameters.target_parser,
+                              application_parameters.kmer_size,
+                              output_mutex);
+#ifdef GW_BUILD_HTSLIB
+                }
+                // SAM or BAM, depends on type of format
+                else
+                {
+                    print_sam(overlaps,
+                              cigars,
+                              *application_parameters.query_parser,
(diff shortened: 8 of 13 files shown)


nvvishanthi:
CC @tijyojwad

nvvishanthi:
Two things to note here, which I'm kind of stuck on:
1) I was having some linker errors earlier on which I haven't got around to resolving :(. I think it's a cmake issue, but I'm not sure how to get it to work right. Probably also want to dynamically link it instead of statically link like I'm doing now
2) I found that the latest samtools master branch doesn't actually compile when I tried build using libhts-dev, so we _might_ want to consider submoduling after all. Or maybe using a newer version

ahehn-nv:
I thought we decided to use a system installation of htslib instead of a submodule. What changed?

ahehn-nv:
Just saw your latest comment. I'm a bit confused - I thought we needed samtools' libhts, not samtools itself. Why do we need to compile the samtools?

nvvishanthi:
@ahehn-nv I was having some linker issues which I think was due to my incorrect cmake so I tried to work around it (this is totally fixable, I just haven't figured it out what's wrong yet)). And then see 2) in my above comment about samtools not building properly with libhts-dev; not sure if we actually care about that or not, but I was able to get things working if I used htslib built from source

nvvishanthi:
> I thought we needed samtools' libhts, not samtools itself. Why do we need to compile the samtools?

You're totally right, we don't need samtools itself. It was just a side-note, I suppose? I figured that since we wanted to be compatible with samtools, it would be bad if a library we were using was incompatible. Maybe it's not actually a problem. I can change to use libhts-dev after I figure out the linking issues I'm having

nvvishanthi:
rerun-tests

cudamapper/src/utils.cpp
```diff
@@ -150,6 +156,182 @@ std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& par
...
+    }
+    // odd number of bases, so we have one left over
+    if (i < len)
+    {
+        *bam_seq_start++ = L[(unsigned char)seq[i]] << 4;
+    }
+
+    alignment->core.l_qseq = len;
+
+    *bam_seq_start = 0xff; // if htslib encounters this at the start of the QUAL section, it writes default value (*)
```
tijyojwad, line 331: let's move this into an `encode_quality` function just to be explicit about it

tijyojwad (changes requested):

cudamapper/include/claraparabricks/genomeworks/cudamapper/utils.hpp
```diff
@@ -59,6 +60,15 @@ void print_paf(const std::vector<Overlap>& overlaps,
 /// \return vector of IndexDescriptors
 std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& parser,
                                                       number_of_basepairs_t max_basepairs_per_index = 1000000);
+/// \brief prints overlaps to stdout in <a href="https://samtools.github.io/hts-specs/SAMv1.pdf">BAM format</a>
```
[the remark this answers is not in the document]
nvvishanthi: fixed both

cudamapper/src/utils.cpp
```diff
@@ -31,6 +34,9 @@ namespace genomeworks
 namespace cudamapper
 {
 
+static inline void encode_seq(bam1_t* alignment, const std::string& seq);
```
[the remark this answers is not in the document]
nvvishanthi: You're right we don't - fixed

cudamapper/src/utils.cpp
nvvishanthi: done

cudamapper/src/utils.cpp
```diff
@@ -150,6 +156,182 @@ std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& par
...
+
+    const size_t len = seq.length();
+
+    // encode bases as nibbles and write them into the alignment struct 2 bases at a time
+    // 4 high bits are first base, 4 low bits are second base
+    uint8_t* bam_seq_start = bam_get_seq(alignment);
+    int i                  = 0;
+    for (i; (i + 1) < len; i += 2)
+    {
+        *bam_seq_start++ = (L[(unsigned char)seq[i]] << 4) + L[(unsigned char)seq[i + 1]];
```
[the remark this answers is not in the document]
nvvishanthi: good catch. missed changing that to static_cast. fixed

cudamapper/src/utils.cpp
```diff
@@ -150,6 +156,182 @@ std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& par
...
+    return;
+}
+
+static inline void encode_cigar(bam1_t* alignment, const std::string& cigar)
+{
+    uint32_t* bam_cigar_start = bam_get_cigar(alignment);
+    memcpy(bam_cigar_start, cigar.c_str(), cigar.length());
+    alignment->core.n_cigar = cigar.length();
+
+    return;
```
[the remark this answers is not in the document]
nvvishanthi: I didn't have anything special in mind for error handling besides checking for return codes and reporting if any of the IO or struct creation fails. Are there special cases that need to be addressed?

cudamapper/src/utils.cpp
```diff
@@ -150,6 +156,182 @@ std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& par
...
+        bam1_t* alignment                  = bam_init1();
+
+        // write QNAME field
+        bam_set_qname(alignment, query_read_name.c_str());
+
+        // At this point we need to handle mem allocs ourselves, find necessary data size
+        size_t total_size = 0;
+        if (!cigars.empty())
+        {
+            total_size = alignment->core.l_qname + alignment->core.l_extranul + cigars[i].length() + ((query_parser.get_sequence_by_id(overlaps[i].query_read_id_).seq.length() + 1) / 2);
```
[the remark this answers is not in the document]
nvvishanthi: fixed

cudamapper/CMakeLists.txt
```diff
@@ -67,7 +67,9 @@ target_include_directories(${MODULE_NAME}
         $<INSTALL_INTERFACE:include>
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
 )
-target_link_libraries(${MODULE_NAME} gwbase gwio cub)
+#target_link_libraries(${MODULE_NAME} gwbase gwio cub ${HTSLIB_DIR}/build/lib/libhts.a)
+add_dependencies(${MODULE_NAME} htslib)
+target_link_libraries(${MODULE_NAME} gwbase gwio cub ${CMAKE_INSTALL_PREFIX}/htslib/lib/libhts.a)
```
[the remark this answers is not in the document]
nvvishanthi: oversight from old commits, thanks for catching this. Fixed

CMakeLists.txt
```diff
@@ -77,13 +77,24 @@ include(cmake/Packaging.cmake)
 include(cmake/CondaEnv.cmake)
 include(cmake/GitVersion.cmake)
 
+#ExternalProject_Get_Property(hts install_dir)
```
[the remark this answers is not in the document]
nvvishanthi: fixed

cudamapper/samples/CMakeLists.txt
```diff
@@ -26,6 +26,7 @@ add_executable(${MODULE_NAME}
 target_link_libraries(${MODULE_NAME}
                       cudamapper
                       )
+                      #${HTSLIB_DIR}/build/lib/libhts.a
```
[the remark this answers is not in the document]
nvvishanthi: Right, this is an artifact of me messing around with cmake. Since it's linked in libcudamapper.a, doesn't need to be re-linked here

cudamapper/src/main.cu
```diff
@@ -370,15 +370,22 @@ void postprocess_and_write_thread_function(const int32_t device_id,
...
-                print_paf(overlaps,
-                          cigars,
-                          *application_parameters.query_parser,
-                          *application_parameters.target_parser,
-                          application_parameters.kmer_size,
-                          output_mutex);
+                GW_NVTX_RANGE(profiler, "main::postprocess_and_write_thread::print_function");
+                if (application_parameters.sam_format)
+                {
+                    print_sam(overlaps, cigars, *application_parameters.query_parser, *application_parameters.target_parser, output_mutex);
```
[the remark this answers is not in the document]
nvvishanthi: done

cudamapper/src/utils.cpp
```diff
@@ -150,6 +156,182 @@ std::vector<IndexDescriptor> group_reads_into_indices(const io::FastaParser& par
...
+        {
+            fprintf(stderr, "print_sam: could not add header value");
+        }
+    }
+
+    // If we have the information to write out the CL section, do it
+    if ((argc != -1) && (argv != nullptr))
+    {
+
+        char* command_line = stringify_argv(argc, argv);
```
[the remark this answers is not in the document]
nvvishanthi: Did for all resources (file, header, command_line, and alignment struct)

cmake/3rdparty.cmake
```diff
@@ -50,3 +50,34 @@ set_property(TARGET cub APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CUB_DIR
...
+
+#if (NOT HTSLIB_LIBNAME)
+#    message("htslib is missing, SAM/BAM format will not be available. Please instal libhts-dev for support")
+#endif()
+
+### Manually build htslib ###
+set(HTSLIB_DIR ${PROJECT_SOURCE_DIR}/3rdparty/htslib/ CACHE STRING
+    "Path to htslib repo")
+set(MAKE_COMMAND "$(MAKE)")
+set(HTSLIB_INSTALL ${MAKE_COMMAND} install prefix=${CMAKE_INSTALL_PREFIX}/htslib)
```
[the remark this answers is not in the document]
nvvishanthi: moved to <build-dir>/3rdparty/htslib

nvvishanthi:
@tijyojwad CI seems to fail because it's missing `autoheader` and `autoconf`. `sudo apt install autoconf` needs to be run first. Could these be added to the CI machines? I can add them to the System Requirements docs if acceptable

CMakeLists.txt
```diff
@@ -41,6 +41,8 @@ option(gw_profiling "Compile a binary for profiling with NVTX markers." OFF)
 option(gw_enable_caching_allocator "Enable caching allocator." ON)
 option(gw_generate_docs "Generate Doxygen documentation" ON)
 option(gw_cuda_gen_all_arch "ON: Generate optimized CUDA code for all architectures | OFF: for detected architectures only" OFF)
+# Optionally build htslib for SAM/BAM support. Requires autoconf to be installed
+option(gw_build_htslib "Build 3rdparty htslib that allows output in SAM/BAM format" OFF)
```
nvvishanthi, line 45: (See others changes in .cmake files and README.md, as well as the ifdef guards in utils files, main.cu, and sample_cudamapper.cpp that relate to this) @tijyojwad I wasn't sure if this was the _best_ idea, but I made the htslib components optional since they require extra packages to be installed (autoconf). We talked about this before, but I'm not sure how applicable it still is. In any case, I added it to the build & install documentation

cmake/3rdparty.cmake
```diff
@@ -50,3 +50,34 @@ set_property(TARGET cub APPEND PROPERTY INTERFACE_INCLUDE_DIRECTORIES "${CUB_DIR
...
 set(KSEQPP_DIR ${PROJECT_SOURCE_DIR}/3rdparty/kseqpp/src CACHE STRING
     "Path to kseqpp repo")
+
+#find_library(HTSLIB_LIBNAME "hts" "/usr/lib/x86_64-linux-gnu" "/usr/local/lib")
+
+#if (NOT HTSLIB_LIBNAME)
+#    message("htslib is missing, SAM/BAM format will not be available. Please instal libhts-dev for support")
+#endif()
+
+### Manually build htslib ###
```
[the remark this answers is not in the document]
nvvishanthi: fixed

CMakeLists.txt
tijyojwad: Yeah I like this optional support capability! Let's keep it

README.md
```diff
@@ -108,6 +108,10 @@ Minimum requirements -
 4. Python 3.6.7+
 5. CMake (>= 3.10.2)
 
+Optional requirements -
+
+1. autoconf (required to build optional htslib package)
```
tijyojwad, line 113: can this be changed to `required to output SAM/BAM files`?

cudamapper/CMakeLists.txt
```diff
@@ -67,7 +67,15 @@ target_include_directories(${MODULE_NAME}
         $<INSTALL_INTERFACE:include>
         $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>
 )
-target_link_libraries(${MODULE_NAME} gwbase gwio cub)
+
+# optionally link against htslib
+if (gw_build_htslib)
```
tijyojwad, line 72: let's check this against the availability of the `htslib` target. if we change the the htslib is built or accessed, then this part of the code won't have to change

cudamapper/CMakeLists.txt
```diff
@@ -67,7 +67,15 @@ target_include_directories(${MODULE_NAME}
...
+    add_compile_definitions(GW_BUILD_HTSLIB)
```
tijyojwad, line 73: I think this is the newer way to handling compile definitions - https://cmake.org/cmake/help/latest/command/target_compile_definitions.html

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -126,7 +127,16 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                 // print overlaps
                 if (print)
                 {
-                    print_paf(overlaps, cigars, *query_parser, *target_parser, KMER_SIZE, print_mutex);
+                    if (!sam_format)
```
tijyojwad, line 130: if `sam_format` is selected, but it's not built with sam support, looks like it might silently not print anything?

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -73,7 +73,8 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                    const std::shared_ptr<const io::FastaParser> query_parser,
                    const std::shared_ptr<const io::FastaParser> target_parser,
                    DefaultDeviceAllocator allocator,
-                   bool print)
+                   bool print,
+                   bool sam_format)
```
tijyojwad, line 77: could we update this bool to be an enum for which type of format we want to print `PAF`, `SAM`, `BAM`?

cudamapper/src/main.cu
```diff
@@ -370,15 +370,28 @@ void postprocess_and_write_thread_function(const int32_t device_id,
...
+                if (!application_parameters.sam_format)
+                {
+                    print_paf(overlaps,
+                              cigars,
+                              *application_parameters.query_parser,
+                              *application_parameters.target_parser,
+                              application_parameters.kmer_size,
+                              output_mutex);
+                }
+#ifdef GW_BUILD_HTSLIB
```
tijyojwad, line 385: same here, we will need some error mechanism to report if sam is selected but HTSLIB is not available

cudamapper/src/utils.cpp
```diff
@@ -115,6 +121,190 @@ void print_paf(const std::vector<Overlap>& overlaps,
...
+               const std::vector<std::string>& cigars,
+               const io::FastaParser& query_parser,
+               const io::FastaParser& target_parser,
+               std::mutex& write_output_mutex,
+               int argc,
+               char* argv[])
+{
+    GW_NVTX_RANGE(profiler, "print_sam::formatting_output");
+
+    // Assume output is stdout, and assume SAM output by default
```
tijyojwad, line 194: how do we choose between sam and bam here?

tijyojwad (changes requested):

cudamapper/src/utils.cpp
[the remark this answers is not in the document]
tijyojwad: yep that's what I meant, and it looks like they're being handled already so all good!

README.md
nvvishanthi: fixed

cudamapper/CMakeLists.txt
nvvishanthi: good idea. Fixed

cudamapper/CMakeLists.txt
nvvishanthi: So this definition is used in multiple targets (libcudamapper, sample_cudamapper, and cudamapper-bin), so I think defining this for the module gives us the benefit of not having to add this define for each target. That said, I defer to you on this (it's just a few more lines of cmake here and there)

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: fixed by adding ifdef guards, printing a warning, and exiting on use without building with sam/bam support

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: 👍 done

cudamapper/src/main.cu
nvvishanthi: see above reply

cudamapper/src/utils.cpp
nvvishanthi: SAM vs BAM depends on adding a parameter to `sam_open()`. Added in latest commit that adds the output enum

cudamapper/CMakeLists.txt
tijyojwad: okay I see, in that case this is fine. Thanks!

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -152,6 +164,20 @@ int main(int argc, char** argv)
...
+        case 'S':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -S cannot be used without htslib" << std::endl;
+            exit(1);
+#endif
+            format = OutputFormat::SAM;
+            break;
+        case 'B':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -S cannot be used without htslib" << std::endl;
```
tijyojwad, line 176: typo: `-S` -> `-B`

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -152,6 +164,20 @@ int main(int argc, char** argv)
         case 'h':
             help = true;
             break;
+        case 'S':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -S cannot be used without htslib" << std::endl;
+            exit(1);
```
tijyojwad, line 170: let's throw `std::runtime_error`

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -152,6 +164,20 @@ int main(int argc, char** argv)
...
+#endif
+            format = OutputFormat::SAM;
+            break;
+        case 'B':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -S cannot be used without htslib" << std::endl;
+            exit(1);
```
tijyojwad, line 177: `throw std::runtime_error`

cudamapper/src/application_parameters.cpp
```diff
@@ -131,6 +133,20 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
             target_indices_in_device_memory     = std::stoi(optarg);
             target_indices_in_device_memory_set = true;
             break;
+        case 'S':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -S cannot be used without htslib" << std::endl;
+            exit(1);
```
tijyojwad, line 139: `throw std::runtime_error`

cudamapper/src/application_parameters.cpp
```diff
@@ -131,6 +133,20 @@ ApplicationParameters::ApplicationParameters(int argc, char* argv[])
...
+#endif
+            format = OutputFormat::SAM;
+            break;
+        case 'B':
+#ifndef GW_BUILD_HTSLIB
+            std::cerr << "ERROR: Argument -B cannot be used without htslib" << std::endl;
+            exit(1);
```
tijyojwad, line 146: `throw std::runtime_error`

tijyojwad (changes requested):

cudamapper/samples/sample_cudamapper.cpp
tijyojwad: Yep that's good!

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -73,7 +73,8 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                    const std::shared_ptr<const io::FastaParser> query_parser,
                    const std::shared_ptr<const io::FastaParser> target_parser,
                    DefaultDeviceAllocator allocator,
-                   bool print)
+                   bool print,
+                   OutputFormat format)
```
mimaric, line 77: `const` to prevent accidental edits and possibly make it easier for compiler to optimize?

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -126,7 +127,17 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
                 // print overlaps
                 if (print)
                 {
-                    print_paf(overlaps, cigars, *query_parser, *target_parser, KMER_SIZE, print_mutex);
+                    if (format == OutputFormat::PAF)
```
mimaric, line 130: This `if` could also be put in an `#ifdef GW_BUILD_HTSLIB`

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -138,11 +149,12 @@ void process_batch(std::vector<IndexDescriptor>& query_index_descriptors,
...
-    int c      = 0;
-    bool help  = false;
-    bool print = false;
+    int c               = 0;
+    bool help           = false;
+    bool print          = false;
+    OutputFormat format = OutputFormat::PAF;
 
-    while ((c = getopt(argc, argv, "hp")) != -1)
+    while ((c = getopt(argc, argv, "hpSB")) != -1)
```
mimaric, line 157: How is `p` used?

cudamapper/src/main.cu
```diff
@@ -370,15 +370,30 @@ void postprocess_and_write_thread_function(const int32_t device_id,
...
             {
-                GW_NVTX_RANGE(profiler, "main::postprocess_and_write_thread::print_paf");
-                print_paf(overlaps,
-                          cigars,
-                          *application_parameters.query_parser,
-                          *application_parameters.target_parser,
-                          application_parameters.kmer_size,
-                          output_mutex);
+                GW_NVTX_RANGE(profiler, "main::postprocess_and_write_thread::print_function");
+                if (application_parameters.format == OutputFormat::PAF)
```
mimaric, line 376: Again, this `if` could also be put in `#ifdef GW_BUILD_HTSLIB`

cudamapper/src/utils.cpp
```diff
@@ -115,6 +121,193 @@ void print_paf(const std::vector<Overlap>& overlaps,
     }
 }
 
+#ifdef GW_BUILD_HTSLIB
+static inline void encode_cigar(bam1_t* alignment, const std::string& cigar)
```
mimaric, line 125: Most of these functions could be put in an anonymous namespace as they are only used in this file

mimaric, line 125: Pointers like `bam1_t* alignment` could still be constant pointers to non-constant values: `bam1_t* const alignment`

cudamapper/src/utils.cpp
```diff
@@ -115,6 +121,193 @@ void print_paf(const std::vector<Overlap>& overlaps,
...
+    // if htslib encounters 0xff at the start of the QUAL section, it writes default value (*)
+    auto bam_qual_start = bam_get_qual(alignment);
+    *bam_qual_start     = 0xff;
+}
+
+void print_sam(const std::vector<Overlap>& overlaps,
+               const std::vector<std::string>& cigars,
+               const io::FastaParser& query_parser,
+               const io::FastaParser& target_parser,
+               OutputFormat format,
```
mimaric, line 188: `const` to prevent accidental changes

cudamapper/src/utils.cpp
```diff
@@ -115,6 +121,193 @@ void print_paf(const std::vector<Overlap>& overlaps,
...
+    using samFilePtr_t = std::unique_ptr<samFile, decltype(&hts_close)>;
+
+    auto file   = (format == OutputFormat::SAM) ? samFilePtr_t(sam_open("-", "wh"), &hts_close) : samFilePtr_t(sam_open("-", "bwh"), &hts_close);
+    auto header = std::unique_ptr<bam_hdr_t, decltype(&sam_hdr_destroy)>(sam_hdr_init(), &sam_hdr_destroy);
+    int result  = 0;
+
+    if (file == nullptr)
+    {
+        fprintf(stderr, "print_sam: could not open file for writing");
+        return;
```
mimaric, line 207: Should we abort at this point or throw an exception?

CMakeLists.txt
mimaric, line 45: Do we make sure CI runs with it on?

mimaric (commented):
Suggested some optional changes, but other than that look good

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: fixed

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: fixed

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: fixed

cudamapper/src/application_parameters.cpp
nvvishanthi: fixed

cudamapper/src/application_parameters.cpp
nvvishanthi: fixed

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: good idea, fixed.

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: I'm a little one the fence about this one since that field will always be populated to the default  value of OutputFormat::PAF, so it's relatively harmless. That said, I changed it

cudamapper/samples/sample_cudamapper.cpp
nvvishanthi: p is as a switch to print to stdout (or print nothing at all). It's a flag that's used in all GW samples

cudamapper/src/main.cu
nvvishanthi: same comment as above, but fixed

cudamapper/src/utils.cpp
nvvishanthi: ah true, done

nvvishanthi: done. moved all internal functions to anonymous namespace

cudamapper/src/utils.cpp
nvvishanthi: done

cudamapper/src/utils.cpp
nvvishanthi: I decided to throw an exception here. It _could_ be caught by the calling application and handled. One use-case I could think of it if the disk that's being written to is full or something. Also, in the future, it's plausible that we operate directly on files instead of stdout > file, so handling which file to write to could be an argument that could be changed when an exception is encountered

CMakeLists.txt
nvvishanthi: (See https://github.com/clara-parabricks/GenomeWorks/pull/556#issuecomment-703031099)
Unfortunately, CI doesn't have the required build dependencies for htslib, so it always fails during configuration :(. Is it possible for these apt dependencies be added to CI?

o8191: @nvvishanthi I created a PR(https://github.com/clara-parabricks/gpuci-build-environment/pull/7) for adding `autoconf` to our images through Anaconda. Although, I don't think this is really needed. I was able to build & install `htslib` from a git clone by following the Basic Installation steps. https://github.com/samtools/htslib/blob/develop/INSTALL#L63-L68

nvvishanthi: @o8191 what did you have to change to get it to build? I couldn't get it to work in CI, with 1) `./configure <args to disable some extra compression libs>` 2) `make` 3) `make install`. It said that the configure script was missing - which is what I expected since there is no autoconf.

This is from the build instructions (about the `autoheader`/`autoconf` usage):
> This step is only needed if configure.ac has been changed, or if configure
does not exist (for example, when building from a git clone).

Doesn't this apply since we're building after a git clone?

nvvishanthi:
rerun-tests

nvvishanthi:
rerun tests

CMakeLists.txt
nvvishanthi: @mimaric MR now passes in CI with it on

tijyojwad (approved):

Issue #554, opened 2020-09-09
GenomeWorks compiling error


asantat:
Hi!
I am working around two days trying to compile GenomeWorks which included into racon genome assembler.
GCC = 9.3.0
cmake = 3.18.20200908-g1d74a64
make = 4.3
nvcc = 11.0
cuDNN = 8.0.2
nvidia driver = 450.51.06

CMakeCache.txt was manually edited with including `CMAKE_CXX_FLAGS:STRING=-DFMT_USE_USER_DEFINED_LITERALS=0` string
the error occured whyle the most of work was done: 
`
[ 81%] Linking CXX static library ../../lib/libcudaaligner.a
Reaping winning child 0x55c2e3d8eaf0 PID 14870 
Live child 0x55c2e3d8eaf0 (lib/libcudaaligner.a) PID 14872 
Reaping winning child 0x55c2e3d8eaf0 PID 14872 
Live child 0x55c2e3d8eaf0 (lib/libcudaaligner.a) PID 14874 
Reaping winning child 0x55c2e3d8eaf0 PID 14874 
Removing child 0x55c2e3d8eaf0 PID 14874 from chain.
Considering target file 'GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/build'.
 File 'GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/build' does not exist.
  Considering target file 'lib/libcudaaligner.a'.
  File 'lib/libcudaaligner.a' was considered already.
 Finished prerequisites of target file 'GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/build'.
Must remake target 'GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/build'.
Successfully remade target file 'GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/build'.
Reaping winning child 0x558a0914daf0 PID 14802 
Live child 0x558a0914daf0 (GenomeWorks/cudaaligner/CMakeFiles/cudaaligner.dir/all) PID 14879 
[ 81%] Built target cudaaligner
Reaping winning child 0x558a0914daf0 PID 14879 
Removing child 0x558a0914daf0 PID 14879 from chain.
Reaping losing child 0x560a19aabcb0 PID 14245 
make: *** [Makefile:171: all] Error 2
Removing child 0x560a19aabcb0 PID 14245 from chain.
`

I updated cmake and make, but compilation still aborting.
[...]

PR #561, opened 2020-09-16
[common] Fixed a compiler warnings and a problem with CUDA 11

ahehn-nv:
Hotfix for #554 and experimental CUDA 11 support.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -40,7 +40,7 @@ option(gw_enable_cudapoa_nw_print "Enable verbose prints within cudapoa NW kerne
 option(gw_profiling "Compile a binary for profiling with NVTX markers." OFF)
 option(gw_enable_caching_allocator "Enable caching allocator." ON)
 option(gw_generate_docs "Generate Doxygen documentation" ON)
-option(gw_cuda_gen_all_arch "ON: Generate optimized CUDA code for all architectures | OFF: for detected architectures only" ON)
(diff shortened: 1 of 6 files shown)


ahehn-nv:
@tijyojwad In order to not have too many hotfix releases, we might want to combine this with a fix for CUDA 11.
I don't have time to work on that right now, but I'm happy to merge a PR against my branch.

tijyojwad:
agreed, @ahehn-nv - @atadkase can you create a quick PR with the fix to enable CUDA 11 into this branch? We decided to make the `gw_cuda_gen_all_arch` `OFF` by default in CMakeLists.txt. This would need an update to the CMakeLists.txt and a change to `ci/common/build-test-sdk.sh` with line 30 updated to include `gw_cuda_gen_all_arch=ON`. Was there anything else that's needed to enable CUDA 11?

tijyojwad (approved):

ahehn-nv:
@tijyojwad Are we sure these changes won't break the python packages - if we only compile for one architecture by default?

tijyojwad:
> @tijyojwad Are we sure these changes won't break the python packages - if we only compile for one architecture by default?

Glad you mentioned this again. I was under the impression the build-test-sdk.sh update took care of cmake flags for python builds as well, but then I double checked and it was actually in a different place. So I just pushed a fix for that. Thanks for bringing it up!

PR #563, opened 2020-09-17
[cudamapper] Filtering for self mappings, identical overlaps, and highly-similar overlaps

edawson:
This PR brings in filtering strategies that mimic those of minimap2.
- [X] Self mappings: minimap2 does not output self mappings (where an overlap between a read and itself covers more than some very high fraction of that read's length). This can be disabled with a command line flag (`-S`).
- [x] Identical overlaps: remove identical overlaps within some number of indices.
- [x] Highly-similar overlaps: remove overlaps that have a percent similarity (defined as the reciprocal overlap) higher than some percentage threshold.


cudamapper/src/overlapper.cpp
```diff
@@ -98,6 +98,51 @@ bool overlaps_mergable(const claraparabricks::genomeworks::cudamapper::Overlap o
...
+           a.target_end_position_in_read_ == b.target_end_position_in_read_;
+}
+
+inline double percent_reciprocal_overlap(const claraparabricks::genomeworks::cudamapper::Overlap& a, const claraparabricks::genomeworks::cudamapper::Overlap& b)
+{
+    if (a.query_read_id_ != b.query_read_id_ || a.target_read_id_ != b.target_read_id_ || a.relative_strand != b.relative_strand)
+    {
+        return 0.0;
+    }
+    int32_t query_overlap = std::min(a.query_end_position_in_read_, b.query_end_position_in_read_) - std::max(a.query_start_position_in_read_, b.query_start_position_in_read_);
```
mimaric, line 117: This and similar should be `position_in_read_t` or even better `number_of_basepairs_t`

mimaric (changes requested):

PR #569, opened 2020-09-23
[sdk] Update CUB submodule to new URL and version 1.9.9

ahehn-nv:


mimaric:
Should we still download CUB when using CUDA 11? I guess one cannot know CUDA version upfront, but in that case we probably should not have downloaded CUB in the include path

ahehn-nv:
Some logic in CMake to use the submodule only for CUDA <11.0 makes sense.
However, for CUDA 11 it shouldn't make much difference, since I updated the submodule to the version shipped with CUDA 11.

mimaric:
I agree. I would suggest linking this PR to issue #570, but not closing it until that is done in a separate PR

tijyojwad:
@ahehn-nv - the CUDA 9.0 test seems to have failed. Can you have a look? perhaps we should make this change once we remove CUDA 9.0 support?

ahehn-nv:
Yes, let's postpone this. CUDA 9 compilation fails due to a bug in CUB - see https://github.com/NVIDIA/cub/pull/207.
Since there's also a ticket for this task, I'll close this PR.

mimaric:
PR https://github.com/NVIDIA/cub/pull/207 has been merged into CUB. The bug it solved was preventing us from using latest CUB version with CUDA 9.
Reopening this PR.
CUB 1.11.0 (RC available) bring big improvement to radix sort which in turn improves the performance of cudamapper.

mimaric:
The tests pass now. I suggest waiting for non-rc version of CUB 1.11.0 and merging then

tijyojwad:
@mimaric @ahehn-nv - CUDA 9 has been removed from our CI now.

mimaric:
@ahehn-nv CUB 1.11.0 has been released (https://github.com/NVIDIA/cub/releases/tag/1.11.0). Can you update this PR?

ahehn-nv:
The PR is already up-to-date. CUB's 1.11.0-rc1 and 1.11.0 tags point to the same commit.

mimaric (approved):

PR #581, opened 2020-10-01
[cudapoa] Unify NW kernels

r-mafi:
over time, numerous versions of Needleman-Wunsch algorithm for different banding modes have been developed. In this PR, we unify adaptive-banded and static-banded versions into a same device kernel.
other changes:
- added adaptive NW with traceback buffer.
- updated BandMode to include `adaptive_band_traceback`
- due to larger register usage in `adaptive_band_traceback`, setting register count to 64 results in large memory spills and reduces performance. Therefore, traceback NW kernels are called with a different launch bounds settings to limit registers to 72.


r-mafi:
rerun tests

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -30,7 +29,8 @@
 #include <claraparabricks/genomeworks/cudapoa/batch.hpp>
 
 // considering 65536 register file size, __launch_bounds__(1024) translates to max 64 register usage
-#define GW_POA_KERNELS_MAX_THREADS_PER_BLOCK 1024
+#define MAX_THREADS_PER_BLOCK_64 1024
```
tijyojwad, line 32: for macros it's good to add a prefix to the name to clarify the scope of usage. hence the `GW_POA_KERNELS` prefix. let's add it (or some version of that) back. and also add `REGISTERS` as a suffix (`_64_REGISTERS`)

cudapoa/src/cudapoa_structs.cuh
```diff
@@ -45,6 +49,11 @@
 #define CUDAPOA_KERNEL_ERROR_ENCOUNTERED UINT8_MAX
 #define CUDAPOA_KERNEL_NOERROR_ENCOUNTERED 0
 
+// NW return error codes
+#define NW_BACKTRACKING_LOOP_FAILED -1
```
tijyojwad, line 53: let's add the `CUDAPOA_KERNEL` prefix to each of these macros

cudapoa/src/cudapoa_kernels.cuh
```diff
@@ -371,7 +342,347 @@ __launch_bounds__(GW_POA_KERNELS_MAX_THREADS_PER_BLOCK)
...
+            return;
+        }
+    }
+}
+
+// Similar to generatePOAKernel(), but with a different launch bounds setting, used to call kernels using traceback buffer
+// i.e. for BM == BandMode::static_band_traceback or BM == BandMode::adaptive_band_traceback
+template <typename ScoreT, typename SizeT, typename TraceT, bool MSA = false, BandMode BM = BandMode::static_band_traceback>
+__launch_bounds__(MAX_THREADS_PER_BLOCK_72)
+    __global__ void generatePOAKernelTB(uint8_t* consensus_d,
```
tijyojwad, line 437: I think you're already explained this to me but I am not able to recall the exact reason now - if we were to combine the two kernels with templatization, is that also causing the register count to go up?

the main reason I'm asking is because a lot of the initial indexing code is copied over, and I'm concerned that they might go out of sync if one was updated

tijyojwad (changes requested):

cudapoa/src/cudapoa_kernels.cuh
r-mafi: I agree, it will be difficult to maintain the two copies of `generatePOAKernel()` and `generatePOAKernelTB()`. I had tried once and it added to the register count, but the code has changed since then, will retry.

cudapoa/src/cudapoa_structs.cuh
r-mafi: done.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: seems a branch of optimization efforts was applied to minimize the length of this macro! :D changed back and added REGISTERS to be more clear.

cudapoa/src/cudapoa_kernels.cuh
r-mafi: retry was successful! done.

cudapoa/src/cudapoa_structs.cuh
```diff
@@ -37,6 +37,10 @@
 // ad-hoc maximum band-width defined in adaptive banding
 #define CUDAPOA_MAX_ADAPTIVE_BAND_WIDTH 1536
 
+// rerun codes used in adaptive banding
+#define SHIFT_ADAPTIVE_BAND_TO_LEFT -10
```
tijyojwad, line 41: also prefix these macros with `CUDAPOA_KERNEL`

tijyojwad (changes requested):

cudapoa/src/cudapoa_structs.cuh
r-mafi: this convention did not seem to be applied in all cases, for example `CELLS_PER_THREAD`, therefore a couple of macros were left unchanged. Added `CUDAPOA` to all, including `CELLS_PER_THREAD`

r-mafi:
rerun tests

o8191:
rerun tests

o8191:
rerun tests

cudapoa/src/cudapoa_structs.cuh
tijyojwad: ah got it, good point. thanks for fixing the rest!

tijyojwad (approved):

PR #589, opened 2020-10-16
[cudaextender] Add and update READMEs

### LINKED ISSUE #582: [cudaextender] Update readme with cudaextender info
Add cudaextender details to the main project's readme.

atadkase:
Closes #582

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,7 +84,7 @@ A prototypical binary to showcase the use of `libcudaaligner.so` APIs.
 
 ### cudaextender
 The `cudaextender` package provides GPU-accelerated seed-extension. Details can be found in
-the package's [Readme](cudaextender/README.md).
+the package's readme.
 
 ## Clone GenomeWorks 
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -81,6 +82,10 @@ The `cudaaligner` package provides GPU-accelerated global alignment. Features in
 
 A prototypical binary to showcase the use of `libcudaaligner.so` APIs.
 
+### cudaextender
+The `cudaextender` package provides GPU-accelerated seed-extension. Details can be found in
+the package's readme.
```
atadkase, line 87: I tried linking the word Readme to cudaextender's readme file through a relative path - unfortunately Doxygen's config doesn't support that.

tijyojwad: ah that's a bummer. Would any of these work? https://stackoverflow.com/questions/19275887/how-do-i-link-between-markdown-documents-in-doxygen

atadkase: I tried all three. Only the first method (supposedly) works with both Github and Doxygen. However, the path has to be relative to Doxygen's build path (and for Github it has to be relative to the file path)

tijyojwad: Oh goodness that's a mess. Let's see if we can resolve both through some path refactoring in our cmake, but for now we can merge. Thanks @atadkase !

tijyojwad (approved):

PR #590, opened 2020-10-19
[cudamapper] Add anchor chain output functions to chainer_utils.cuh.

edawson:
Adds the necessary functions for outputting the chains of anchors that represent an overlap.


edawson:
This adds the functions needed to generate the anchor chains from a list of overlaps, assuming the overlap list has not been previously filtered (though masking using the select_mask is permissible).

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,231 @@
...
+    Overlap overlap;
+    overlap.num_residues_ = num_anchors;
+
+    overlap.query_read_id_  = start.query_read_id_;
+    overlap.target_read_id_ = start.target_read_id_;
+    assert(start.query_read_id_ == end.query_read_id_ && start.target_read_id_ == end.target_read_id_);
+
+    overlap.query_start_position_in_read_ = min(start.query_position_in_read_, end.query_position_in_read_);
+    overlap.query_end_position_in_read_   = max(start.query_position_in_read_, end.query_position_in_read_);
+    bool is_negative_strand               = end.target_position_in_read_ < start.target_position_in_read_;
```
mimaric, line 54: This can be `const` or even moved into `if` statement

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,231 @@
...
+
+namespace genomeworks
+{
+
+namespace cudamapper
+{
+namespace chainerutils
+{
+
+__host__ __device__ Overlap create_simple_overlap(const Anchor& start, const Anchor& end, const int32_t num_anchors)
```
mimaric, line 43: I would expect that there is a very similar function somewhere in overlapper? Can it be reused?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,98 @@
...
+                                              Overlap* overlaps,
+                                              double* scores,
+                                              bool* max_select_mask,
+                                              int32_t* predecessors,
+                                              const int32_t n_anchors,
+                                              const int32_t min_score);
+///
+///@brief Allocate a 1-dimensional array representing an unrolled 2D-array
+/// (overlap X n_anchors_in_overlap) of anchors within each overlap. Rather than
+/// copy the anchors, the final array holds the indices of the global index array.
```
mimaric, line 63: What's "global index array" here?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,98 @@
...
+///@param overlaps An array of Overlaps. Must have a well-formed num_residues_ field.
+///@param unrolled_anchor_chains  An array of int32_t. Will be resided on return.
+///@param num_overlaps The number of overlaps in the overlaps array.
+///@param _allocator  The DefaultDeviceAllocator
+///@param _cuda_stream The cudastream to allocate memory within.
+void allocate_anchor_chains(device_buffer<Overlap>& overlaps,
+                            device_buffer<int32_t>& unrolled_anchor_chains,
+                            device_buffer<int32_t>& anchor_chain_starts,
+                            int32_t num_overlaps,
+                            int32_t& num_total_anchors,
```
mimaric, line 74: This parameter is not documented

mimaric (changes requested):
Added a few basic comments. Can you please add more documentation?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,98 @@
...
+
+#pragma once
+
+#include <fstream>
+#include <sstream>
+#include <cstdlib>
+
+// Needed for accumulate - remove when ported to cuda
+#include <numeric>
+#include <limits>
```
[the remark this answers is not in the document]
edawson: Addressed in 5860fb1e

cudamapper/src/chainer_utils.cu
edawson: There is indeed, but I think it's reliant on the anchor chain being continuous:
https://github.com/clara-parabricks/GenomeWorks/blob/9a2e064b274ce7e6739fdc8a214fa7bd03f17ac7/cudamapper/src/overlapper_triggered.cu#L184-L233

I agree I should refactor both of these, though I think I need both because don't want to pass an array of anchors (and the original thrust-based fusion requires such an array).

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,231 @@
...
+#include <fstream>
+#include <sstream>
+#include <cstdlib>
+
+// Needed for accumulate - remove when ported to cuda
+#include <numeric>
+#include <limits>
+
+#include <cub/cub.cuh>
+#include <thrust/execution_policy.h>
```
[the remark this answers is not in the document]
edawson: No, it was not - fixed in 5860fb1e.

cudamapper/src/chainer_utils.cuh
edawson: Fixed in 2a864de1

cudamapper/src/chainer_utils.cuh
edawson: Yikes. This is just the anchors array. Should now be fixed in 2a864de1.

cudamapper/src/chainer_utils.cu
edawson: Made this `const`. I think it's a bit clearer what the meaning is if it's not used as the if condition, but I could also add a comment.

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,98 @@
...
+
+__global__ void backtrace_anchors_to_overlaps(const Anchor* anchors,
+                                              Overlap* overlaps,
+                                              double* scores,
+                                              bool* max_select_mask,
+                                              int32_t* predecessors,
+                                              const int32_t n_anchors,
+                                              const int32_t min_score);
+///
+///@brief Allocate a 1-dimensional array representing an unrolled 2D-array
```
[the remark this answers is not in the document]
edawson: Ah, got it. Corrected my shortcuts and this should now be fixed up.

cudamapper/src/chainer_utils.cu
mimaric: I agree, we can leave it as a `const` value

cudamapper/src/chainer_utils.cu
mimaric: You're right, that implementation is really strange. `overlapper_triggered.cu` will hopefully be removed once the new overlapper is done.
I'm just thinking if this should somehow be more directly connected to `Overlap` object...

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+///
+/// \param overlaps An array of Overlaps. Must have a well-formed num_residues_ field
+/// \param unrolled_anchor_chains  An array of int32_t. Will be resided on return.
+/// \param anchor_chain_starts An array holding the index in the anchors array of the first
+/// anchor in an overlap.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \param num_total_anchors The number of anchors in the anchors array.
+/// \param _allocator The DefaultDeviceAllocator for this overlapper.
+/// \param _cuda_stream The cudastream to allocate memory within.
+void allocate_anchor_chains(device_buffer<Overlap>& overlaps,
```
mimaric, line 90: IIUC `output_overlap_chains_by_backtrace()` and `output_overlap_chains_by_RLE()` need `anchor_chains` array which is allocated by `allocate_anchor_chains()`. If is necessary to this allocation outside of these functions or can this function be called by them?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+};
+
+///
+/// \brief Create an overlap from the first and last anchor in the chain and
+/// the total number of anchors in the chain.
+///
+/// \param start The first anchor in the chain.
+/// \param end The last anchor in the chain.
+/// \param num_anchors The total number of anchors in the chain.
+__host__ __device__ Overlap create_simple_overlap(const Anchor& start,
```
mimaric, line 48: This function is only called by `backtrace_anchors_to_overlaps()`. I understand that it cannot be in an anonymous namespace in order be able to call it form unittest, but it could be moved to `details` namespace in order to signal the users that they do not have to think about it

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+    else
+    {
+        overlap.relative_strand                = RelativeStrand::Forward;
+        overlap.target_start_position_in_read_ = start.target_position_in_read_;
+        overlap.target_end_position_in_read_   = end.target_position_in_read_;
+    }
+    return overlap;
+}
+
+void allocate_anchor_chains(device_buffer<Overlap>& overlaps,
```
mimaric, line 60: Please make sure that the order of functions in `.cuh` and `.cu` is the same?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+                                              Overlap* overlaps,
+                                              double* scores,
+                                              bool* max_select_mask,
+                                              int32_t* predecessors,
+                                              const int32_t n_anchors,
+                                              const int32_t min_score);
+
+///
+/// \brief Allocate a 1-dimensional array representing an unrolled 2D-array
+/// (overlap X n_anchors_in_overlap) of anchors within each overlap. Rather than
```
mimaric, line 78: >(overlap X n_anchors_in_overlap)

So if I get it right it was not a 2D-array but more like a "saw" with each overlap having a different number of anchors, right?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// copy the anchors, the final array holds the indices within the anchors array
+/// of the anchors in the chain.
+///
+/// \param overlaps An array of Overlaps. Must have a well-formed num_residues_ field
+/// \param unrolled_anchor_chains  An array of int32_t. Will be resided on return.
```
mimaric, line 83: typo: one extra space (now that I'm reading it thoroughly :) )

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \param anchor_chain_starts An array holding the index in the anchors array of the first
+/// anchor in an overlap.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \param num_total_anchors The number of anchors in the anchors array.
+/// \param _allocator The DefaultDeviceAllocator for this overlapper.
```
mimaric, line 88: Again a matter of taste, but we generally to not use underscore at before variable names. We only use it trailing for member variables in classes

cudamapper/src/chainer_utils.cuh
mimaric, line 90: `device_buffer<Overlap>& overlaps` can be `const`, right?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// (overlap X n_anchors_in_overlap) of anchors within each overlap. Rather than
+/// copy the anchors, the final array holds the indices within the anchors array
+/// of the anchors in the chain.
+///
+/// \param overlaps An array of Overlaps. Must have a well-formed num_residues_ field
+/// \param unrolled_anchor_chains  An array of int32_t. Will be resided on return.
+/// \param anchor_chain_starts An array holding the index in the anchors array of the first
+/// anchor in an overlap.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \param num_total_anchors The number of anchors in the anchors array.
```
mimaric, line 87: I don't get this one. What is "anchors array"?

mimaric, line 87: For ease of understanding I would suggest grouping output value at the beginning of function signature. We are not doing it everywhere, but we probably should

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \brief Allocate a 1-dimensional array representing an unrolled 2D-array
+/// (overlap X n_anchors_in_overlap) of anchors within each overlap. Rather than
+/// copy the anchors, the final array holds the indices within the anchors array
+/// of the anchors in the chain.
+///
+/// \param overlaps An array of Overlaps. Must have a well-formed num_residues_ field
+/// \param unrolled_anchor_chains  An array of int32_t. Will be resided on return.
+/// \param anchor_chain_starts An array holding the index in the anchors array of the first
+/// anchor in an overlap.
+/// \param num_overlaps The number of overlaps in the overlaps array.
```
mimaric, line 86: Is this parameter necessary? Is the the same as `overlaps.size()`?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \param num_total_anchors The number of anchors in the anchors array.
+/// \param _allocator The DefaultDeviceAllocator for this overlapper.
+/// \param _cuda_stream The cudastream to allocate memory within.
+void allocate_anchor_chains(device_buffer<Overlap>& overlaps,
+                            device_buffer<int32_t>& unrolled_anchor_chains,
+                            device_buffer<int32_t>& anchor_chain_starts,
+                            int32_t num_overlaps,
+                            int32_t& num_total_anchors,
+                            DefaultDeviceAllocator& _allocator,
+                            cudaStream_t& _cuda_stream);
```
mimaric, line 96: Both allocator and stream should be passed by value (allocator is just a wrapper around a `shared_ptr` which points to actual memory resource)

mimaric, line 96: It's not mandatory, but we generally use stream `0` as default value

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+                            device_buffer<int32_t>& unrolled_anchor_chains,
+                            device_buffer<int32_t>& anchor_chain_starts,
+                            int32_t num_overlaps,
+                            int32_t& num_total_anchors,
+                            DefaultDeviceAllocator& _allocator,
+                            cudaStream_t& _cuda_stream)
+{
+    // sum the number of chains across all overlaps
+    device_buffer<char> d_temp_buf(_allocator, _cuda_stream);
```
mimaric, line 69: Is it necessary to declare `d_temp_buf` here? Why not just create it with the right size when actually needed?

mimaric, line 69: We generally name variables with trailing `_d` and `_h`, not leading `d_` and `h_` (unlike CUB)

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+
+namespace genomeworks
+{
+
+namespace cudamapper
+{
+namespace chainerutils
+{
+
+struct OverlapToNumResiduesOp
```
mimaric, line 33: Can this be moved to `.cu`? Maybe even inside `allocate_anchor_chains()`? (not sure if that one will work)

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+                           temp_storage_bytes,
+                           d_residue_counts,
+                           d_num_total_anchors.data(),
+                           num_overlaps,
+                           _cuda_stream);
+
+    d_temp_buf.clear_and_resize(temp_storage_bytes);
+    d_temp_storage = d_temp_buf.data();
+
+    cub::DeviceReduce::Sum(d_temp_storage,
```
mimaric, line 87: If I understand correctly you are just doing a reduction here? Could this be replaced by `thrust::reduce`? In that case that would result in less code.
For the way to specify allocator and stream to thrust see https://github.com/clara-parabricks/GenomeWorks/blob/34d1d5231fd6385adf146809fd1c36cec4f1a841/common/base/include/claraparabricks/genomeworks/utils/cudasort.cuh#L171

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+                                  temp_storage_bytes,
+                                  d_residue_counts,
+                                  anchor_chain_starts.data(),
+                                  num_overlaps,
+                                  _cuda_stream);
+
+    d_temp_buf.clear_and_resize(temp_storage_bytes);
+    d_temp_storage = d_temp_buf.data();
+
+    cub::DeviceScan::ExclusiveSum(d_temp_storage,
```
mimaric, line 112: Again, if `thrust::exclusive_scan()` can be used if would result in less code

mimaric (changes requested):
So far reviewed `allocate_anchor_chains()` and left some general comments

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+///
+/// \brief Perform a backtrace on the predecessors array
+/// to generate overlaps from anchors. Anchors must have been chained
+/// by a chaining function that filles the predecessors and scores array.
+///
+/// \param anchors An array of anchors.
+/// \param overlaps An array of overlaps to be filled.
+/// \param scores An array of scores. Only chains with a score greater than
+/// min_score will be backtraced.
+/// \param max_select_mask A boolean mask, used to mask out any subchains during
```
mimaric, line 61: This should probably be documented a bit better, I cannot understand it without looking into implementation

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// the chain_starts array.
+/// \param anchor_chains An array (allocated by allocate_anchor_chains) which will hold
+/// the indices of anchors within each chain.
+/// \param anchor_chain_starts An array which holds the indices of the first anchor for
+/// each overlap in the overlaps array.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \return __global__
+__global__ void output_overlap_chains_by_RLE(const Overlap* overlaps,
+                                             const Anchor* anchors,
+                                             const int32_t* chain_starts,
```
mimaric, line 140: How confident are we that we won't have more than 2^31 anchors? An example I usually use regularly has 300M anchors

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+__global__ void backtrace_anchors_to_overlaps(const Anchor* anchors,
+                                              Overlap* overlaps,
+                                              double* scores,
+                                              bool* max_select_mask,
+                                              int32_t* predecessors,
+                                              const int32_t n_anchors,
+                                              const int32_t min_score)
+{
+    const std::size_t d_tid = blockIdx.x * blockDim.x + threadIdx.x;
+    if (d_tid < n_anchors)
```
mimaric, line 185: Canonical way for doing this is
```
If (d_tid > n_acnhors)
{
  return;
}
```
but this is fine as well

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+    {
+
+        int32_t global_overlap_index = d_tid;
```
mimaric, line 188: `const` (and in many other places)

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+        if (scores[d_tid] >= min_score)
```
mimaric, line 189: Again, assigning one thread to each overlap likely won't result in good performance

mimaric (changes requested):

cudamapper/src/chainer_utils.cuh
edawson: I could write a wrapper, but the two `output_overlap_chains_*` functions are kernels, so I can't call the allocation from within them (I don't think).

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \brief Perform a backtrace on the predecessors array
+/// to generate overlaps from anchors. Anchors must have been chained
+/// by a chaining function that filles the predecessors and scores array.
+///
+/// \param anchors An array of anchors.
+/// \param overlaps An array of overlaps to be filled.
+/// \param scores An array of scores. Only chains with a score greater than
+/// min_score will be backtraced.
+/// \param max_select_mask A boolean mask, used to mask out any subchains during
+/// the backtrace.
```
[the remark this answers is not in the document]
edawson: Makes sense. I had to reinstall the doxygen comment generator I used and the defaults are way off apparently. I think it should be mostly fixed now.

cudamapper/src/chainer_utils.cuh
edawson: I think creating an overlap from two anchors and a number of residues is useful and could be a user-facing API function, but the function could be better named.

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+
+__global__ void output_overlap_chains_by_RLE(const Overlap* overlaps,
+                                             const Anchor* anchors,
+                                             const int32_t* chain_starts,
+                                             const int32_t* chain_lengths,
+                                             int32_t* anchor_chains,
+                                             int32_t* anchor_chain_starts,
+                                             int32_t num_overlaps)
+{
+    int32_t d_thread_id = blockIdx.x * blockDim.x + threadIdx.x;
```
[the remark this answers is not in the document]
edawson: Oh - that makes sense. Is convention to avoid it or it's just unnecessary?

cudamapper/src/chainer_utils.cuh
edawson: That's correct, it is technically an array of ragged arrays.

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+                            DefaultDeviceAllocator& _allocator,
+                            cudaStream_t& _cuda_stream);
+
+///
+/// \brief Calculate the anchors chains used to produce each overlap in the
+/// overlap array for anchors chained by backtrace_anchors_to_overlaps.
+///
+/// \param overlaps An array of overlaps.
+/// \param anchors The array of anchors used to generate overlaps.
+/// \param select_mask An array of bools, used to mask overlaps from output.
```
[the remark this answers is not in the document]
edawson: `output_overlap_chains_by_RLE()` does not use a select_mask; it is used by `output_overlap_chains_by_backtrace()` to mark anchors at the end of chains.

cudamapper/src/chainer_utils.cu
edawson: I definitely could - I had copy-pasted this pattern for declaring CUB function temp memory when I first started and never questioned it.

cudamapper/src/chainer_utils.cu
edawson: Do thrust/CUB now have the same performance? The CUB docs seem to indicate better performance than thrust. A quick Google search seems to provide little evidence either way.

cudamapper/src/chainer_utils.cuh
edawson: Not terribly confident actually. I've refactored it to an unsigned int.

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,139 @@
...
+
+    int32_t num_total_anchors;
+    chainerutils::allocate_anchor_chains(d_overlaps,
+                                         unrolled_anchor_chains,
+                                         chain_starts,
+                                         overlaps.size(),
+                                         num_total_anchors,
+                                         allocator,
+                                         cu_ptr);
+    // void allocate_anchor_chains(device_buffer<Overlap> overlaps,
```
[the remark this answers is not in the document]
ahehn-nv: resolving for @mimaric

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+
+struct OverlapToNumResiduesOp
+{
+    __device__ __forceinline__ int32_t operator()(const Overlap& overlap) const
+    {
+        return overlap.num_residues_;
+    }
+};
+
+///
```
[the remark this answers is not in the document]
ahehn-nv: resolving for @mimaric

cudamapper/src/chainer_utils.cuh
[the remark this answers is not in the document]
ahehn-nv: resolving for @mimaric

cudamapper/src/chainer_utils.cuh
edawson: Yes it can - moved.

cudamapper/src/chainer_utils.cuh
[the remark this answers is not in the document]
edawson: Fixed in 756739f.

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \param chain_starts An array which holds the indices of the first anchor for
+/// each overlap in the overlaps array.
+/// \param chain_lengths An array which holds the length of each run of anchors, corresponding to
+/// the chain_starts array.
+/// \param anchor_chains An array (allocated by allocate_anchor_chains) which will hold
+/// the indices of anchors within each chain.
+/// \param anchor_chain_starts An array which holds the indices of the first anchor for
+/// each overlap in the overlaps array.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \return __global__
```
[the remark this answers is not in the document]
edawson: Yep, another generator issue. Not sure how to fix the config on this one, but I've manually fixed this at the moment.

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+    int32_t stride      = blockDim.x * gridDim.x;
```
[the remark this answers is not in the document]
edawson: Fixed in 756739f74.

cudamapper/src/chainer_utils.cuh
mimaric: You're right, sorry about that

cudamapper/src/chainer_utils.cuh
mimaric: Ok, I would suggest reformulating the documentation then. To me 2D-array sounds like a matrix, i.e. all arrays in second dimension have the same length

cudamapper/src/chainer_utils.cu
mimaric: It depends. Thrust is largely built on top on CUB so in general Thrust should not be faster than CUB (not sure if there are exception to this rule).
Going in the opposite direction, CUB is more fine-grained so it's sometimes possible to get better performance using some problem-specific properties. For example in `sort_by_two_keys()` after first sort we only move the indices instead of the whole array and we use CUB's API which allows to specify the number of bytes that have to be sorted as we know that not all keys are possible: https://github.com/clara-parabricks/GenomeWorks/blob/bc855b37a62620a8f32388bffc426a334dbbaded/common/base/include/claraparabricks/genomeworks/utils/cudasort.cuh#L139
As a rule of thumb I would suggest using Thrust as it is more readable and only switching to CUB if there are such problem-specific properties or if profiling shows that that is a bottleneck so it would be worth giving CUB a try

cudamapper/src/chainer_utils.cu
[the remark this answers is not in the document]
mimaric: I would say it's our convention

cudamapper/src/chainer_utils.cuh
mimaric: I was going for 32/64-bit discussion and not 31/32

cudamapper/src/chainer_utils.cuh
mimaric: I agree. That's why I was suggesting moving it closed to `Overlap`'s definition: https://github.com/clara-parabricks/GenomeWorks/pull/590#discussion_r510250360

edawson: I see. I've renamed it to `create_overlap` in c72e5c96.

cudamapper/src/chainer_utils.cu
edawson: I've renamed all of these in 9f75411a491f79a6a5dda56eb4cf68a95910042b.

edawson: Changed in 6df3b764.

cudamapper/src/chainer_utils.cuh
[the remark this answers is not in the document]
edawson: All of these are fixed in 39f7232b73.

cudamapper/src/chainer_utils.cu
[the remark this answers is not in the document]
edawson: This has been fixed.

cudamapper/src/chainer_utils.cuh
edawson: That does make sense - raised to 64 bits in 781c2fd2.

cudamapper/src/chainer_utils.cu
edawson: Resolved by 78da54195e1d117af4d7399b7fd222304a6b9316.

cudamapper/src/chainer_utils.cu
edawson: Following up on this after some offline discussion - the backtrace is not currently the main bottleneck, but I'll explore two possibilities for how to improve coalescence here (and elsewhere). One is to launch one warp per backtrace; the other is to use int4 to better pack the reads.

cudamapper/src/chainer_utils.cuh
edawson: Note - I couldn't get it into allocate_anchor_chains(). It now resides in the `.cu` file.

cudamapper/src/chainer_utils.cu
edawson: I have implemented a version that now uses one active thread within a single warp per overlap.

cudamapper/src/chainer_utils.cuh
edawson: I made another attempt at this - I hope it's now clearer.

cudamapper/src/chainer_utils.cu
edawson: Swapped for thrust::exclusive_scan in  6244636

cudamapper/src/chainer_utils.cu
edawson: Swapped to thrust in 6244636 .

cudamapper/src/chainer_utils.cu
edawson: I've reverted the single-warp-per-overlap to a single-thread-per-overlap version based on offline discussions.

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+    for (int i = d_thread_id; i < num_overlaps; i += stride)
```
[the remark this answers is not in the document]
edawson: After offline discussions about this, I think the consensus is it should be improved (or later removed) but that it is lower priority given that these functions are not major bottlenecks yet.

cudamapper/src/chainer_utils.cuh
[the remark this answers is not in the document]
mimaric: ok, makes sense

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,221 @@
...
+                                                   int32_t* anchor_chain_starts,
+                                                   int32_t num_overlaps,
+                                                   bool check_mask)
+{
+    int32_t d_thread_id = blockIdx.x * blockDim.x + threadIdx.x;
+    int32_t stride      = blockDim.x * gridDim.x;
+
+    // Processes one overlap per iteration,
+    // "i" corresponds to an overlap
+    for (int i = d_thread_id; i < num_overlaps; i += stride)
```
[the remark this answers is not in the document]
mimaric: As discussed in that `output_overlap_chains_by_RLE()` we're leaving it as it is for now

cudamapper/src/chainer_utils.cu
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: done

mimaric: done

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// will have their anchor chains calculated.
+/// \return __global__
+__global__ void output_overlap_chains_by_backtrace(const Overlap* overlaps,
+                                                   const Anchor* anchors,
+                                                   const bool* select_mask,
+                                                   const int32_t* predecessors,
+                                                   int32_t* anchor_chains,
+                                                   int32_t* anchor_chain_starts,
+                                                   int32_t num_overlaps,
+                                                   bool check_mask);
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+/// \param end The last anchor in the chain.
+/// \param num_anchors The total number of anchors in the chain.
+__host__ __device__ Overlap create_simple_overlap(const Anchor& start,
+                                                  const Anchor& end,
+                                                  const int32_t num_anchors);
+
+///
+/// \brief Perform a backtrace on the predecessors array
+/// to generate overlaps from anchors. Anchors must have been chained
+/// by a chaining function that filles the predecessors and scores array.
```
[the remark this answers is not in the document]
mimaric: done

cudamapper/src/chainer_utils.cuh
mimaric: I still find it a bit hard to understand what is a subchain here

cudaaligner/samples/sample_cudaaligner.cpp
```diff
@@ -21,6 +21,7 @@
 #include <claraparabricks/genomeworks/utils/genomeutils.hpp>
 
 #include <cuda_runtime_api.h>
+#include <getopt.h>
```
mimaric, line 24: Why is this needed?

cudamapper/samples/sample_cudamapper.cpp
```diff
@@ -29,6 +29,7 @@
 #include <iostream>
 #include <string>
 #include <mutex>
+#include <unistd.h>
```
mimaric, line 32: Why is this needed?

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+* Unless required by applicable law or agreed to in writing, software
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "chainer_utils.cuh"
+
+#include <cub/cub.cuh>
```
mimaric, line 19: CUB is not needed anymore

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+
+namespace genomeworks
+{
+
+namespace cudamapper
+{
+namespace chainerutils
+{
+
+struct OverlapToNumResiduesOp
```
mimaric, line 36: Is this still needed?

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+    {
+        overlap.relative_strand                = RelativeStrand::Forward;
+        overlap.target_start_position_in_read_ = start.target_position_in_read_;
+        overlap.target_end_position_in_read_   = end.target_position_in_read_;
+    }
+    return overlap;
+}
+
+__global__ void backtrace_anchors_to_overlaps(const Anchor* const anchors,
+                                              Overlap* overlaps,
```
mimaric, line 80: This can be `Overlap* const overlaps`

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+                                              double* const scores,
```
mimaric, line 81: This can be `const double* const scores` (please check all other such places in PR)

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+                                              bool* const max_select_mask,
+                                              int32_t* const predecessors,
+                                              const int64_t n_anchors,
+                                              const int32_t min_score)
```
mimaric, line 85: Optional: `scores` is `double`. Would it make sense to have `min_score` also be double?

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,199 @@
...
+    auto thrust_exec_policy = thrust::cuda::par(allocator).on(cuda_stream);
+    num_total_anchors       = thrust::reduce(thrust_exec_policy,
+                                       thrust::make_transform_iterator(overlaps.begin(), ConvertOverlapToNumResidues()),
+                                       thrust::make_transform_iterator(overlaps.end(), ConvertOverlapToNumResidues()),
+                                       0);
+
+    unrolled_anchor_chains.clear_and_resize(num_total_anchors);
+    anchor_chain_starts.clear_and_resize(overlaps.size());
+
+    thrust::exclusive_scan(thrust_exec_policy,
```
mimaric, line 136: Can this be simplified using `thrust::transform_exclusive_scan`?
https://thrust.github.io/doc/group__transformed__prefixsums_gaf67d66dbbff500349ff56671b9ee9c4b.html#gaf67d66dbbff500349ff56671b9ee9c4b

mimaric (changes requested):

cudamapper/src/chainer_utils.cuh
edawson: Attempted to reword "subchain" in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudaaligner/samples/sample_cudaaligner.cpp
edawson: Removed in 405ec7e4e0b090a3323da5970c038a219078c5bc. This was a compile error for me after a merge, but it looks like it was then included twice.

cudamapper/samples/sample_cudamapper.cpp
edawson: I'm actually not sure - removed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cu
edawson: Removed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cu
edawson: Nope - removed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cu
edawson: Fixed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cu
edawson: Fixed in 405ec7e4e0b090a3323da5970c038a219078c5bc, and I went through the PR. All pointers passed in are now `* const`.

cudamapper/src/chainer_utils.cu
edawson: Yes - fixed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cu
edawson: Yes, as can the reduce above that. Fixed in 405ec7e4e0b090a3323da5970c038a219078c5bc.

cudamapper/src/chainer_utils.cuh
mimaric: Makes sense now!

cudamapper/src/overlapper_triggered.cu
```diff
@@ -24,11 +24,6 @@
 
 #include <claraparabricks/genomeworks/utils/cudautils.hpp>
 
-#ifndef NDEBUG // only needed to check if input is sorted in assert
```
mimaric, line 27: Does this still compile in debug mode? I think lines 255/260 require these headers

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,134 @@
...
+* distributed under the License is distributed on an "AS IS" BASIS,
+* WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+* See the License for the specific language governing permissions and
+* limitations under the License.
+*/
+
+#include "gtest/gtest.h"
+
+#include <thrust/device_vector.h>
+#include <thrust/host_vector.h>
```
mimaric, line 20: These do not seem to be used anywhere

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,134 @@
...
+    Anchor c;
+    c.query_read_id_           = 12;
+    c.query_position_in_read_  = 15000;
+    c.target_read_id_          = 46;
+    c.target_position_in_read_ = 16000;
+
+    Overlap p_bc = chainerutils::create_overlap(b, c, 22);
+    ASSERT_EQ(p_bc.target_start_position_in_read_, 16000);
+    ASSERT_EQ(p_bc.target_end_position_in_read_, 20400);
+    ASSERT_EQ(static_cast<char>(p_bc.relative_strand), static_cast<char>(RelativeStrand::Reverse));
```
mimaric, line 67: These two cannot be compared directly (might be something about `enum class` I'm not aware of)?

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,134 @@
...
+    c.query_position_in_read_  = 15000;
+    c.target_read_id_          = 46;
+    c.target_position_in_read_ = 16000;
+
+    Overlap p_bc = chainerutils::create_overlap(b, c, 2);
+
+    std::vector<Anchor> anchors;
+    anchors.push_back(a);
+    anchors.push_back(b);
+    anchors.push_back(c);
```
mimaric, line 102: As you're using `std::vector` you can simple do `std::vector<Anchor> anchors = {a, b, c};`

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,134 @@
...
+    ASSERT_EQ(p_bc.target_end_position_in_read_, 20400);
+    ASSERT_EQ(static_cast<char>(p_bc.relative_strand), static_cast<char>(RelativeStrand::Reverse));
+}
+
+TEST(TestChainerUtils, Anchor_Chain_Allocation_Tests)
+{
+    DefaultDeviceAllocator allocator = create_default_device_allocator();
+
+    CudaStream cuda_stream = make_cuda_stream();
+    auto cu_ptr            = cuda_stream.get();
```
mimaric, line 75: This stream could be a given a more verbose name

mimaric (changes requested):
Please have a look at still unresolved comments from previous reviews as well (you can ignore comment about `create_overlap`)

cudamapper/src/chainer_utils.cuh
edawson: This is not necessarily the same as `overlaps.size()`; I've been bitten by this multiple times at this point. Calling `cub::Flagged` does not create an output array guaranteed to be the exact size (only equal to or greater than the number of flagged elements). If there's a `trim` function for the device_buffer class I could call that.

cudamapper/tests/Test_CudamapperChainerUtils.cu
edawson: I'm certain this was the case at some point, but now it seems to work fine - I've updated to a new GCC but I don't think that would make it different. We'll see if the tests accept it.

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,149 @@
...
+                                                   const bool* select_mask,
+                                                   const int32_t* predecessors,
+                                                   int32_t* anchor_chains,
+                                                   int32_t* anchor_chain_starts,
+                                                   int32_t num_overlaps,
+                                                   bool check_mask);
+///
+/// \brief Calculate the anchors chains used to produce each overlap in the
+/// overlap array for anchors chained by RLE.
+///
```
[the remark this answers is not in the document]
edawson: Fixed (by 530adfd65e913d03fe7d6e1709786d91a9b4b608)

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,201 @@
...
+                                                 overlaps.begin(),
+                                                 overlaps.end(),
+                                                 ConvertOverlapToNumResidues(),
+                                                 0,
+                                                 sum_op);
+
+    unrolled_anchor_chains.clear_and_resize(num_total_anchors);
+    anchor_chain_starts.clear_and_resize(overlaps.size());
+
+    // thrust::exclusive_scan(thrust_exec_policy,
```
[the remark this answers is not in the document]
edawson: Fixed

cudamapper/src/overlapper_triggered.cu
edawson: Fixed

cudamapper/tests/Test_CudamapperChainerUtils.cu
edawson: Yep, removed.

cudamapper/src/chainer_utils.cu
```diff
@@ -0,0 +1,201 @@
...
+    return overlap;
+}
+
+__global__ void backtrace_anchors_to_overlaps(const Anchor* const anchors,
+                                              Overlap* const overlaps,
+                                              const double* const scores,
+                                              bool* const max_select_mask,
+                                              const int32_t* const predecessors,
+                                              const int64_t n_anchors,
+                                              const double min_score)
```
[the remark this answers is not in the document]
edawson: I've made this a float - I'll have to refactor all the other doubles on other branches after this gets merged.

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,134 @@
+/*
```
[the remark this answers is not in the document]
edawson: 530adfd65e913d03fe7d6e1709786d91a9b4b608 adds tests for all of these.

cudamapper/tests/Test_CudamapperChainerUtils.cu
edawson: Fixed in 530adfd65e913d03fe7d6e1709786d91a9b4b608

cudamapper/tests/Test_CudamapperChainerUtils.cu
edawson: I've given this a more verbose name in 530adfd65e913d03fe7d6e1709786d91a9b4b608.

mimaric:
rerun tests

cudamapper/src/chainer_utils.cuh
mimaric: `num_overlaps ` does not exist anymore

cudamapper/tests/Test_CudamapperChainerUtils.cu
```diff
@@ -0,0 +1,370 @@
...
+    std::vector<Overlap> overlaps;
+    overlaps.resize(overlaps_d.size());
+    cudautils::device_copy_n(overlaps_d.data(), overlaps_d.size(), overlaps.data(), backtrace_cu_stream_ptr);
+    std::vector<int32_t> terminals_h;
+    terminals_h.resize(overlaps.size());
+    cudautils::device_copy_n(overlap_terminal_anchors.data(), overlap_terminal_anchors.size(), terminals_h.data(), backtrace_cu_stream_ptr);
+    ASSERT_EQ(overlaps_d.size(), 3);
+
+    ASSERT_EQ(overlaps[0].num_residues_, 1);
+    ASSERT_EQ(overlaps[0].target_read_id_, UINT32_MAX);
```
mimaric, line 127: Better use `std::numeric_limits<decltype(overlaps[0].target_read_id_)>::max()`
https://en.cppreference.com/w/cpp/types/numeric_limits

cudamapper/src/overlapper_triggered.cu
```diff
@@ -24,11 +24,11 @@
...
 #include <algorithm>
+
+#ifndef NDEBUG
```
mimaric, line 29: Why is `algorithm` needed in release mode as well?

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,129 @@
...
+/// \param overlaps An array of overlaps.
+/// \param anchors The array of anchors used to generate overlaps.
+/// \param chain_starts An array which holds the indices of the first anchor for each overlap in the overlaps array.
+/// \param chain_lengths An array which holds the length of each run of anchors, corresponding to the chain_starts array.
+/// \param anchor_chains An array (allocated by allocate_anchor_chains) which will hold the indices of anchors within each chain.
+/// \param anchor_chain_starts An array which holds the indices of the first anchor for each overlap in the overlaps array.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// Launch configuration: any number of blocks/threads, no dynamic shared memory, cuda_stream must be the same in which anchors/overlaps/scores/mask/predecessors/chains/starts were allocated.
+/// example: backtrace_anchors_to_overlaps<<<2048, 32, 0, cuda_stream>>>(...)
```
mimaric, line 117: Further descriptions should be placed between `\brief` and `\param`

cudamapper/src/chainer_utils.cuh
```diff
@@ -0,0 +1,129 @@
...
+/// \param unrolled_anchor_chains An array of int32_t. Will be resized on return.
+/// \param anchor_chain_starts An array holding the index of the first anchor in an overlap within the anchors array used during chaining.
+/// \param num_overlaps The number of overlaps in the overlaps array.
+/// \param num_total_anchors The total number of anchors across all overlaps.
+/// \param allocator The DefaultDeviceAllocator for this overlapper.
+/// \param cuda_stream The cudastream to allocate memory within.
+void allocate_anchor_chains(const device_buffer<Overlap>& overlaps,
+                            device_buffer<int32_t>& unrolled_anchor_chains,
+                            device_buffer<int32_t>& anchor_chain_starts,
+                            int64_t& num_total_anchors,
```
mimaric, line 78: Is this necessary? I see it's used to set the size of `unrolled_anchor_chains`

mimaric (changes requested):

cudamapper/src/chainer_utils.cuh
mimaric: @edawson What do you think about this?

mimaric: @ahehn-nv as well

ahehn-nv: sounds like a good improvement to me. Since output arguments are less likely to be optional, the beginning makes sense (more than the end).

PR #592, opened 2020-10-26
[logging] fix spdlog incompatibility with gcc9

tijyojwad:
1. WAR fix for spdlog with gcc9. disable logging for gcc9
2. enable SAM support by default and fix cmake compile option for SAM

### DIFF
--- a/common/base/include/claraparabricks/genomeworks/logging/logging.hpp
+++ b/common/base/include/claraparabricks/genomeworks/logging/logging.hpp
@@ -94,7 +94,11 @@
 #ifdef GW_CUDA_BEFORE_9_2
 // Due to a header file incompatibility with nvcc in CUDA 9.0
 // logging through the logger class in GW is disabled for any .cu files.
+#elif __GNUC__  >= 9
 #pragma message("Logging disabled for CUDA Toolkit < 9.2")
+// Due to a ISO C++ standard incompatibility the spdlog fails to pass
+// pedantic requirements.
+#pragma message("Logging disabled for GCC >= 9")
 #else
 #include <spdlog/spdlog.h>
 #endif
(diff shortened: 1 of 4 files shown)


nvvishanthi:
@tijyojwad since  the definition of `GW_BUILD_HTSLIB` is done per-target in this PR, and right now it's only done for libcudamapper, I believe you would have to add it to the main `cudamapper-bin` target as well as `sample_cudamapper` target.

It should be easy, adding the following to cudamapper/CMakeLists.txt:
`target_compile_definitions(${MODULE_NAME}-bin PUBLIC GW_BUILD_HTSLIB)`

and adding the following to cudamapper/samples/CMakeLists.txt:
`target_compile_definitions(${MODULE_NAME} PUBLIC GW_BUILD_HTSLIB)`

common/base/include/claraparabricks/genomeworks/logging/logging.hpp
```diff
@@ -94,7 +94,11 @@
 #ifdef GW_CUDA_BEFORE_9_2
 // Due to a header file incompatibility with nvcc in CUDA 9.0
 // logging through the logger class in GW is disabled for any .cu files.
+#elif __GNUC__ >= 9
 #pragma message("Logging disabled for CUDA Toolkit < 9.2")
```
r-mafi, line 98: shouldn't this line should be moved above elif?

nvvishanthi:
Actually, the above may not be true, since PUBLIC implies that the compile definitions are exposed when linking against it. So you can ignore the above comment, sorry. LGTM

nvvishanthi (approved):

common/base/include/claraparabricks/genomeworks/logging/logging.hpp
tijyojwad: Totally right, it should be :). Just fixed.

tijyojwad:
Thanks @nvvishanthi ! Made those changes to the PR now.

r-mafi (approved):

PR #595, opened 2020-11-05
[cudamapper] add important gcc flags

tijyojwad:
Update missing gcc flag in cudamapper cmake and fix error resulting from missing flag

### DIFF
--- a/cudamapper/CMakeLists.txt
+++ b/cudamapper/CMakeLists.txt
@@ -76,7 +76,11 @@ if (TARGET htslib_project)
 else()
      target_link_libraries(${MODULE_NAME} gwbase gwio cub)
 endif()
-target_compile_options(${MODULE_NAME} PRIVATE -Werror)
+
+target_compile_options(${MODULE_NAME} PRIVATE -Werror -Wall -Wextra)
+if (gw_optimize_for_native_cpu)
+    target_compile_options(cudapoa PRIVATE -march=native)
+endif()
 
 add_doxygen_source_dir(${CMAKE_CURRENT_SOURCE_DIR}/include)
 
--- a/cudamapper/CMakeLists.txt
+++ b/cudamapper/CMakeLists.txt
@@ -79,7 +79,7 @@ endif()
 
 target_compile_options(${MODULE_NAME} PRIVATE -Werror -Wall -Wextra)
 if (gw_optimize_for_native_cpu)
-    target_compile_options(cudapoa PRIVATE -march=native)
+    target_compile_options(${MODULE_NAME} PRIVATE -march=native)
 endif()
 
 add_doxygen_source_dir(${CMAKE_CURRENT_SOURCE_DIR}/include)
--- a/cudamapper/src/cudamapper_utils.cpp
+++ b/cudamapper/src/cudamapper_utils.cpp
@@ -38,7 +38,7 @@ std::vector<gw_string_view_t> split_into_kmers(const gw_string_view_t& s, const 
     const std::size_t kmer_count = s.length() - kmer_size + 1;
     std::vector<gw_string_view_t> kmers;
 
-    if (s.length() < kmer_size)
+    if (get_size(s) < kmer_size)
     {
         kmers.push_back(s);
         return kmers;
--- a/cudamapper/src/overlapper.cpp
+++ b/cudamapper/src/overlapper.cpp
@@ -136,10 +136,10 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps, const boo
 {
     const auto num_overlaps = get_size(overlaps);
     bool in_fuse            = false;
-    int fused_target_start  = 0;
-    int fused_query_start   = 0;
-    int fused_target_end    = 0;
-    int fused_query_end     = 0;
+    position_in_read_t fused_target_start  = UINT32_MAX;
+    position_in_read_t fused_query_start   = UINT32_MAX;
+    position_in_read_t fused_target_end    = UINT32_MAX;
+    position_in_read_t fused_query_end     = UINT32_MAX;
     int num_residues        = 0;
     Overlap prev_overlap;
     std::vector<bool> drop_overlap_mask;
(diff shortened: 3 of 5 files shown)


cudamapper/CMakeLists.txt
```diff
@@ -76,7 +76,11 @@ if (TARGET htslib_project)
 else()
      target_link_libraries(${MODULE_NAME} gwbase gwio cub)
 endif()
-target_compile_options(${MODULE_NAME} PRIVATE -Werror)
+
+target_compile_options(${MODULE_NAME} PRIVATE -Werror -Wall -Wextra)
+if (gw_optimize_for_native_cpu)
+    target_compile_options(cudapoa PRIVATE -march=native)
```
nvvishanthi, line 82: Out of curiosity, why is the -march flag set for the cudapoa target here? Is it supposed to be `${MODULE_NAME}` or was this on purpose?

cudamapper/src/cudamapper_utils.cpp
```diff
@@ -38,7 +38,7 @@ std::vector<gw_string_view_t> split_into_kmers(const gw_string_view_t& s, const
     const std::size_t kmer_count = s.length() - kmer_size + 1;
     std::vector<gw_string_view_t> kmers;
 
-    if (s.length() < kmer_size)
+    if (get_size(s) < kmer_size)
```
nvvishanthi, line 41: Copying over from previous PR: @mimaric fixed to use signed_integer_util functions

cudamapper/src/utils.cpp
```diff
@@ -156,13 +156,13 @@ static inline void encode_seq(bam1_t* const alignment, const std::string& seq)
             15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15,
             15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15, 15};
 
-    const size_t len = seq.length();
+    const int32_t len = get_size(seq);
```
nvvishanthi, line 159: Copying over from previous PR: @mimaric fixed to use signed_integer_util functions

cudamapper/src/overlapper.cpp
```diff
@@ -350,7 +351,7 @@ void Overlapper::rescue_overlap_ends(std::vector<Overlap>& overlaps,
 
         while (rescue_rounds < max_rescue_rounds)
         {
-            details::overlapper::extend_overlap_by_sequence_similarity(overlap, query_view, target_view, 100, 0.9);
+            details::overlapper::extend_overlap_by_sequence_similarity(overlap, query_view, target_view, extension, required_similarity);
```
nvvishanthi, line 354: @edawson is this the change you were referring to?

cudamapper/src/overlapper.cpp
```diff
@@ -136,11 +136,11 @@ void Overlapper::post_process_overlaps(std::vector<Overlap>& overlaps, const boo
...
     bool in_fuse            = false;
-    int fused_target_start;
-    int fused_query_start;
-    int fused_target_end;
-    int fused_query_end;
-    int num_residues = 0;
+    int fused_target_start  = 0;
+    int fused_query_start   = 0;
+    int fused_target_end    = 0;
+    int fused_query_end     = 0;
```
nvvishanthi, line 142: Also @edawson is there a latent bug here? Compiler was throwing a uninitialized variable error here, which could happen in some cases. Is initializing to 0 a reasonable change or do we need to set the variable previously like is done on line 168?

cudamapper/CMakeLists.txt
tijyojwad: Oh yes definitely needs to be `MODULE_NAME`. good catch!

cudamapper/src/utils.cpp
```diff
@@ -305,7 +304,11 @@ void print_sam(const std::vector<Overlap>& overlaps,
         // TODO: write AUX data if available
 
         std::lock_guard<std::mutex> lg(write_output_mutex);
-        int out = sam_write1(file.get(), header.get(), alignment.get());
+        result = sam_write1(file.get(), header.get(), alignment.get());
+        if (result < 0)
+        {
+            fprintf(stderr, "print_sam: could not write alignment");
```
mimaric, line 310: Should we do something more here? Like throw an exception or similar

mimaric, line 310: We use `std::cerr` instead of `fprintf`. It's not ideal, but we should still try to be consistent

nvvishanthi: That might be a good idea. `sam_write1()` has a couple of failure modes because internally it does some thread creation, (potentially) memory allocations, and then some io. Should there be different exceptions for each or just a generic runtime_error?

nvvishanthi: Fixed everywhere in latest version

cudamapper/CMakeLists.txt
nvvishanthi: Fixed in latest version

cudamapper/src/utils.cpp
mimaric: I think a generic exception would suffice, one can then investigate further if needed. I just think it's important to signal the failure to the user in a way better than just a print

nvvishanthi: 👍 fixed in latest commit

cudamapper/src/overlapper.cpp
edawson: Yes, that's the one I wanted to make.

cudamapper/src/overlapper.cpp
edawson: I think this should be fine - it doesn't look like these get used or set elsewhere (besides line 168). It might be good to set them to INT_MAX since zero is a valid value, but otherwise I think this should work.

o8191:
rerun tests

o8191:
rerun tests

cudamapper/src/overlapper.cpp
ahehn-nv: Optional: I would prefer `std::numeric_limits<position_in_read_t>::max()` over the MAX macros, that way it will still work in case we ever change the underlying type. (You need to add `#include <limits>`.)

nvvishanthi: That's a good point. Changed to use numeric_limits in the latest version

PR #598, opened 2020-11-12
[cudapoa] Update cudapoa to use DeviceAllocator

atadkase:
- Adds DeviceAllocator as default allocator for caching device allocations.
- Updated size limit to be a int64_t everywhere.

### DIFF
--- a/common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
+++ b/common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
@@ -64,8 +64,8 @@ public:
     /// Allocates the buffer
     /// \param buffer_size
     DevicePreallocatedAllocator(size_t buffer_size)
-    : buffer_size_(roundup_allocation(buffer_size))
-    , buffer_ptr_(create_buffer(buffer_size_))
+        : buffer_size_(roundup_allocation(buffer_size))
+        , buffer_ptr_(create_buffer(buffer_size_))
     {
         assert(buffer_size_ > 0);
         MemoryBlock whole_memory_block;
--- a/common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
+++ b/common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
@@ -64,13 +64,13 @@ public:
     /// Allocates the buffer
     /// \param buffer_size
     DevicePreallocatedAllocator(size_t buffer_size)
-        : buffer_size_(buffer_size)
-        , buffer_ptr_(create_buffer(buffer_size))
+    : buffer_size_(roundup_allocation(buffer_size))
+    , buffer_ptr_(create_buffer(buffer_size_))
     {
-        assert(buffer_size > 0);
+        assert(buffer_size_ > 0);
         MemoryBlock whole_memory_block;
         whole_memory_block.begin = 0;
-        whole_memory_block.size  = buffer_size;
+        whole_memory_block.size  = buffer_size_;
         free_blocks_.push_back(whole_memory_block);
     }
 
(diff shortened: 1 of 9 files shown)


cudapoa/benchmarks/multi_batch.hpp
```diff
@@ -51,11 +51,12 @@ class MultiBatch
         size_t total = 0, free = 0;
         cudaSetDevice(0);
         cudaMemGetInfo(&free, &total);
-        size_t mem_per_batch = 0.9 * free / num_batches_;
+        DefaultDeviceAllocator allocator = create_default_device_allocator(0.901*free);
```
atadkase, line 54: WAR for the bug that DeviceAllocator cannot allocate everything it owns.

r-mafi:
@atadkase, looks like 4 out of 29 tests fail. Have you checked if they run locally on your machine? I reviewed and the changes look good.

atadkase:
rerun tests

common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
```diff
@@ -64,13 +64,13 @@ public:
     /// Allocates the buffer
     /// \param buffer_size
     DevicePreallocatedAllocator(size_t buffer_size)
-        : buffer_size_(buffer_size)
-        , buffer_ptr_(create_buffer(buffer_size))
+        : buffer_size_(roundup_allocation(buffer_size))
```
atadkase, line 67: Device allocator fix for not being able to allocate entire block.

r-mafi: a couple of points: 
1- I was going to suggest printing a message in cases where allocated spaces is different and slightly larger than requested space, but since rounding up is only making the allocated space multiple of 256, the different could be extremely small compared to typical several gigs of allocator space request, therefore a message it is not really needed.
2- every call to `allocate_memory_block()` will round up allocated block to 256. Your fix should resolve most of the cases, but here is a hypothetical scenario where this fix won't be sufficient:
consider total space requested is 1020B and it will be split into 4 subsequent blocks of 255. Each of 255 requests are rounded to 256, the total required space would be 1024, which using `roundup()` at the initial call, 1020 is changed to 1024 and there won't be a problem. But if 1020 is split into 5 blocks of 204, then total required space would be 5*256 which is larger than initial rounded-up allocated 1024B.
Is this example displays a real limitation? I'd say no. To create this example, I had to choose unrealistically small request value for device allocator and split it to large number of blocks. In practice where allocated block is multiple GB, the number of split requests should be several millions to cause such a problem which is an unlikely scenario. So in summary, I think your fix is a good practical solution.

r-mafi (approved):

common/base/include/claraparabricks/genomeworks/utils/device_preallocated_allocator.cuh
ahehn-nv: Adding @mimaric who wrote the DevicePreallocatedAllocator originally.

mimaric: Yes, this is an edge case that is not covered by the original design.
The reason allocations are aligned to 256B is that `cudaMalloc` is doing the same: https://docs.nvidia.com/cuda/cuda-c-programming-guide/index.html#device-memory-accesses.
The user, however, should not be concerned by this, block should be as large as the user requested, only the allocator should know that remaining `((requested_size - 1) / 256 + 1) * 256 - requested_size` bytes are "junk".
I will create an issue and fix it. In the meantime `roundup_allocation` workaround should suffice

mimaric: @r-mafi I agree with your second point, but that is the "regular" issue of memory fragmentation due to alignment which also affect any (not specifically designed) allocation functions, including `cudaMalloc`, so I don't think we should be concerned by it

mimaric: Added issue #600

atadkase: Thanks @mimaric and @r-mafi. It makes sense. Should I go ahead and merge this in?

tijyojwad (approved):

Issue #527, opened 2020-08-04
[cudapoa] add description to error codes


r-mafi:
currently we only print out the error code. It helps to have a description added. We can even for some cases, add hints about which corresponding parameters in `BatchConfig` object can be modified to fix the error.

Issue #446, opened 2020-05-27
[cudapoa] error types in cudaPOA


r-mafi:
There is no use of `StatusType::seq_len_exceeded_maximum_nodes_per_window` and can be removed.

On the same note, there is a subtle difference between maximum number of nodes per window and maximum graph dimension. The latter needs to be multiple of 4. We can remove `max_nodes_per_window` in `BatchSize` and only use `max_matrix_graph_dimension`. Same applies for banded version of the variables.

PR #608, opened 2020-12-01
[cudapoa] add description and hint to error codes

r-mafi:
closing #446, #527


cudapoa/src/cudapoa.cpp
```diff
@@ -34,6 +34,57 @@ StatusType Init()
...
+        break;
+    case StatusType::loop_count_exceeded_upper_bound:
+        error_message = "Kernel Error: Traceback in Needleman-Wunsch algorithm failed";
+        error_hint    = "Suggestion  : You may retry with a different banding mode";
+        break;
+    case StatusType::output_type_unavailable:
+        error_message = "Kernel Error: Output type not available";
+        error_hint    = "Suggestion  : Check MSA/Consensus selection for output type";
+        break;
+    case StatusType::generic_error:
```
tijyojwad, line 77: since this state and default are the same, we could just make it fall through
```
case StatusType::generic_error:
default:
    error_message = ...
```

Also, let's just say "Unknown" error since it may not be kernel related

tijyojwad (changes requested):
This is a fantastic change, @r-mafi ! I love it. Thank you so much for doing this!

cudapoa/src/cudapoa.cpp
r-mafi: good point. fixed.

tijyojwad (approved):

r-mafi:
rerun tests

r-mafi:
rerun tests

PR #610, opened 2020-12-03
[cudamapper] fix substr api call in kmerization function

### LINKED ISSUE #591: [cudamapper] bug in extracting kmers
[here](https://github.com/clara-parabricks/GenomeWorks/blob/dev-v0.6.0/cudamapper/src/cudamapper_utils.cpp#L49), the second argument of `std::substr` is length of the substring, not the end position. Should be changed from `i + kmer_size` to `kmer_size`

tijyojwad:
Closes #591


r-mafi (approved):

tijyojwad:
rerun tests

PR #612, opened 2020-12-03
[cudaaligner] Fixed a bug in approximate alignment backtrace

ahehn-nv:
Fixed the bug causing a failing assert found by @r-mafi .


cudaaligner/tests/Test_ApproximateBandedMyers.cpp
```diff
@@ -56,8 +56,70 @@ class TestApproximateBandedMyers : public ::testing::TestWithParam<TestCase>
...
+TEST(TestApproximateBandedMyersStatic, ImplicitNWEntries1)
+{
+    // The banded Myers implementation uses implicit entries for the 0-th row of the NW matrix (and the band),
+    // which is assumed to be the worst case value. This does not pose a problem on the top left block and
+    // the diagonal band, but on the lower right block there are cases where the backtrace runs through this worst case
+    // - which is technically still part of the band. This tests this specific corner case.
+
+    // Band of the NW matrix: (* = tested implicit 0-row entry)
+    //																																	      *      *
+    //       A  A  C  C  G  G     T     T     A     A     A     A     C     C     C     C     G     G     G     G     G     T     T     A     A      A  C  G  G  T  T
```
tijyojwad, line 92: can you explain this layout a bit again? I'm not quite following the setup

tijyojwad (changes requested):

tijyojwad:
changes look good to me overall! just left a clarifying question for the test

cudaaligner/tests/Test_ApproximateBandedMyers.cpp
ahehn-nv: Good point, the layout isn't self-explanatory if you haven't just worked on the implementation. I improved the comment. Could you check if it is clear now?

r-mafi: Great you came up with a simplified corner test case. Thanks for updating the comments, with the row indices, it is more clear now, but still I find it a bit confusing. May be since I'm more of an imagery than a text guy. Something like the following image might help people like me to process the example faster.
<pre>
________________
| | 
| | | 
    | |
      | |
        | |
          | | |
___________ | |_
</pre>
also not sure how for row 20, implicit row 0 comes into play, but I assume that's because I'm not familiar with the banded Myers algorithm.

r-mafi: or alternatively instead of the image, actually display the row shifts per columns in the table, in addition to the comment you added here: 

> diagonal band (each column shifted by one row)

tijyojwad: Much clearer now, thanks @ahehn-nv !

tijyojwad (approved):

cudaaligner/tests/Test_ApproximateBandedMyers.cpp
tijyojwad: oh no, my github page didn't refresh and I didn't see @r-mafi's comment when I closed and merged! Please address as needed and create a separate PR. Sorry for the added work :(.

r-mafi: it was just a minor suggestion, all good.

ahehn-nv: I could add the picture, but I'd leave the data arrangement as it is, since this is how it is stored in memory.

Issue #604, opened 2020-11-26
[cudapoa] throw an error in case all weights are zero


r-mafi:
heavy-bundle algorithm in POA to generate consensus assumes non-zero base weights. We should check and throw an error in case this assumption is not satisfied, otherwise cudaPOA will potentially generate some incorrect consensus output (since all the paths have the same weight of 0).

PR #613, opened 2020-12-04
[cudapoa] check to ensure not all the base weights are zero

r-mafi:
closes issue #604

### DIFF
--- a/cudapoa/src/cudapoa_batch.cuh
+++ b/cudapoa/src/cudapoa_batch.cuh
@@ -134,15 +134,23 @@ public:
         // If a new group can be added, attempt to add all entries
         // in the group. If they can't be added, record their status
         // and continue adding till the end of the group.
+        bool poa_empty = true;
         for (auto& entry : poa_group)
         {
             StatusType entry_status = add_seq_to_poa(entry.seq,
                                                      entry.weights,
                                                      entry.length);
-
+            if (entry_status == StatusType::success)
+            {
+                poa_empty = false;
+            }
             per_seq_status.push_back(entry_status);
         }
 
+        if (poa_empty)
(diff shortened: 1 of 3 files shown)


r-mafi:
rerun tests

cudapoa/src/cudapoa_batch.cuh
```diff
@@ -502,11 +502,23 @@ protected:
...
+                if (weights[i] > 0)
+                {
+                    all_base_weights_are_zero = false;
+                }
+            }
+
+            // all base weights can not be 0
+            if (all_base_weights_are_zero)
+            {
+                throw std::invalid_argument("All base weights can not be zero");
```
tijyojwad, line 519: let's return an error code here like "cudapoa_zero_weighted_sequence" instead of throwing an exception. that way these sequences can be skipped by the user if they so choose

tijyojwad (changes requested):

cudapoa/src/cudapoa_batch.cuh
r-mafi: yes, that way is less interruptive. My initial thought was if returned as error code it may remain hidden, since we add it to the status vector in `Batch::add_poa_group(vector<StatusType>& per_seq_status,  const Group& poa_group)`, but return status of  `StatusType Batch::add_poa_group()` will be `success`, even if all sequences to be added to the group fail.
as a workaround, I added a couple of error codes, `zero_weighted_poa_sequence` and `empty_poa_group`

tijyojwad: awesome, I like this!

tijyojwad (approved):

PR #617, opened 2020-12-14
[cmake] unify cmake cxx flags across GW

### LINKED ISSUE #291: [cga] Centralize compilation flags
Currently every module defines its own compilation flags. In some cases this is the wanted behavior, but in others (C++ standard, warning level..) it causes confusion.

Think about which flags should be set for the whole SDK.

tijyojwad:
Centralize the cmake flags to ensure all modules inherit common configurations.

Closes #291


cudaaligner/benchmarks/main.cpp
```diff
@@ -80,6 +80,8 @@ typename std::enable_if<!std::is_same<AlignerT, AlignerGlobalMyersBanded>::value
 template <typename AlignerT>
 typename std::enable_if<std::is_same<AlignerT, AlignerGlobalMyersBanded>::value, std::unique_ptr<Aligner>>::type create_aligner_tmp_dispatch(int32_t genome_size, int32_t alignments_per_batch, DefaultDeviceAllocator allocator, cudaStream_t stream, int32_t device_id)
 {
+    (void)genome_size;
```
mimaric, line 83: It does not make much difference, but you can use C++ cast: `static_cast<void>(genome_size);`

cudapoa/tests/Test_CudapoaAddAlignment.cu
```diff
@@ -41,6 +41,7 @@ public:
                    std::vector<uint8_t> read, std::vector<int8_t> base_weights, std::vector<int16_t> alignment_graph, std::vector<int16_t> alignment_read)
         : graph(nodes, outgoing_edges, node_alignments, node_coverage_counts)
         , read_(read)
+        , base_weights_(base_weights)
```
mimaric, line 44: I assume this is ok, but not sure

mimaric (commented):
C++ changes look good, other than two optional comments I made.
Adding @ahehn-nv for CMake part

cudapoa/tests/Test_CudapoaAddAlignment.cu
tijyojwad: this was throwing an `unused parameter` error - but the API of the class exposing accessing base_weights (though I assume it's not used anywhere yet).

cudaaligner/benchmarks/main.cpp
tijyojwad: addressed

ahehn-nv (approved):
looks good to me.

mimaric:
Good for me as well, but one test is failing and complains about `min()` and `max()` for 16-bit in `common/base/include/claraparabricks/genomeworks/utils/limits.cuh`

tijyojwad:
rerun tests

tijyojwad:
Hey @ahehn-nv and @mimaric  - I'm running into a gcc error after the merge from https://github.com/clara-parabricks/GenomeWorks/pull/626 which added `libcudaxx` dependency. The `-Werror` now kicks in with an `unused argument` error for the following - 

```
/jenkins/workspace/clara-genomics/gpuci/clara-genomics-analysis/prb/clara-genomics-analysis-cpu-build/3rdparty/libcudacxx/include/cuda/std/atomic:161:73: error: unused parameter ‘_Scope’ [-Werror=unused-parameter]

 inline __host__ __device__ void atomic_thread_fence(memory_order __m, thread_scope _Scope = thread_scope_system) {

                                                                         ^

cc1plus: all warnings being treated as errors

CMake Error at cudaaligner_generated_myers_gpu.cu.o.Release.cmake:280 (message):

  Error generating file

  /jenkins/workspace/clara-genomics/gpuci/clara-genomics-analysis/prb/clara-genomics-analysis-cpu-build/build/cudaaligner/CMakeFiles/cudaaligner.dir/src/./cudaaligner_generated_myers_gpu.cu.o
```

Now this shows up in the core `cudaaligner` compilation, so I had to ignore this particular check in `myers_gpu.cu`. Let me know if there's a better way to handle this.

ahehn-nv:
Not sure if there is a better way.
If I remember correctly, disabling warnings like that was recommended by the "C++ Coding Standards" book, right?
Personally, I don't really like putting things like this in the code, especially since they are compiler specific. I'd just remove the `-Werror` from the compilation flags, live with the warnings that are beyond our control and make sure that our code compiles warning free manually. I.e. fix a warning whenever I see one and make all regular contributors aware that their code should compile warning free (and use `git blame` extensively! ;) ). But that's just my personal view.

tijyojwad:
I agree that once it goes it, it's hard to take out especially because there's no frequent review of these exceptions. That said, I personally prefer to err on the side of caution and leave the strict checking with `-Werror` in there since it forces us to evaluate what's causing the warning. With best effort, I've rarely seen a consistent effort put in by all developers to ensure warning free compilation. Since it's the recommended way, I'd say let's keep it as is and if we find ourselves adding a whole lot of exceptions, then we can revisit taking out `-Werror` :).

PR #622, opened 2021-01-27
Add missing error checks to avoid silent errors

ahehn-nv:
Adding error checks after all kernel calls to avoid silent errors.
This is especially important for `cudaErrorNoKernelImageForDevice` errors. Those errors happen when the kernels are launched on a GPU which does not match the architecture the code was compiled for and the runtime has cannot jit-compile the kernels due to missing or incompatible PTX code.


mimaric (approved):

EpiSlim (approved):
Looks good to me. I would add the missing compute capability to the error message with `deviceProp::major`, but this is definitely a luxury to have.

ahehn-nv:
Good idea. I added it.

EpiSlim (approved):

PR #629, opened 2021-02-16
[cudaaligner] Introduce FixedBandAligner interface

ahehn-nv:
* Added an abstract base class FixedBandAligner (which provides a `aligner->reset_max_bandwidth(int32_t)`).
* Renamed `aligner->reset_bandwidth(int32_t)` to `aligner->reset_max_bandwidth(int32_t)`.
* Fixed a memory-leak in device_buffer's move-assignment

### DIFF
--- a/cudaaligner/src/aligner_global_myers_banded.cpp
+++ b/cudaaligner/src/aligner_global_myers_banded.cpp
@@ -62,6 +62,11 @@ struct memory_distribution
     int64_t remainder;
 };
 
+int64_t get_total_memory_required(const memory_distribution& mem)
+{
+    return 2 * mem.sequence_memory + mem.results_memory + mem.query_patterns_memory + 2 * mem.pmvs_matrix_memory + mem.score_matrix_memory + mem.remainder;
+}
+
 memory_distribution split_available_memory(const int64_t max_device_memory, const int32_t max_bandwidth)
 {
     assert(max_device_memory >= 0);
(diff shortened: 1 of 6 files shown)


cudaaligner/src/aligner_global_myers_banded.cpp
```diff
@@ -84,7 +89,8 @@ memory_distribution split_available_memory(const int64_t max_device_memory, cons
     r.query_patterns_memory = static_cast<int64_t>(fmax_device_memory / mem_req_total_per_bp * mem_req_query_patterns);
     r.pmvs_matrix_memory    = static_cast<int64_t>(fmax_device_memory / mem_req_total_per_bp * mem_req_pmvs_matrix);
     r.score_matrix_memory   = static_cast<int64_t>(fmax_device_memory / mem_req_total_per_bp * mem_req_score_matrix);
-    r.remainder             = max_device_memory - (2 * r.sequence_memory + r.results_memory + r.query_patterns_memory + 2 * r.pmvs_matrix_memory + r.score_matrix_memory);
+    r.remainder             = 0;
```
mimaric, line 92: Why set to `0` and then set to actual value?

ahehn-nv: Because `get_total_memory_required(r)` depends on `r.remainder`. Probably this needs a comment.
First, the total memory required by everything without the remainder is computed and then the remainder is set to `max_device_memory` minus this required memory.

mimaric: Ahh, yes, now I see it, but it is a bit confusing

mimaric: Yes, a comment would be nice

ahehn-nv: Done.

mimaric (approved):

PR #634, opened 2021-02-23
@GW_USE_DEVICE_ALLOCATOR_FILE@

pb-cdunn:
Set `@GW_USE_DEVICE_ALLOCATOR_FILE@` based on `gw_enable_caching_allocator`, to include only one DefaultDeviceAllocator for a given installation.

This way, after the library is built and installed, the user does not need to specify `-DGW_ENABLE_CACHING_ALLOCATOR`,
which would currently select the `DefaultDeviceAllocator` to use.

I have not tested this yet. (I think we will need to add the build-directory to the include path.) I wanted to get opinions on this idea first.

To be clear, here is the error that can hit if we fail to use the same compilation flags as when the library was installed:
```
src/libpancake.so.1.1.0: undefined reference to
`claraparabricks::genomeworks::cudaaligner::create_aligner(
    claraparabricks::genomeworks::cudaaligner::AlignmentType, int, CUstream_st*, int,
    claraparabricks::genomeworks::CudaMallocAllocator<char>, long)'
```


GPUtester:

ahehn-nv:
HI @pb-cdunn , thanks for submitting this PR. This certainly is a problem that needs to be addressed. We are evaluating your PR and will get back to you soon.

ohadmo:
ok to test

PR #636, opened 2021-02-24
[cudamapper] Ignore -Wunused-parameter for Thrust

mimaric:
Added pragmas for ignoring `-Wunused-parameter` in Thrust.

I only added it in needed places. @ahehn-nv do you think we should simply do it everywhere where Thrust is included?


mimaric:
rerun tests

ahehn-nv (approved):

mimaric:
rerun tests

PR #640, opened 2021-03-15
[sdk][cudaaligner] Misc. improvements

ahehn-nv:
SDK:
* improve inlining of `gpu_assert`
* reduced frequency of `cudaSetDevice()` calls in `scoped_device_switch`
* fixed a shadow warning

cudaaligner:
* added `reserve_n_matrices()` function to batch_device_matrices
* init CUDA runtime in `cudaaligner::Init()`.

### DIFF
--- a/cudaaligner/src/cudaaligner.cpp
+++ b/cudaaligner/src/cudaaligner.cpp
@@ -16,6 +16,7 @@
 
 #include <claraparabricks/genomeworks/cudaaligner/cudaaligner.hpp>
 #include <claraparabricks/genomeworks/logging/logging.hpp>
+#include <claraparabricks/genomeworks/utils/cudautils.hpp>
 
 namespace claraparabricks
 {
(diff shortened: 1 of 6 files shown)


cudaaligner/src/cudaaligner.cpp
```diff
@@ -29,6 +30,7 @@ namespace cudaaligner
 StatusType Init()
 {
     initialize_logger(claraparabricks::genomeworks::logging::LogLevel::warn);
+    GW_CU_CHECK_ERR(cudaFree(nullptr)); // initialize CUDA runtime
```
mimaric, line 33: Makes sense, I'm just wondering why is it necessary

mimaric (approved):

cudaaligner/src/cudaaligner.cpp
ahehn-nv: Not strictly necessary, but since we have an `Init()` function, we may also actually do the initialization.

PR #641, opened 2021-03-15
[cudaaligner] Use RLE for alignment

ahehn-nv:
* Instead of storing a flat sequence of `AlignmentType` operations, the aligner now stores a sequence of `AlignmentType` operations ("actions") and a sequence of `counts` which runlength-encode the flat sequence.
* Added functions to the `AlignmentImpl` class to store and obtain the RLE sequences and convert them to a CIGAR string. The class maintains backward compatibility for aligners that still provide the flat sequence. Those aligners need to be adapted in the future.
* banded Myers algorithm: Handle special cases, where query length and/or target length are 0.


mimaric (approved):

ahehn-nv:
Walked Milos through the changes to double check again.

